PR #106, opened 2012-07-24
rotate thumbnails if there is exifdata to know how

marianoguerra:
rotate thumbnails according to Orientation exif metadata field if available to avoid pictures taken with the camera rotated to be rotated as well.

I think they appear rotated because PIL may remove the EXIF metadata from the thumbnail and then the image viewers can't autorotate the image when displaying it.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1036,10 +1036,11 @@ class Nikola(object):
             return
         try:
             import Image as _Image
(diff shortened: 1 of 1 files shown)


gaelagaton431-netizen:
mariano

gaelagaton431-netizen:
mariano

Issue #645, opened 2013-09-02
Strip feed texts of dangerous tags


ralsina:
http://feedvalidator.org recommends removing iframe and a bunch of other tags from feed contents:

http://feedvalidator.org/docs/warning/SecurityRisk.html

lxml.etree.strip_tags() does sort of work for this, but removes the tags leaving the contents, which may not be the right thing to do.

This is probably the right thing to use, it can both remove (keep contents) and kill (remove contents) tags:

http://lxml.de/api/lxml.html.clean.Cleaner-class.html

ralsina:
While this is possible, and there was even, at a point, code to do it in a PR, the end result was not good, because too much interesting data got removed, and feed readers can actually handle most of it nowadays. 

If someone wants to do it, and can find a way avoiding that problem, feel free.

PR #648, opened 2013-09-02
Clean HTML in feeds of "dangerous" tags (#645)

ralsina:


ralsina:
This is clearly not happening.

PR #1300, opened 2014-05-24
Remove AddThis by default

Kwpolska:
Remove AddThis, as discussed on the ML.

This might not be the perfect time to do it, it might need to wait till v8.  Opinions?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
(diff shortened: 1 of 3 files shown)


ralsina:
Let's just kill it.

Kwpolska:
Ignoring the release cycle?  Awesome!

Merging now (also fixing a leftover from `git merge`).

PR #1564, opened 2015-01-02
fix #1563 -- fix rss.xsl path for subdir sites

### LINKED ISSUE #1563: All RSS feeds contain invalid link to rss.xsl for subdirectory deployments
Spotted another 404 in the logs. `http://example.com/blog/rss.xml` contains `href="../assets/xml/rss.xsl"` which leads outside of the blog directory.  Same thing with other rss feeds in subdirectories.

Kwpolska:
This is #1563.

cc @max-arnold

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -63,6 +63,8 @@ Features
 Bugfixes
 --------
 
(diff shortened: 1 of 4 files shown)


max-arnold:
Thanks!

PR #1565, opened 2015-01-03
Avoid rebuild of tag/category pages/feeds when tag/category is added/removed.

### LINKED ISSUE #1410: Stop rebuilding RSS files for tags that don’t need it
1. Add a few posts with some tags
2. nikola build
3. Edit any one of the posts and add a new tag
4. nikola build

All tags’ RSS files are rebuilt to facilitate adding one new tag. Even when not all of them have changing content. They get new `<lastBuildDate>` which causes utilities like rsync and git to treat them as new files.

We should be more selective about what feeds need rebuilding.

felixfontein:
This patch avoids rebuilding of tag and category pages and RSS feeds if a tag or category was added or removed; this was caused because every tag and category page and RSS feed had a list of all tags and categories incorporated into its ```uptodate``` list (via ```kw```).

Fixes #1410.

### DIFF
--- a/nikola/plugins/task/tags.py
+++ b/nikola/plugins/task/tags.py
@@ -84,7 +84,7 @@ class RenderTags(Task):
         self.site.scan_posts()
         yield self.group_task()
 
-        yield self.list_tags_page(kw)  # this also adds category and tag list to kw
+        yield self.list_tags_page(kw)
 
         if not self.site.posts_per_tag and not self.site.posts_per_category:
             return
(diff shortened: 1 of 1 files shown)


Kwpolska:
Previously, tag pages (eg. `/tags/foo.html`) had a list of all the tags in `kw`, because it leaked from the tag index (`/tags/index.html`).

Now, the tag index has the list in its own copy of `kw` (and *only* that copy has the tag list), and tag pages have a list of posts they need in  `uptodate` that comes from the `context` via `generic_post_list_renderer`, and also depend on the post files.

Correct?

felixfontein:
Precisely. (If category and tag folder are different, then both indices have only the list they need in their copy of ```kw```.)

PR #1580, opened 2015-01-16
RSS2 expects lastBuildDate to be GMT.

punchagan:
PyRSS2Gen expects the lastBuildDate to be in GMT, and ignores any
timezone information when generating the RSS.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1021,7 +1021,7 @@ class Nikola(object):
             title=title,
             link=link,
             description=description,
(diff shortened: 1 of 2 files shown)


Kwpolska:
Why are we using something so ancient… we should switch to `feedgenerator` one day.

Also, I have no idea why AppVeyor *failed to create process*.  Probably not our fault.

ralsina:
Well, this code dates back to 2001 or so in one form or another :-)

PR #1582, opened 2015-01-17
Fix #1581

### LINKED ISSUE #1581: Build fails with IPython 3.0
Trying to use ipython notebooks with the current dev version of IPython (3.0.0) fails building with some warnings etc. because the ``nbformat`` interface has changed a little:

```
...WARNING: UserWarning: .../ipython-dev/IPython/nbformat/current.py:19: IPython.nbformat.current is deprecated.

- use IPython.nbformat for read/write/validate public API
- use IPython.nbformat.vX directly to composing notebooks of a particular version
...
... WARNING: UserWarning: .../ipython-dev/IPython/nbformat/current.py:75: reads_json is deprecated, use reads
...
AttributeError: cells
```

This is fairly easily fixed and I will send a PR shortly.

mforbes:
This adds a bit of code to deal with the API changes in the latest development version of IPython 3.0.0.  Note: I read the notebook into the current format (as opposed to ``nbformat.NO_CONVERT``) so that the ``HTMLExporter`` can properly output the data.  If the conversion is not done, then this will fail if there are ``cells`` in the notebook as shown in the issue.

There is probably more that could be cleaned up (the generated notebook for an initial post should probably use the current format rather than a hard-coded nbformat 3 template for example), but I wanted to keep this a minimal change to preserve backward compatibility.

### DIFF
--- a/nikola/plugins/compile/ipynb/__init__.py
+++ b/nikola/plugins/compile/ipynb/__init__.py
@@ -31,15 +31,14 @@ import io
 import os
 
 try:
+    import IPython
     from IPython.nbconvert.exporters import HTMLExporter
-    import IPython.nbformat
     if IPython.version_info[0] >= 3:     # API changed with 3.0.0
-        def nbformat_reads(s):
-            # Update to the current format so that the export works.
-            return IPython.nbformat.reads(
-                s, IPython.nbformat.current_nbformat)
+        from IPython import nbformat
+        current_nbformat = nbformat.current_nbformat
     else:
-        nbformat_reads = IPython.nbformat.current.reads_json
+        import IPython.nbformat.current as nbformat
+        current_nbformat = 'json'
 
     from IPython.config import Config
     flag = True
@@ -66,8 +65,7 @@ class CompileIPynb(PageCompiler):
         exportHtml = HTMLExporter(config=c)
         with io.open(dest, "w+", encoding="utf8") as out_file:
             with io.open(source, "r", encoding="utf8") as in_file:
-                nb = in_file.read()
-                nb_json = nbformat_reads(nb)
+                nb_json = nbformat.read(in_file, current_nbformat)
             (body, resources) = exportHtml.from_notebook_node(nb_json)
             out_file.write(body)
 
--- a/nikola/plugins/compile/ipynb/__init__.py
+++ b/nikola/plugins/compile/ipynb/__init__.py
@@ -32,7 +32,15 @@ import os
 
 try:
     from IPython.nbconvert.exporters import HTMLExporter
-    from IPython.nbformat import current as nbformat
+    import IPython.nbformat
+    if IPython.version_info[0] >= 3:     # API changed with 3.0.0
+        def nbformat_reads(s):
+            # Update to the current format so that the export works.
+            return IPython.nbformat.reads(
+                s, IPython.nbformat.current_nbformat)
+    else:
+        nbformat_reads = IPython.nbformat.current.reads_json
+
     from IPython.config import Config
     flag = True
 except ImportError:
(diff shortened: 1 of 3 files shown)


nikola/plugins/compile/ipynb/__init__.py
```diff
@@ -32,7 +32,15 @@
...
     from IPython.nbconvert.exporters import HTMLExporter
-    from IPython.nbformat import current as nbformat
+    import IPython.nbformat
+    if IPython.version_info[0] >= 3:     # API changed with 3.0.0
+        def nbformat_reads(s):
+            # Update to the current format so that the export works.
+            return IPython.nbformat.reads(
+                s, IPython.nbformat.current_nbformat)
+    else:
+        nbformat_reads = IPython.nbformat.current.reads_json
```
Kwpolska: Can’t you just use plain `reads` on all versions?

```py
nb_json = IPython.nbformat.reads(s)
```

seems to be enough on v2 and v3, at least according to the docs.

Kwpolska: Also, you would have to `import IPython.nbformat.current` and work with it all the time.

Here is a revised solution:

Kwpolska:
```py
if IPython.version_info[0] >= 3:
    nbformat_reads = lambda s: IPython.nbformat.reads(s)
else:
    import IPython.nbformat.current
    nbformat_reads = IPython.nbformat.current.reads_json
```

mforbes: I believe that ``reads()`` requires a format as a second argument.

Kwpolska: Ah yes, it does.  I was reading the wrong file.

```py
    nbformat_reads = lambda s: IPython.nbformat.reads(s, IPython.nbformat.current_nbformat)
```

mforbes:
I think this should work but have not tested it with IPython 2.0 yet...

nikola/plugins/compile/ipynb/__init__.py
```diff
@@ -31,8 +31,15 @@
...
 try:
+    import IPython
     from IPython.nbconvert.exporters import HTMLExporter
-    from IPython.nbformat import current as nbformat
+    if IPython.version_info[0] >= 3:     # API changed with 3.0.0
+        from IPython import nbformat
+        current_nbformat = nbformat.current_nbformat
+    else:
+        import IPython.nbformat.current as nbformat
+        current_nbformat = 'json'
```
mforbes: Maybe this should be a Unicode string (``u'json'``)?  (I don't think it matters though, but have not really done much with unicode strings yet...)

Kwpolska: It already is Unicode:

```py
from __future__ import unicode_literals, print_function
                       ^^^^^^^^^^^^^^^^
```

Kwpolska:
Works fine with v2.3.1:

```
In [1]: import IPython.nbformat.current as nbformat

In [2]: import io

In [3]: in_file = io.open('plot-test.ipynb', 'r', encoding='utf-8')

In [4]: nb_json = nbformat.read(in_file, u'json')

In [5]: nb_json
Out[5]: 
{'nbformat': 3,
 'nbformat_minor': 0,
…

In [6]: type(_)
Out[6]: IPython.nbformat.v3.nbbase.NotebookNode

```

If it works with the current v3 dev:

 * add yourself to /AUTHORS.txt
 * add this to /CHANGES.txt (on the **top** of the file):

   ```rst
New in master
=============

Features
--------

Bugfixes
--------

* Fixed compatibility with IPython 3.x (Issue #1582)
   ```

Merging as soon as this is done.

mforbes:
Shouldn't that be: Issue #1581?

```
* Fixed compatibility with IPython 3.x (Issue #1581)
```

Kwpolska:
Sure.

Kwpolska:
:+1: Thanks for your contribution!

nikola/plugins/compile/ipynb/__init__.py
muxuezi: cool! It's work after I upgrade to IPython3.0.
Thanks!

Issue #1523, opened 2014-12-05
Compiler plugins should add dependency on config options


ralsina:
For example, the markdown compiler plugin will do different things if you change MARKDOWN_EXTENSIONS so the compiler plugins should declare those somewhere so Nikola can rebuild those as needed.

PR #1585, opened 2015-01-17
Compiler config deps

ralsina:
This should fix Issue #1523

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,7 @@ Bugfixes
 --------
 
 * Fixed compatibility with IPython 3.x (Issue #1581)
+* Compilers mark tasks as out of date if compiler-specific options and plugins change (Issue #1523)
 
 New in v7.3.0
 =============
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -53,16 +53,18 @@ class CompileMarkdown(PageCompiler):
     site = None
 
     def set_site(self, site):
+        self.enabled_extensions = []
         for plugin_info in site.plugin_manager.getPluginsOfCategory("MarkdownExtension"):
             if plugin_info.name in site.config['DISABLED_PLUGINS']:
                 site.plugin_manager.removePluginFromCategory(plugin_info, "MarkdownExtension")
                 continue
-
+            self.enabled_extensions.append(plugin_info.name)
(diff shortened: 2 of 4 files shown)


felixfontein:
I think ```post.add_dependency_uptodate``` is not used correctly here. It is more or less directly added (for lists, its elements are added) to doit's ```uptodate``` field, which expects lists of ```False```, ```True```, ```None```, callables, or strings (which are executed as shell commands) (see http://pydoit.org/dependencies.html#uptodate). I think ```tools.config_changed``` should be used here to (somehow) wrap the data added.

ralsina:
These are all lists of strings, so is that ok?

felixfontein:
I think doit will try to execute each string as a shell command, and use its return value to determine whether the task changed or not. That's not what you want, I think?

ralsina:
Indeed.

ralsina:
Just out of curiosity, is add_dependency_uptodate used anywhere so I can see how it's done there?

felixfontein:
If you want to update if the list of strings changed (maybe you want to sort the list to avoid list reordering triggering rebuilds), use ```tools.config_changed({1: your_string_list}, 'some unique identifier for the plugin')``` or so.

felixfontein:
I found an example in one of my (not yet published) plugins:

```post.add_dependency_uptodate(utils.config_changed({1: some_data}, 'nikola.plugins.path.to.comment.and.its.name:' + post.base_path), False, 'page')```

ralsina:
ok, pushing now.

ralsina:
The ID you say "unique"... that means unique in the context of that specific task, right? Not unique across all the tasks?

felixfontein:
It should be unique over all posts and tasks, because it could happen that different tasks add something to the uptodate dependency of a post, and many posts are on one index/list/... page, and all those uptodate dependencies are put into the uptodate dependency on the generated index/list/...

I'd use a combination of the task/plugin/... name and the post's source path.

ralsina:
Ugh again then.

ralsina:
Last question: it has to be unique across the whole build **and stable from one build to another**?

felixfontein:
Yes. (Of course, it should be stable iff the options it contains don't change.)
So if you want to force a rebuilt if a list of strings changes (but its order doesn't depend), sort the list, put it into a ```config_changed``` object and make its tag depend on the plugin and the post.

PR #1587, opened 2015-01-17
Compiler config deps

ralsina:
Use add_dependency_uptodate correctly

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -224,6 +224,7 @@ class PageCompiler(BasePlugin):
         'description': '',
         'type': 'text',
     }
+    config_dependencies = []
 
     def register_extra_dependencies(self, post):
         """Add additional dependencies to the post object.
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -53,18 +53,18 @@ class CompileMarkdown(PageCompiler):
     site = None
 
     def set_site(self, site):
-        self.enabled_extensions = []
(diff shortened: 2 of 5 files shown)


ralsina:
I think this way makes more sense. Comments appreciated :-)

Kwpolska:
:+1: from me.

felixfontein:
This still needs some packaging, like in a ```config_changed``` object. I'll add that.

felixfontein:
Well, there are two things we could do:
  a) ask that only correct things are put into ```compiler.config_dependencies```;
  b) change ```compiler.config_dependencies``` into a set and ask for strings to be put into it, then ```post.uptodate()``` has to create a sorted list out of it and put it into a ```tools.config_changed``` object.

Which one do you prefer?

ralsina:
Let's keep the complexity in one place, otherwise for each compiler we have to know all this stuff about why it has to be a set and why it has to have a unique ID, and how unique it has to be, etc.

I'd say the compiler just has a list of strings, then the post creates a config_changed up to its own requirements out of it.

I think this is better because while there is a chance 3rd parties will want to write a compiler plugin, noone outside the project is touching Post.

felixfontein:
Sounds good to me. And if anyone needs to add something more complex than simply a string (or several of them), or if the data depends on the post as well, they can still use the more complicated mechanism.

Kwpolska:
Without this, nikola is broken badly.  Merging.

felixfontein:
It is indeed. But there was something missing for pandoc and markdown: the compilers also inserted a list of strings into the list (instead of strings). Merging that, too.

PR #1589, opened 2015-01-18
Fixing order for other metadata.

felixfontein:
This PR fixes the order of metadata written via ```utils.write_metadata()```. Currently, a 'random' order (induced by the dict used) is taken.

I'm currently working on a small tag management plugin, which needs to read and write metadata. I want to avoid that every time the tags are modified, other metadata ends up in a new random order.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -35,6 +35,7 @@ import hashlib
 import io
 import locale
 import logging
+import natsort
 import os
 import re
 import json
@@ -1241,8 +1242,8 @@ def write_metadata(data):
             pass
 
     # Leftover metadata (user-specified/non-default).
-    for k, v in data.items():
-        meta.append(f.format(k, v))
+    for k in natsort.natsorted(list(data.keys())):
+        meta.append(f.format(k, data[k]))
 
     meta.append('')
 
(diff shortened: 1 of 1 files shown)


Kwpolska:
> small tag management plugin

Are you aware of http://plugins.getnikola.com/#tags ?

nikola/utils.py
```diff
@@ -1241,8 +1242,8 @@ def write_metadata(data):
             pass
 
     # Leftover metadata (user-specified/non-default).
-    for k, v in data.items():
-        meta.append(f.format(k, v))
+    for k in natsort.natsorted(list(data.keys())):
```
Kwpolska:
```py
    for k in natsort.natsorted(list(data.keys()), alg=natsort.ns.F | natsort.ns.IC):
```

felixfontein:
No, I wasn't aware of that plugin. Looks like that does most of what I wanted to do :)

(I originally started writing my plugin to have something nice to get a list of all tags with number of use; I only later got the idea to add tag rename/... support. Well, no need to work that out, then.)

felixfontein:
Ok, added your suggestion. Using the additional ```natsorted``` arguments doesn't matter here too much, but it's nice to have anyway.

PR #1601, opened 2015-01-27
Removing empty line that produces an invalid '-->' output when importing md

magmax:
When importing markdown, this line produces metadata like these:

```
<!--
.. title: foo
.. slug: bar
.. date: 2001/01/01

-->
```

That is processed genarating the line:

```
<p>--&gt;</p>
```

What is invalid.

This fix will generate metadata like:
```
<!--
.. title: foo
.. slug: bar
.. date: 2001/01/01
-->
```

What is processed as expected.

### DIFF
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -76,7 +76,7 @@ class CompileMarkdown(PageCompiler):
             with io.open(source, "r", encoding="utf8") as in_file:
                 data = in_file.read()
             if not is_two_file:
-                data = re.split('(\n\n|\r\n\r\n)', data, maxsplit=1)[-1]
+                data = re.split('(\-\->)', data, maxsplit=1)[-1]
             output = markdown(data, self.extensions)
             out_file.write(output)
 


Kwpolska:
This will break reStructuredText though.

Markdown, HTML, etc. need:

```html
<!--
.. title: foo
-->

text
```

reStructuredText needs:

```rst
.. title: foo

text
```

magmax:
No, it doesn't because in file `nikola/plugins/compile/rest/__init__.py`:

```
            if onefile:
                fd.write(write_metadata(metadata))
                fd.write('\n')
            fd.write(content)
```

So the line feed is already there.

Kwpolska:
Okay.  What do you mean by **importing**?  Do you mean `nikola new_post -i /path/to/other/md/file.md`?

If yes, you fixed the wrong place.  On your branch, all the compilers generate garbage in new posts:

```
.. type: text-->

Write your post here.
```

```
.. type: text
Write your post here.
```

And in imported posts:

```
(nikola-py2)[kwpolska@kw-cassandra /tmp]% nikola init -q n
[2015-01-27T08:44:01Z] INFO: init: Created empty site at n.
(nikola-py2)[kwpolska@kw-cassandra /tmp]% cd n
(nikola-py2)[kwpolska@kw-cassandra n]% vim conf.py import.md
# markdown support added to POSTS
(nikola-py2)[kwpolska@kw-cassandra n]% cat import.md
This is a post that I am willing to import.
(nikola-py2)[kwpolska@kw-cassandra n]% nikola new_post -f markdown -t imp -i import.md
Importing Existing Post
-----------------------

Title: imp
Scanning posts.....done!
[2015-01-27T08:54:19Z] INFO: new_post: Your post's text is at: posts/imp.md
(nikola-py2)[kwpolska@kw-cassandra n]% cat posts/imp.md
<!-- 
.. title: imp
.. slug: imp
.. date: 2015-01-27 08:54:19 UTC
.. tags: 
.. category: 
.. link: 
.. description: 
.. type: text-->

This is a post that I am willing to import.
```

1. What are you doing to *import* posts?
2. What are you importing?
3. Can you reproduce with a **clean site** (extended only to support markdown)?

magmax:
Hmmm.... Ok. I start thinking you are right and the best place to fix it is in the metadata read, instead the write.

I will try to fix it as soon as I can.

Thank you!

magmax:
I have a headache with this issue.

The last change will fix my problem, but I think it is changing the behavior so much that it may fail to someone that was using it incorrectly.

IMHO, there should be just one way to do that: as a comment or not.

But without my fix, whenever you import a markdown with several spaces before the `-->`

Kwpolska:
:-1:

Fix your plugin.  Period.

Issue #1602, opened 2015-01-28
Plugin and extract metadata.


Carreau:
Hi there ! 

Is there a way for plugin to define a way to extract metadata from file ? 
In particular with ipynb plugin as the fileformat is json, it would be relatively easy,
but I don't find information about that. 

I've looked at BasePlugin and PageCompiler (that get a class default_metadata).

Would adding an (optional) `extract_metadata(...)` to `PageCompiler` and call it in the right place(s) be accepted ? The good way to do it?

Could someone give me some pointers on where that should be? 

Ping @damianavila

ralsina:
Currently the data extraction is in the Post class, but it would make tons of sense to refactor it into the compiler plugins.

Carreau:
> Currently the data extraction is in the Post class, but it would make tons of sense to refactor it into the compiler plugins.

Thanks for the quick answer.

With backward compatibility I guess, so 
 - default_metadata
 - if availlable, merge with : from plugin
 - if availlable, merge with : .meta file ?

I'll try to take a stab at it.

ralsina:
Right, more or less that, I think :-)

ralsina:
This was fixed by #1603

PR #1603, opened 2015-01-28
Allow plugins to provide metadata

Carreau:
This allow plugins to provide a way to extract metadata from file the way
they wish to.

Metadata will be search first from the default value of as before this
PR. Metadata extracted by the plugin will then be merged into the
metadata dict, which in turn will be extended by the value in the
`.meta` file if present.

--- 

See start of discussion from #1602 

I tried to get the api to extract metadata by way of plugins, as close as possible as other `get_meta` calls. 
I also implement this for 1 plugin (`.ipynb`) the one I'm interested in.

This should not change current behavior for reading metadata from other places.

### DIFF
--- a/docs/extending.txt
+++ b/docs/extending.txt
@@ -374,6 +374,10 @@ They must provide:
 If the compiler produces something other than HTML files, it should also implement ``extension`` which
 returns the preferred extension for the output file.
 
+These plugin can also be used to extract metadata from file. To do so, the
+plugin may implement ``read_metadata`` that will return a dict containing some
+metadata contained in the file.
+
 RestExtension Plugins
 ---------------------
 
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -245,6 +245,12 @@ class PageCompiler(BasePlugin):
         """The preferred extension for the output of this compiler."""
         return ".html"
 
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """
+        Read the metadata from a post, and return a metadata dict
+        """
+        return {}
+
 
 class RestExtension(BasePlugin):
     name = "dummy_rest_extension"
--- a/nikola/plugins/compile/ipynb/__init__.py
+++ b/nikola/plugins/compile/ipynb/__init__.py
@@ -69,6 +69,19 @@ class CompileIPynb(PageCompiler):
             (body, resources) = exportHtml.from_notebook_node(nb_json)
             out_file.write(body)
 
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """read metadata directly from ipynb file.
+
+        As ipynb file support arbitrary metadata as json, the metadata used by Nikola
+        will be assume to be in the 'nikola' subfield.
+        """
+        source = post.source_path
+        with io.open(source, "r", encoding="utf8") as in_file:
+            nb_json = nbformat.read(in_file, current_nbformat)
+        # metadata shoudl always exist, but we never know
+        # if Someone craft an ipynb by hand.
+        return nb_json.get('metadata', {}).get('nikola', {})
+
     def create_post(self, path, **kw):
         content = kw.pop('content', None)
         onefile = kw.pop('onefile', False)
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -917,8 +917,6 @@ def get_meta(post, file_metadata_regexp=None, unslugify_titles=False, lang=None)
     """
     meta = defaultdict(lambda: '')
 
-    newstylemeta = True
-
     try:
         config = post.config
     except AttributeError:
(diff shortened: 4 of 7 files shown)


nikola/post.py
```diff
@@ -136,6 +136,9 @@ def __init__(
         self._dependency_uptodate_page = defaultdict(list)
 
         default_metadata, self.newstylemeta = get_meta(self, self.config['FILE_METADATA_REGEXP'], self.config['UNSLUGIFY_TITLES'])
+        compiler_meta = self.compiler.read_metadata(self, self.config['FILE_METADATA_REGEXP'], self.config['UNSLUGIFY_TITLES'])
```
Kwpolska: IMO, you should integrate this into `get_meta` instead.

nikola/plugin_categories.py
```diff
@@ -245,6 +245,12 @@ def extension(self):
         """The preferred extension for the output of this compiler."""
         return ".html"
 
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """
+        Read the metadata from a post, and retur a metadata dict
```
Kwpolska: return*

also, you should actually return an empty **dict** below

Carreau:
Sorry for the time to respond, comment should be addressed.
moving things to `get_meta` seem a bit weird, also due to the cause that `get_meta` get passed self as first arg, but is not a method.

Kwpolska:
So what?  `read_metadata` also gets the Post object, and also as `self`.  `read_metadata` should not be put outside of the standard get_meta pipeline.

Carreau:
> So what? read_metadata also gets the Post object, and also as self

No, i just ment that passing self as first argument just felt unnatural to me, I was thus thinking that from inside `get_metadata` I couldn't get access to `self.compiler`, but in fact I can, as it is name `post.compiler` from inside the function. Just me not used to the codebase, sorry if I express myself badly.

I've also forgot to push the cleanup commit that remove my debugs, and comments. 
Tell me if you want me to squash everything into one commit.

nikola/post.py
```diff
@@ -928,6 +928,8 @@ def get_meta(post, file_metadata_regexp=None, unslugify_titles=False, lang=None)
     meta.update(_)
 
     if meta:
+        compiler_meta = post.compiler.read_metadata(post, file_metadata_regexp, unslugify_titles, lang)
```
Kwpolska: why is this still there?

Kwpolska:
1. Failing tests: https://travis-ci.org/getnikola/nikola/jobs/49094459#L871 → dummy posts used for tests do not have a `compiler` attribute
2. Failing flake8: https://travis-ci.org/getnikola/nikola/jobs/49094463#L136
3. The part under `if meta:` should not be there; you should parse compiler meta only once in the code (don’t do it if an explicit `.meta` file exists?  Refactor `.meta` loading?)
4. Add this to CHANGES.txt (at the very top, under Features) and yourself to AUTHORS.txt.

nikola/post.py
Carreau: Still ? You asked me to move it in it's in `get_meta` which I did. Though `get_meta`return in two places, so if the `post.compiler.read_metadata` want to have an effect in all casses, I need to call it before each of the `return`s. Which explain why the call still appear twice in the diff.

Carreau:
> Failing tests: https://travis-ci.org/getnikola/nikola/jobs/49094459#L871 → dummy posts used for tests do not have a compiler attribute
> Failing flake8: https://travis-ci.org/getnikola/nikola/jobs/49094463#L136
> The part under if meta: should not be there; you should parse compiler meta only once in the code (don’t do it if an explicit .meta file exists? Refactor .meta loading?)
> Add this to CHANGES.txt (at the very top, under Features) and yourself to AUTHORS.txt.

Will do.

Kwpolska:
I recommend replacing the `if meta:` block with

```py
if not meta:
    post.is_two_file = False
```

and then putting the `compiler_meta` code only in one place.  It should also appear above the `if lang is None:` block.  This sounds the most sensible to me.

Carreau:
Squashed, comment taken into account. 
Does that looks better to you ?

docs/extending.txt
```diff
@@ -374,6 +374,10 @@ They must provide:
 If the compiler produces something other than HTML files, it should also implement ``extension`` which
 returns the preferred extension for the output file.
 
+These plugin can also be used to extract metadata from file. To do so, the
```
Kwpolska: These plugins*

a dict containing the metadata*

nikola/plugins/compile/ipynb/__init__.py
```diff
@@ -69,6 +69,19 @@ def compile_html(self, source, dest, is_two_file=True):
...
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """read metadata directly from ipynb file.
+
+        As ipynb file support arbitrary metadata as json, the metadata used by Nikola
+        will be assume to be in the 'nikola' subfield.
+        """
+        source = post.source_path
+        with io.open(source, "r", encoding="utf8") as in_file:
+            nb_json = nbformat.read(in_file, current_nbformat)
+        # metadata shoudl always exist, but we never know
```
Kwpolska:
```python
# metadata should always exist, but we never know
# if someone crafted the ipynb by hand.
```

Kwpolska:
Code looks good, only needs some typo fixes in the documentation.  Will fix myself.

Thanks for contributing!

Kwpolska:
(PS. I forgot to tell you squashing commits is unnecessary; we don’t really care.)

Carreau:
> Thanks for contributing!

Thanks for maintaining Nikola and guiding me through the codebase !

PR #1606, opened 2015-02-01
Thumbnail updates

pellenilsson:
Some improvements for my scale_images plugin:

* Thumbnail images in a post would normally be larger than the thumbnails in a gallery, so thumbnail size is now controlled with a separate setting.
* The special treatment of panorama images in galleries plugin has been turned off in scale_images since it would produce very wide thumbnails that messed up page layout.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,8 @@ Features
 
 * Added ``root`` path handler (via Issues #1008, #1573)
 * Added RSS feeds to gallery HEAD (part of Issue #786)
+* Don't create larger thumbnails for panorama images in scale_images
+  plugin
 
 Bugfixes
 --------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,8 +6,9 @@ Features
 
 * Added ``root`` path handler (via Issues #1008, #1573)
 * Added RSS feeds to gallery HEAD (part of Issue #786)
-* Don't create larger thumbnails for panorama images in scale_images
-  plugin
+* Separate new option IMAGE_THUMBNAIL_SIZE for setting size of
+  thumbnails created by scale_images plugin. Also don't create larger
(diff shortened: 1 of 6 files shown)


Kwpolska:
> The special treatment of panorama images in galleries plugin has been turned off in scale_images since it would produce very wide thumbnails that messed up page layout.

Is there a *page layout*?  Last time I checked, galleries just [dump photos at random](http://getnikola.com/galleries/demo/) without any *layout* involved.

And at the same time, your change only destroys image quality.  The thumbnails’ sizes is determined in the output HTML, and it’s not even trying to make the aspect ratio right.

Also, I’m skeptical about `IMAGE_THUMBNAIL_SIZE`, why not produce images of the same size everywhere and be consistent?  It should also default to `180` not to break existing sites.

pellenilsson:
The intended usage scenario for these "thumbnails" is to have illustrations in articles that are enlargeable, like on Wikipedia and similar sites. Please see this page on my site under construction for an example (figures 3, 5 and some others are enlargeable):

http://nyckelpigan.info:8802/pantarei/posts/snowline-tutorial/

THUMBNAIL_SIZE is set a bit lower than the article width so that all images fit nicely into the layout, but figure 9 is treated as a panorama and therefore becomes far too wide to fit.

Kwpolska:
How would it look, if not for the panorama magic?  Also, you seem to be using **images** here and not **galleries**.

I guess you could use it in the post images.

pellenilsson:
It would look like this:

http://nyckelpigan.info:8802/pantarei_nopanorama/posts/snowline-tutorial/

The commit in question will only affect thumbnail images in posts, not galleries.

PS I'm currently travelling and will have limited means of responding for a week or so.

Kwpolska:
Looks okay.  Let’s make 400 the default and hope nobody notices/cares or they just set it back to 180.

PR #1622, opened 2015-02-10
Correction for fix #1615 -- Use makedirs from nikola.utils

### LINKED ISSUE #1615: Cannot deploy after removing cache
Hello,

the deploy command crashes like this:

IOError: [Errno 2] No such file or directory: 'cache/lastdeploy'

Why, yes, of course, I can remove "cache", can't I? Seems I have to "mkdir" it. Also I would like to disable its usage, as the point is unclear to my, using rsync does the trick to me.

pellenilsson:


Kwpolska:
Oops, sorry for that.

Issue #1562, opened 2014-12-31
```EarlyTask``` plugins


felixfontein:
Hi,

I'd like to have a plugin category `EarlyTask`, for tasks which are executed before the site is rendered (i.e. an analogue to `LateTask`). I personally need that for a plugin (or better, combination of plugins) I wrote, currently I used the `Task` plugin class but it happens that some tasks are run after page compiling, while my page compiling plugin needs their result -- and so it fails.

Does anyone mind if I add something like that? Or would it be better to have a general priority system, so you can assign a task a priority (usual tasks could get 10, and late tasks 100, so you could add a task with priority 2 and one with priority 7 to ensure that the one with priority 2 appears in the task list before the one with priority 7 and before all regular rendering tasks and late tasks)?

(I have the vague feeling that I already read something about `EarlyTask`s somewhere here, but I cannot remember where. So it's probably not my own idea :) )

Cheers,
Felix

felixfontein:
Cool, everything's already done! :)
A priority system won't be exactly complicated, since most things can have the same priority -- there are still doit's dependencies to handle that. But then, that was just a suggestion, I'm perfectly happy with ```EarlyTask``` as it is now. Thanks!
Well, and of course, a Happy New Year to you all!

schettino72:
Probably too late but i will chime in anyway :)

doit has 2 distinct phases. task-creation and task-execution.

task-creation (generate task metadata) is usually done in doit using the task_xxx functions ina dodo.py module. But in Nikola it is done through the Nikola site object. Since in Nikola you can add more task-creators through plugins `LateTask` was created to make sure the Nikola site object was set-up before executing task-creators from plugins... 

So `LateTask` means late-generated-task not late-executed-task. `LateTask` is nikola concept and `doit` has no idea about it.

task-execution ordering must be handled by doit using the task property `task_dep`. 

A "priority system" may make sense for Nikola since the global Nikola site object is modified by many plugins, but it doesnt make sense for `doit` itself.

felixfontein:
What do you think? Would it be hard to modify ```DoitNikola``` (which inherits from ```DoitMain```) and ```NikolaTaskLoader``` (inherited from ```TaskLoader```) to get the following behavior:

 * The set of tasks are split up into three (or more) categories: early, usual, late.
 * For all commands (except ```build```), the task loader will behave as it is now, i.e. generate tasks for early, then for usual, then for late, and return a big list containing all tasks.
[...]

PR #1632, opened 2015-03-08
Support for EarlyTasks.

### LINKED ISSUE #1553: Improve tools for making an auto-generated post
Right now auto-generated posts or pages require two executions of `nikola build`. The reason for this is that scanning for posts is out of user control. So since the scan is done before the tasks are generated, the newly added files are missed. While the inconvenience of running build twice is small, a more robust machinery would still be good to have. Auto-generated posts are created by planetoid, and I needed an auto-generated page to create a publication list. For attempts to make a solution and more details see https://groups.google.com/forum/#!topic/nikola-discuss/J9k1sCM8VO0

felixfontein:
The earlytask_impl branch implements support for EarlyTasks in Nikola, which are tasks that are executed before Nikola scans for posts. See #1562 for more details. After some testing (sorry it took so long) I think it works rather well now, so it's time to start a PR :-)

What this branch improves over current Nikola:
- there's a better separation between Tasks and LateTasks: LateTasks won't be processed before Tasks aren't done; this makes in particular a difference when multithreading/multiprocessing is used;
- scanning for posts is now collected to happen at precisely one place: in stage 0, which is before the stage where all Task objects are processed, and after all EarlyTask objects are processed;

What this branch requires:
- an up-to-date version of doit

What this branch cannot deliver at the moment:
- `nikola build <filename>` does not work without further support in doit (see https://github.com/pydoit/doit/issues/20, haven't had time to really work on it yet);
- `nikola build <taskname>` only works for tasks named after the plugin itself, and not for (potential) otherwise named tasks created by the plugins;
- when running nikola in parallel (via `-n 4` for example), one has to tell doit (via `-P thread`) to use threads instead of processes.

(fixes #1553 and #1562)

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1632)


Kwpolska:
How *up-to-date* is up-to-date? Change the dependency in requirements.txt to match.

Also, can't we tell doit to use threads from code?

Kwpolska:
Also, the tests fail on Travis, which is a very bad thing you absolutely need to fix, especially because there are dependency errors in integration tests.

felixfontein:
As up-to-date as it can be ;-) I updated the requirement, but before merging we probably have to update it again.

Tell doit to use threads from code: yes, that's the plan. I'll check for a more elegant way later (instead of just inserting ```-P thread``` into the command line arguments).

Tests: they probably fail because the new post list rendering code requires the post to be already rendered. I updated the dependencies.

felixfontein:
Hmm, at the moment I don't see a completely satisfactory way to force the use of threads via code, except by injecting ```-P thread``` into the command line arguments. Overwriting `_execute` is not a good idea (in case the signature of doit's `Run._execute` every changes), but that's essentially the place where the default value comes from. @schettino72 or do I miss something?

schettino72:
It needs doit 0.27 or above. Apart from setting the minimum version on setup.py and requirements files, please also set [minversion](http://pydoit.org/cmd_run.html#minversion).

Please refresh my memory... Why parallel execution with multi-process doesnt work?

You can set (almost) any command line attribute on `DOIT_CONFIG`, you need to check the source code for the key name as it might be different from the command line parameter name. In your case it is:

```
DOIT_CONFIG = {
    'num_process': 4,
    'par_type': "thread",
}
````

felixfontein:
Ah, I forgot about ```DOIT_CONFIG```. I now added ```'par_type'``` as well as ```'minversion'```. Thanks for mentioning that!

About multi-processing: the implementation of multi-processing in doit requires to pickle task objects, which doesn't work well with closures (search for "pickled" in #1562). Since Nikola, default plugins and 3rd party plugins use closures in tasks, this could be somewhat problematic. At least for me, it always fails quite early when using multi-processing.

Kwpolska:
You set miversion incorrectly.  It should be `(0, 27, 0)` (a tuple and not a string)

felixfontein:
Hmm, this looks like a bug in doit (or a Python string problem) to me; after all, the documentation (http://pydoit.org/cmd_run.html#minversion) says you should use a string. I guess the problem is in ```version_tuple``` in ```doit/cmd_base.py```:

> ```parts = ver_in.split('.') if isinstance(ver_in, str) else ver_in```

in combination with:

> ```from __future__ import print_function, unicode_literals```

in ```nikola/__main__.py```, where later ```DOIT_CONFIG['minversion'] = '0.27.0'``` is set.

A simple fix (which ignores the main problem) would be to use a tuple, but that somewhat bothers me since it is not clear if that's officially allowed (the documentation only mentions strings, not tuples). @schettino72: should using tuples be officially allowed?

felixfontein:
(The problem doesn't happen with Python 3, BTW, that's why I never experienced it locally...)

schettino72:
It should support both string and tuple. The documentation on the site is incomplete, docs on source code are good: https://github.com/pydoit/doit/blob/master/doit/cmd_base.py#L14

And there is a bug on python2 it should check for `six.string_types` not for plain `str`.
https://github.com/pydoit/doit/blob/master/doit/cmd_base.py#L19

It is an easy fix (patches welcome), but I suggest you guys just use a tuple or you will have to wait for next doit release.

schettino72:
FYI. I fixed the the minversion problem. I also improved the docs saying both strings and tuples are supported. To be released on doit 0.28

https://github.com/pydoit/doit/commit/1b8f8b422906772a81f6c49c2935efc22e6e1763

felixfontein:
Thanks a lot! I'm now using a tuple.

ralsina:
So doit 0.28 is out. What's the status of this? Does it work? Does it have any limitations?

Kwpolska:
It cannot work yet, because we need to fix things to actually support doit v0.28.0.  See #1655.

felixfontein:
The main limitation is still that it is currently not possible to specify targets on the command line, like ```nikola build output/index.html```, and also it is not possible to specify task names which are different from their plugin's name. While I think the second might not be that important, the first one still bothers me quite a lot. How this will resolve depends a lot on what doit will do about it; the current discussion can be found here: https://github.com/pydoit/doit/pull/46

felixfontein:
I now rebased the earlytask_impl branch to current master; this also includes support for doit 0.28. (The old branch for doit 0.27 is preserved as https://github.com/getnikola/nikola/tree/earlytask_impl_old, just in case.)

Note that with doit 0.28, you still cannot specify targets via ```nikola build <targetname>```, but what you can do (at least with the current doit master) ```nikola build <taskname>:<targetname>```, for example ```nikola build render_tags:output/categories/demo.xml```. (Simply running ```nikola build render_tags``` works too, of course.)

ralsina:
@felixfontein in what ways does this break backwards compatibility? If it really does, then I'll do a 7.4.2 before merging it.

felixfontein:
Well, currently what doesn't work is ```nikola build <file name to build>```, i.e. specifying a target directly (see my comment from 3 days ago for more details). Also, running ```nikola build <name of task>``` doesn't work when ```<name of task>``` doesn't equals the plugin's name. Whether the first one will change (to be backwards compatible) depends on what will be implemented in doit; the second can work if the other task names are explicitly given to doit, there's no support for that implemented in `earltask_impl` at the moment.

There might be some other minor incompatibilities, in particular with other task plugins than the default ones.

Also, I'd prefer to rebase/merge with master again after when you merged the scanposts refactoring, since these two branches will have some conflicts.

ralsina:
The task/target specification limitations, while a change, I don't think is too bad. We could even check that  plugins generate tasks with the right name.

felixfontein:
Hmm, a problem I just noted: if a task has more than one target, there are two problems:
  a) if you write `nikola build <taskname>:<targetname>`, it won't find the task if you don't specify the first target;
  b) `nikola check -f` marks all other targets except the first one as an unknown file.
Both points also affect current Nikola. Since most task plugins never generate more than one file per task, this is hardly ever noticed. Also, one can get around of a) by leaving the task name away.
(In fact, I only noticed this since one of my plugins generates two files per task -- a HTML file and a CSS file --, and I was surprised to find the second file among the orphans.)

I don't know what to do about a) (@schettino72: any idea?). For b), this is currently a result because doit's `list --all` command returns a list of tasks in the form `<taskname>:<target>`, where `target` is the first target in the list if more than one is given. One has to manually call `doit info <taskname>:<target>` for every taskname and (first) target to get a list of all targets. Hmm, now that I think about it, both problems (a and b) are somewhat related.

ralsina:
I suppose nikola check could do something smarter than reading the output of the list command. It just never seemed necessary so far.

If you file a bug I'll get around to it eventually.

felixfontein:
Ok, I filed it as issue #1754.

schettino72:
@felixfontein I guess you have some misunderstanding about task names. All tasks have a `basename` and sub-tasks also have a `name`.  So the full task name is `<basename>:<name>`.
 
Since the name must be unique it is common to use the target path as  name. But of course that if you rely on getting the target from the name you will have problems if there is more than one target.
So there is no such thing as `<taskname>:<target>`...  

`doit list` has a parameter `template` but as of now it cant display `targets` that should be easy to be extended.

felixfontein:
@schettino72: I've noticed that this is a misunderstanding only yesterday :) This means that the current implementation of `nikola check` is pretty much based on a wrong assumption (which works fine in many cases, though).
`doit list` also showing targets is probably the simplest way to fix this (since most of the code of `check.py` doesn't need to be changed).

felixfontein:
But that also means that with current doit, that you cannot build a specific target with the `earlytask_impl` branch of Nikola. @ralsina,@Kwpolska: can you live with that? I'd be happy if I can just specify tasks (and not individual targets).

ralsina:
On 26/05/15 17:24, felixfontein wrote:
>
> But that also means that with current doit, that you cannot build a 
> specific target with the |earlytask_impl| branch of Nikola. @ralsina 
> <https://github.com/ralsina>,@Kwpolska <https://github.com/Kwpolska>: 
> can you live with that? I'd be happy if I can just specify tasks (and 
> not individual targets).
>

Telling nikola to build a specific file is IMHO an important thing.

Kwpolska:
@felixfontein Please do not bother me, I do not care about this branch and want this thread away from my inbox.  But specifying targets is important, and not having that is a no-go.

felixfontein:
The two new commits add partial support for target regexes as for https://github.com/pydoit/doit/pull/58.

felixfontein:
FYI: will now work again with the doit master branch. Building specific targets is still work in progress, though.

felixfontein:
Support for specifying targets is now included in doit master. As soon as a new version of doit is released, I'll adjust `requirements.txt` and do a rebase. Then hopefully the tests will succeed again as well :)

felixfontein:
Rebasing to current master.

felixfontein:
So, finally the task is ready for merging. A new doit version is released (0.29.0), and now all tests pass.

felixfontein:
I rebased with current master so that the `earlytask_impl` branch now has a thread-safe `LocaleBorg` (important for builds with more than one language).

felixfontein:
At least some of the current failures seem to be caused by the problem described in pydoit/doit#152. I'll take another look when that's fixed.

felixfontein:
Does anyone care to keep this? I've stopped using and maintaining this a long time ago (some eight months or so), and it doesn't work well anyway (since doit doesn't implement some required features).

Kwpolska:
:+1: for killing it

PR #1641, opened 2015-03-22
fix autorebuild

bricewge:
When using `nikola auto` the first build is successful but it doesn't rebuild the website on changes, it gave the following errors:
``
ERROR:root:[Errno 2] Aucun fichier ou dossier de ce type: 'nikolabuild'
ERROR:root:maybe you haven't installed nikolabuild
``
The command is missing a space before the arguments.


Kwpolska:
Thanks for your contribution!

PR #1643, opened 2015-03-23
Update pandoc.py

jboynyc:
Class `CompilePandoc` does not have an attribute `site`; it is part of the configuration.


Kwpolska:
We also need to make `self.site` a thing.

jboynyc:
Wow, that was fast!

Kwpolska:
It’s just an awake dev with new e-mail notifications.

And I also amended `master` with a `super()` call.

PR #1647, opened 2015-04-03
support tranforms to allow simpler restructuredtext extensions

mgaitan:
This patch allows Rest plugins to register `Transform` classes to be used in the NikolaReader instance. 

see https://github.com/getnikola/plugins/pull/73 as an example

### DIFF
--- a/nikola/plugins/compile/rest/__init__.py
+++ b/nikola/plugins/compile/rest/__init__.py
@@ -96,7 +96,7 @@ class CompileRest(PageCompiler):
                         'syntax_highlight': 'short',
                         'math_output': 'mathjax',
                         'template': default_template_path,
-                    }, logger=self.logger, source_path=source, l_add_ln=add_ln)
+                    }, logger=self.logger, source_path=source, l_add_ln=add_ln, site=self.site)
                 if not isinstance(output, unicode_str):
                     # To prevent some weird bugs here or there.
                     # Original issue: empty files.  `output` became a bytestring.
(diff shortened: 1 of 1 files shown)


nikola/nikola.py
```diff
@@ -241,7 +241,7 @@ def _enclosure(post, lang):
         return url, length, mime
 
 
-class Nikola(object):
+class nikola(object):
```
Kwpolska: NO NO NO

nikola/plugins/compile/rest/__init__.py
```diff
@@ -233,7 +241,7 @@ def rst2html(source, source_path=None, source_class=docutils.io.StringInput,
              parser=None, parser_name='restructuredtext', writer=None,
              writer_name='html', settings=None, settings_spec=None,
              settings_overrides=None, config_section=None,
-             enable_exit_status=None, logger=None, l_add_ln=0):
+             enable_exit_status=None, logger=None, l_add_ln=0, site=None):
```
Kwpolska: It would be better to just pass a `transforms=None` argument.  Also, `getattr` is unnecessary.

nikola/nikola.py
ralsina: PEP8 peeps.

Kwpolska: also, god knows what depends on this name not changing

Kwpolska:
Merged with some fixes (site.**rst_**transforms, also guaranteed to exist)

mgaitan:
Chris, let me say it: you are great!

On Tue, Apr 7, 2015 at 8:29 AM, Chris Warrick <notifications@github.com>
wrote:

> Merged with some fixes (site.*rst_*transforms, also guaranteed to exist)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/getnikola/nikola/pull/1647#issuecomment-90514085>.
>



-- 
mgaitan.github.io
textosypretextos.com.ar <http://textosyprextextos.com.ar>

PR #1649, opened 2015-04-04
Fixes `nikola check -l` in certain cases

hiliev:
Link checking is failing if `URL_TYPE` is set to `absolute` or `full_path`. In absolute mode, the wrong object is used to retrieve the path, while in full-path mode the leading `/` is not stripped (in addition to having the wrong object being used), therefore `os.path.join()` fails to prepend the output directory path.

### DIFF
--- a/nikola/plugins/command/check.py
+++ b/nikola/plugins/command/check.py
@@ -210,9 +210,9 @@ class CommandCheck(Command):
                 elif url_type in ('full_path', 'absolute'):
                     if url_type == 'absolute':
                         # convert to 'full_path' case, ie url relative to root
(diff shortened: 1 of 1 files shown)


hiliev:
Doesn't work with non-root deployments and `URL_TYPE = 'full_path'`.

Kwpolska:
You could’ve just fixed your patch and committed to the same repo without closing.

hiliev:
I realised that I haven't properly followed the guidelines and haven't created a branch. Dropping and forking again seemed the easiest way to do it properly (tm).

Kwpolska:
The branches rule is not mandatory for forks, it’s for people with access to this repo.

Issue #1650, opened 2015-04-05
Link checker broken for absolute and full-path URL types


hiliev:
It seems that the link checker (`nikola check -l`) doesn't work properly when `URL_TYPE` is set to `absolute` or `full_path`:

    $ nikola init --demo demo; cd demo
    ... change URL_TYPE to "full_path" or "absolute"
    $ nikola build
    $ nikola check -l
    Scanning posts....done!
    [2015-04-05T01:11:42Z] ERROR: check: Error with: output/2012/index.html 'str' object has no attribute 'path'
    [2015-04-05T01:11:42Z] ERROR: check: Error with: output/archive.html 'str' object has no attribute 'path'
    [2015-04-05T01:11:42Z] ERROR: check: Error with: output/galleries/demo/index.html 'str' object has no attribute 'path'
    ...

Nikola is vanilla `7.3.1` from `pip`, but `check.py` is the same in HEAD. I have a working patch in my forked repository that fixes it but being new to GitHub (and Git as a whole) I have a bit of trouble following the process outlined in `CONTRIBUTING`. Now that the issue is documented and open, shall I create pull request?

Kwpolska:
Yes, you should create a PR if you have a patch.  Also, make sure it works with every configuration, as per #1649.

hiliev:
I did check with most configurations imaginable. That's how #1651 turned up.

ralsina:
So, closing because it's fixed by #1652

PR #1652, opened 2015-04-05
Fixed link checker

hiliev:
This patch fixes the behaviour of `nikola check -l` for the cases where `URL_TYPE` is set to `full_path` or `absolute` (issue #1650). It also fixes the use of `SITE_URL` instead of `BASE_URL` when determining the relative portion of the path in the `absolute` case, which leads to bogus broken link reports when both variables are set in the configuration file and have different values.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,8 @@ Features
 Bugfixes
 --------
 
+* Fix ``nikola check -l`` for absolute and full-path URL styles
+  (Issue #1650)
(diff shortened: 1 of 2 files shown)


Kwpolska:
Please resolve conflicts. Read https://help.github.com/articles/configuring-a-remote-for-a-fork/ and https://help.github.com/articles/syncing-a-fork/ and merge `upstream/master`

hiliev:
I guess, I have to find a Windows system and really fix it.

Kwpolska:
Thanks for your contribution!

PR #1655, opened 2015-04-11
doit 0.28 release - breaking Nikola (or not)

Kwpolska:
I am planning to release doit 0.28 next week.

A) it would be good if you guys make a release with the doit pinned version before I release 0.28, this way at least new installation would get the correct doit version. is it possible?

B) I managed to get a patch[1] in doit that seems to be good enough to maintain backwards compatibility (at least for Nikola). But I didnt test much... note this is not on `master` branch.

Another option would be to make doit detect when Nikola is using the old API and give an error message like: "doit version higher than supported version. Please downgrade to doit 0.27 with: pip install doit=0.27"

IMO the second option is safer. If you guys prefer the patch I am also ok with, but please test it a bit.
So please let me what you guys think...

[1] https://github.com/pydoit/doit/commit/8666f7381939e5866a02435a5a0cb3538590c0aa


schettino72:
nikola/utils.py:Commands is broken too. This is not used by Nikola itself. Is it for plugins?

nikola/plugin_categories.py
```diff
@@ -93,6 +93,10 @@ def __init__(self, *args, **kwargs):
         BasePlugin.__init__(self, *args, **kwargs)
         DoitCommand.__init__(self)
 
+    def __call__(self, config=None, **kwargs):
```
schettino72: amazing hack :)

Looking at yapsy you could sub-class `PluginManager.instanciateElement()` to avoid this.

Kwpolska:
It’s a friendly way to run things via `nikola console`.  I’ll look into it later.

schettino72:
> > Now you guys won't even need to sub-class get_cmds()...

> And how would we tell doit about our custom commands?

Sorry. Ignore that. You would need to pass the entry point instead of the object. Doesnt make sense for Nikola because it already load it through plugins.

schettino72:
BUMP. Any expected date for a release? no pressure :grin:

Kwpolska:
@schettino72 So, doit==0.28.0 is out.  I don’t feel confident releasing anything right now (neither master pinned to 0.27.0 nor the new branch).  As such, I made an **emergency v7.3.1.1 release** that only changes the doit version and nothing else.

schettino72:
Yes. it is out. I added a notice if any program tries to use the old API (I tested it with Nikola). So at least the user gets a nice error message.

https://github.com/pydoit/doit/commit/0606570989c5c3702829391dba2a2cbe8e64bc6d

Kwpolska:
@schettino72 How can I get **instances** of doit-provided commands?  The `site.commands` proxy and `nikola check` both require them: `site.commands` needs to run `get_options()` (instance-bound), `nikola check` needs to talk to `nikola list` with the current context.

Kwpolska:
And for the record, current things that need to change:

- [ ] `nikola.utils.Commands` needs to iterate over `.get_cmds()` and use `v.get_options()`
- [ ] `nikola.plugins.command.check` needs to access `list` differently

Kwpolska:
@schettino72 Is this really the way to go?

```py
self._doitargs['cmds']['list'](config=self.config, **self._doitargs)
```

(`self._doitargs = **kwargs` in the `__call__` hack)

If yes, Commands would need to be generated by all plugins that need it (probably only `nikola console`) with a crazy incantation like this…

schettino72:
@Kwpolska look at the doit `help` command for an example.

1) get the list of command class in the constructor, (and also save the other kwargs parameters)
 - https://github.com/pydoit/doit/blob/master/doit/cmd_help.py#L102

2) Just create commands instances as you wish :)
 - https://github.com/pydoit/doit/blob/master/doit/cmd_help.py#L154

So, if you save the parameters with some nice names, it doesnt look so much like a crazy incantation :D

Kwpolska:
I’m more of a singleton fan, but okay.  I’m leaving the spaghetti in `check` and rewriting the Command thing.

Everything should be fine now.

Kwpolska:
Needs review. @schettino72 @ralsina?

nikola/utils.py
```diff
@@ -1305,34 +1305,51 @@ class Commands(object):
...
-    def __init__(self, main):
+    def __init__(self, main, config, doitargs):
         """Takes a main instance, works as wrapper for commands."""
         self._cmdnames = []
-        for k, v in main.get_commands().items():
-            self._cmdnames.append(k)
+        self._main = main
+        self._config = config
+        self._doitargs = doitargs
+        for k, v in self._main.get_cmds().items():
```
schettino72: You should use the `cmds` provided on `doitargs`, so please don't use `DoitMain.get_cmds()` .

schettino72: Note that `cmds` is a `PluginDict` [1], not a plain dict. You should get its items using `cmds.todict()`. Or if you need just one value, `cmds.get_plugin()`.

Plugins are lazy-loaded... for core commands plugins are not used so you get the command class without problems. But for plugins the dict value will only contain the entry point.


[1] https://github.com/pydoit/doit/blob/master/doit/plugin.py#L53

Kwpolska: fixed in b9f081e

Kwpolska:
Merging — I hope nothing will break.

PR #1658, opened 2015-04-17
CATEGORY_PAGES_DESCRIPTIONS correctly affect category pages

masayuko:
This patch fixes the bug ignoring CATEGORY_PAGES_DESCRIPTIONS values when CATEGORY_PAGES_ARE_INDEXES is set to False.


Kwpolska:
Good catch! Merging.

(The AppVeyor failure is not your fault, lxml has released a new version and there are no windows binaries yet)

PR #1660, opened 2015-04-18
FEED_LENGTH correctly affect the number of feeds in the main RSS

masayuko:
This patch fixes the bug that FEED_LENGTH value is ignored at the main RSS file if FEED_LENGTH is more than 10.


Kwpolska:
I’d rather replace `[:10]` with `[:kw['feed_length']]`.

Kwpolska:
Thanks!

PR #1663, opened 2015-04-22
Sanitize default IMAGE_FOLDERS output

Kwpolska:
Change the default (litter files in `/`) to something saner (`/images/`).

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,8 @@ Features
 Bugfixes
 --------
 
+* Make ``images`` the default output directory for IMAGE_FOLDERS
+  (Issue #1663)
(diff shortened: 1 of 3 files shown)


Kwpolska:
any opinions on this?  @ralsina, @pellenilsson?

pellenilsson:
Personally I use the `images` directory in the same way as the `files` directory, so if I want an image to end up in `/posts/example-post/assets/images/img.png` I would put it in `/images/posts/example-post/assets/images/img.png`. That way I can have the images sorted after what post they belong to.

But installing all images in `/images` constitutes another well ordered strategy, and the setting can always be overridden by the user, so I have no strong opinions on this one.

If changing this, you may want to edit the handbook under the Image Galleries header a bit (the fourth paragraph and the code block).

PR #1665, opened 2015-04-23
Support Latest posts

masayuko:
I would like to show a latest posts list on every post page.
Please consider this pull request.


felixfontein:
You need to add dependencies of every post page with the last 10 posts (or at least some basic info about them). Also note that this will recompile *every* post page when you add a new post, which takes quite a long time for large blogs.

For that reason, the default number of posts should be 0 and not 10, as 0 will not cause this.

Kwpolska:
No.

This is completely against Nikola’s principles.  If latest posts are displayed, you will basically kill all the benefits of automated rebuilds.

schettino72:
What if you get parts of the page using AJAX? You could build this once and inject this in the page only in the client side when the page is displayed...

Kwpolska:
You could easily achieve this by hand using the existing JSON for tag clouds, maybe with some code modifications.

ralsina:
Or fetching RSS and displaying that using a widget.

masayuko:
Thank you for reviewing my proposal.
I will consider another approach.

PR #1666, opened 2015-04-24
Wordpress code blocks (Fix #1186)

### LINKED ISSUE #1186: Wordpress importer doesn't translate [code]
In my Wordpress block, I made good use of their code blocks, which are delimited like 

```
[code]
code here
[/code]
```

or, to use a language

```
[code language="py"]
python_code_here
[/code]
```

(I think `language` can also be spelled just `lang`). 

If the wordpress importer uses Markdown, it could just translate `[code]` to \`\`\`, `[code lang='x']` to \`\`\`x, and enable fenced code blocks.

ralsina:


ralsina:
@asmeurer something is wrong here, can you please check why this test is not passing?

https://github.com/getnikola/nikola/blob/master/tests/test_command_import_wordpress.py#L318

I fixed others but that one looks like it should work and doesn't.

PR #1668, opened 2015-04-24
fix #1644 -- work around issues with IDNs

### LINKED ISSUE #1644: IDN SITE_URL is not converted to Punycode
The SITE_URL is not converted correctly to Punycode. For example, when initialising a new Blog and writing:

Site URL [http://getnikola.com/]: http://exämple.com/täst/
, this will result in conf.py to:
SITE_URL = "http://ex\u00e4mple.com/t\u00e4st/"

Correct should be that the domain name is converted to Punycode: 
SITE_URL = "http://xn--exmple-cua.com/t\u00e4st/"

The result is that for example Firefox throws an error when clicking on the logo.

I guess that (only) the domain part needs to be isolated from SITE_URL and then converted with "exämple.com".encode("idna") to xn--exmple-cua.com.

Nikola should also keep in mind that the user may edit the SITE_URL in conf.py directly and write the IDN without punycode directly, so for example:
SITE_URL = "http://exämple.com/täst"
Therefore the Punycode convert should best be applied while building, not in the blog init.

Kwpolska:
Encode IDNs to Punycode in ``nikola init`` and in links;
show an error if the site URL is not Punycode.


Kwpolska:
cc @ralsina @TheHellstorm

waiting for reviews.

`nikola init` output:

```
Site URL [http://getnikola.com/]: ąbć.pl/ó
    ERROR: You must specify a protocol (http or https).
Site URL [http://getnikola.com/]: http://ąbć.pl/ó
    The URL does not end in '/' -- adding it.
    Converting to Punycode: http://xn--b-vhah.pl/ó/
```

(also sneaked in `/` and `http(s)://` validation!)

`nikola build` output (if `http://ąbć.pl/ó/` is typed in manually):

```
[2015-04-24T17:21:00Z] ERROR: Nikola: Your BASE_URL or SITE_URL contains an IDN expressed in Unicode.  Please convert it to Punycode.
[2015-04-24T17:21:00Z] ERROR: Nikola: Punycode of ąbć.pl: xn--b-vhah.pl
```

Isso is not fixed, can’t do that easily (we would have to explicitly check for it, which is a lot of work — documented an impact of IDNs in the manual)

ralsina:
LGTM

TheHellstorm:
Thanks!

PR #1669, opened 2015-04-24
Replace MathJax with KaTeX

ralsina:
Not for merging, of course, but hey, let's play with it.


Kwpolska:
What happened to the `bootstrap` theme?!

ralsina:
Now that is what happens when you use jinjify wrong

Kwpolska:
now please fix the `bootstrap` theme (restore it to the way it is on `master` and patch for katex)

Also, you can now just type `scripts/jinjify.py` and it will do all the thinking for you.

ralsina:
done. This works for delimited math like using ``$$foo+bar$$ directly`` but doesn't pick up ``:math:`foo+bar``` or the math directive, that needs extra thinking.

Kwpolska:
We might need some special config for the auto-render plugin.

ralsina:
Now it works for math role but not for math directive for reasons I don't know.
Also, it has to iterate over all elements with math class, so base is broken because it has no JQuery and my JS sucks

Kwpolska:
Your hacky patch is unnecessary.  All we need to do is change the delimiters for auto-render.  (also, you forgot the CSS)

For example:

```
<script>
renderMathInElement(document.body, {delimiters: [
  {left: "\\begin{equation*}", right: "\\end{equation*}", display: true},
  {left: "\\[", right: "\\]", display: true},
  {left: "\\(", right: "\\)", display: false}
]})
</script>
```

We should make those configurable via the config file, because some weird people really want dollars (deprecated and error-prone) as delimiters.  We also need `$latex…$` for WordPress.com converts.

I’ll work on this.

Kwpolska:
This should be quite functional now, and KaTeX is now fully bundled — an empty site grew from 2.5 MB to 3.8 MB)  We still have a few references we might want to change.

We might also want to restore `mathjax.js` (without actually including it anywhere) for compatibility with old themes — @ralsina, any opinions?

cc @damianavila — do you know of any problems, other than having to upgrade all custom themes?

damianavila:
>cc @damianavila — do you know of any problems, other than having to upgrade all custom themes?

Not sure, I did not play with katex before, the notebook uses Mathjax so maybe there is some degree of incompatibility... if there is a possibility, I would leave the mathjax support, at least, as an option...

asmeurer:
Please don't make KaTeX the default. It has limited math support compared to mathjax (see https://github.com/Khan/KaTeX/issues/43).

ralsina:
This won't happen because of KaTeX limitations.

asmeurer:
It would be great to add KaTeX support, but make it an option alongside MathJax, and do it in a way that doesn't change existing sites unless you explicitly add something to conf.py (I don't personally care about the defaults for new sites).

Kwpolska:
It’s not easy to juggle two different systems. You can always add KaTeX to your own site via BODY_END and not using the `mathjax` tags.

przemoc:
It's worth to note that limitations in KaTeX are slowly vanishing. For instance [v0.4.0](https://github.com/Khan/KaTeX/releases/tag/v0.4.0) (released over a month ago):
> Added support for some \begin/\end environments, including support for array, matrix, pmatrix, bmatrix, vmatrix, and Vmatrix

and [v0.5.0](https://github.com/Khan/KaTeX/releases/tag/v0.5.0) (released a week ago):
> Add the cases and Bmatrix environments.

One day it may become good enough for becoming a default engine for math input in Nikola.

Kwpolska:
When that happens (and someone tells us about it!), we’ll happily include it. We are not qualified to tell it’s good or not…

Kwpolska:
Also, feature list: https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX

Issue #1676, opened 2015-05-03
CDN handling is not very smart


ralsina:
Currently, we have a USE_CDN flag and we do things like:

```
        %if use_cdn:
            <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>
            <script src="//maxcdn.bootstrapcdn.com/twitter-bootstrap/2.3.2/js/bootstrap.min.js"></script>
            <script src="/assets/js/all.js"></script>
        %else:
            <script src="/assets/js/all-nocdn.js"></script>
        %endif
```

That's why bundles looks like this. Things we want bundled when we don't use a CDN:

```
assets/js/all-nocdn.js=jquery.min.js,bootstrap.min.js,jquery.colorbox-min.js,moment-with-locales.min.js,fancydates.js
```

Things we want bundled when we _do_ use a CDN, because jquery and others are handled in the template:

```
assets/js/all.js=jquery.colorbox-min.js,moment-with-locales.min.js,fancydates.js
```

This is very fragile!

One possible improvement is that we only keep all.js and then, IF the USE_CDN flag is set, we smartly remove from  all.js the ones that are available in the CDN and we update a global context variable (say, CDN_JS_URLS) to contain them.

So the template would look like this:

```
        %if use_cdn:
               ${CDN_JS_URLS}
        %endif
            <script src="/assets/js/all-nocdn.js"></script>
```

This could be a problem if the ORDER of JS loading matters but that problem already exists :-)

---

All of the above is when we are using bundles. When not using bundles, use_cdn implies the template author has to do all the work again:

```
        %if use_cdn:
            <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.min.js"></script>
            <script src="//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js"></script>
        %else:
            <script src="/assets/js/jquery.min.js"></script>
            <script src="/assets/js/bootstrap.min.js"></script>
            <script src="/assets/js/moment-with-locales.min.js"></script>
            <script src="/assets/js/fancydates.js"></script>
        %endif
```

Here we could have a helper that takes a JS or CSS name as specified in the WIP branch and returns a local or CDN URL depending on use_cdn, so that the template, again, is simpler.

Kwpolska:
:-1: from me.  All those URLs are **theme**–specific, we can’t just put them in the global context.

ralsina:
[...]

PR #1678, opened 2015-05-03
WIP: Smarter cdn #1676

ralsina:
Try to be smarter and automatically use a CDN (or not) without having to do complex code in the templates.


ralsina:
Having https://github.com/getnikola/nikola/pull/1680 would make this possible by making sure bundles are built after ``copy_assets`` and before ``pages``

Kwpolska:
We could easily avoid dependency injection by using `cache/cdnjs` and `cache/cdncss` files that all pages would depend on.  The files would be created by the `bundles` plugin (specified as `targets`), and dependencies of all page builds.  (we could probably pass the lists from bundles to pages without reading the actual files on page generation)

scripts/get-cdnjs-list.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+
+import json
+import subprocess
+
+links = {}
+data = subprocess.check_output(['links', '-dump', 'https://cdnjs.com/'])
```
Kwpolska: We do not have to scrape the website.  There is a nice .json file we could use with a little parsing (might need an input format change to account for different files and versions): https://github.com/cdnjs/cdnjs/wiki/Extensions%2C-Plugins%2C-Resources#developers

ralsina:
I'd love @Aeyoun and @Kwpolska 's input on this one :-)

Currently only the bootstrap3 theme is adapted, but you can see how it gets simpler and more robust. Currently the mako template is not jinjify-able tho.

Kwpolska:
> Currently the mako template is not jinjify-able tho.

This is probably colorbox translations’ fault.  I suggest using string formatting, like this (untested):

```python
cdn_url('jquery.colorbox::i18n/jquery.colorbox-{lang}.js',
        use_cdn,
        '/assets/js/colorbox-i18n/jquery.colorbox-{lang}.js',
        lang=colorbox_locales[lang])
```

(using `**kwargs` and `str.format` on the Python side)

scripts/get-cdnjs-list.py
Aeyoun: There should be some config with an array of which libraries to include. Hard-coding a few into the default themes is not really flexible.

ralsina: @Aeyoun I don't understand. The list of available libraries cdnjs supports is quite extensive.

Aeyoun: @ralsina, but there is no setting to tell Nikola what libraries to load. Each library is hardcoded in the theme and not loaded based on a configurable setting.

ralsina: @Aeyoun I always assumed the theme author chooses what he loads...

ralsina:
This approach is not convincing to me anymore.

I'll explore alternatives.

PR #1690, opened 2015-05-05
Fix #1689

### LINKED ISSUE #1689: Very-old-format-metadata breaks
The really-old-metadata (untagged lines in a meta file) crashes.

ralsina:
Doh

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,11 @@
+New in master
+=============
+
+Bugfixes
+--------
+
+* Fix very old metadata format support (Issue #1689)
+
 New in v7.4.0
 =============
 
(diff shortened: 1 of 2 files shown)


Kwpolska:
The correct fix would be to remove `updated` from this list, because otherwise the labels won’t match up.  Applying that and merging.

cc @Aeyoun

ralsina:
or put updated last and use 8?

Kwpolska:
just upgrade to the new .meta format already

ralsina:
May I point out that *we* had not upgraded for *our* site? :-)

Kwpolska:
I’ll write a converter plugin tomorrow.

Issue #1715, opened 2015-05-12
Allow users to set DOIT_CONFIG options in conf.py


bnmnetp:
I want to relocate the doit database for my project, but I'm prone to forgetting the --db-file option.  I thought it would be nice to set that in conf.py.  It turns out you can set all sorts of doit options using DOIT_CONFIG but setting options to DOIT_CONFIG is ignored by Nikola __main__

PR #1716, opened 2015-05-12
Look for DOIT_CONFIG in config so it can be set in conf.py

bnmnetp:
Addresses feature request in #1715

### DIFF
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -62,7 +62,6 @@ _RETURN_DOITNIKOLA = False
 
 
 def main(args=None):
-
     colorful = False
     if sys.stderr.isatty() and os.name != 'nt':
         colorful = True
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -238,14 +238,18 @@ class NikolaTaskLoader(TaskLoader):
         self.quiet = quiet
 
     def load_tasks(self, cmd, opt_values, pos_args):
-        DOIT_CONFIG = config.get('DOIT_CONFIG', {})
         if self.quiet:
-            DOIT_CONFIG['verbosity'] = 0
-            DOIT_CONFIG['reporter'] = 'zero'
+            DOIT_CONFIG = {
+                'verbosity': 0,
+                'reporter': 'zero'
+            }
         else:
-            DOIT_CONFIG['reporter'] = ExecutedOnlyReporter
-            DOIT_CONFIG['outfile'] = sys.stderr
+            DOIT_CONFIG = {
+                'reporter': ExecutedOnlyReporter,
+                'outfile': sys.stderr
+            }
         DOIT_CONFIG['default_tasks'] = ['render_site', 'post_render']
+        DOIT_CONFIG.update(self.nikola._doit_config)
         tasks = generate_tasks(
             'render_site',
             self.nikola.gen_tasks('render_site', "Task", 'Group of tasks to render the site.'))
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -241,12 +241,12 @@ class NikolaTaskLoader(TaskLoader):
         if self.quiet:
             DOIT_CONFIG = {
                 'verbosity': 0,
-                'reporter': 'zero'
+                'reporter': 'zero',
             }
         else:
             DOIT_CONFIG = {
                 'reporter': ExecutedOnlyReporter,
-                'outfile': sys.stderr
+                'outfile': sys.stderr,
             }
         DOIT_CONFIG['default_tasks'] = ['render_site', 'post_render']
         DOIT_CONFIG.update(self.nikola._doit_config)
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -239,7 +239,7 @@ class NikolaTaskLoader(TaskLoader):
         self.quiet = quiet
 
     def load_tasks(self, cmd, opt_values, pos_args):
-        DOIT_CONFIG = config.get('DOIT_CONFIG',{})
+        DOIT_CONFIG = config.get('DOIT_CONFIG', {})
         if self.quiet:
             DOIT_CONFIG['verbosity'] = 0
             DOIT_CONFIG['reporter'] = 'zero'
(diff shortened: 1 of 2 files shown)


nikola/__main__.py
```diff
@@ -238,16 +239,16 @@ def __init__(self, nikola, quiet=False):
         self.quiet = quiet
 
     def load_tasks(self, cmd, opt_values, pos_args):
+        if 'DOIT_CONFIG' in config:
+            DOIT_CONFIG = config['DOIT_CONFIG']
+        else:
+            DOIT_CONFIG = {}
```
ralsina: It's easier to just do DOIT_CONFIG = config.get('DOIT_CONFIG', {})

bnmnetp: Indeed it is.  I have this habit of writing all of my code like I'm preparing to show it in my CS1 course.

nikola/__main__.py
```diff
@@ -62,6 +62,7 @@
 
 
 def main(args=None):
+
```
Kwpolska: Don’t put empty lines at the beginnings of functions.

nikola/__main__.py
```diff
@@ -238,16 +238,13 @@ def __init__(self, nikola, quiet=False):
...
-                'reporter': 'zero',
-            }
+            DOIT_CONFIG['verbosity'] = 0
+            DOIT_CONFIG['reporter'] = 'zero'
         else:
-            DOIT_CONFIG = {
-                'reporter': ExecutedOnlyReporter,
-                'outfile': sys.stderr,
-            }
+            DOIT_CONFIG['reporter'] = ExecutedOnlyReporter
```
Kwpolska: Wouldn’t it be nicer to let users’ `DOIT_CONFIG` override this?  Some people might want to override our opinionated reporter choice.

```py
if self.quiet:
    DOIT_CONFIG = {…}  # from original code
else:
    DOIT_CONFIG = {…}

DOIT_CONFIG.update(config.get('DOIT_CONFIG', {}))
```

We also need to [pop `DOIT_CONFIG` out of the config dict](https://github.com/getnikola/nikola/blob/master/nikola/nikola.py#L285) in `nikola.py` (we don’t need to save it anywhere though) so `self.configured` works correctly.

nikola/__main__.py
bnmnetp: Removed.  I guess I didn't clean up properly after removing some debug prints.

nikola/__main__.py
bnmnetp: updating the dictionary as you suggest makes sense to me.  I hadn't thought that you guys would want anyone to override what you had coded in there.

Since  ``DOIT_CONFIG`` may or may not be in config.   Why don't I just check and pop it out of config in the ``load_tasks`` function?

bnmnetp: So, it seems like going back to the original and adding this one line:

```
        DOIT_CONFIG.update(config.pop('DOIT_CONFIG', {}))
```
would do what you want.

Kwpolska: `load_tasks` is called after configuration is passed to the site object.  We can’t do that.

bnmnetp: So then we can't just pop it out and throw it away in ``__init__``  It has to stay somewhere so that ``load_tasks`` can get to it.

Kwpolska: Then pop it out in `Nikola.__init__` and use `nikola._doit_config` here. (besides, where does `config` come from in this code?)

bnmnetp: config is declared as a global on line 58 of ``__main__``  and then populated on line 114 in ``main``  after loading conf.py

```
        config = conf.__dict__
```

Kwpolska: ah, a global.  That’s evil.

Still, you can just get it from `self.nikola` if you pop it out there.

bnmnetp: I tried:

```
AttributeError: 'Nikola' object has no attribute '_doit_config'
```

Kwpolska: Yes, that’s what I was suggesting. Add after [`nikola.py` line 287](https://github.com/getnikola/nikola/blob/master/nikola/nikola.py#L287):

```py
        self._doit_config = config.pop('DOIT_CONFIG', {})

**Comment by schettino72** [comment]
I guess normal Nikola users might not know about doit at all. So maybe better expose it as a different name than `DOIT_CONFIG`.

**Comment by Kwpolska** [comment]
“Normal Nikola users” should not *know* about this setting.

**Comment by schettino72** [comment]
> “Normal Nikola users” should not know about this setting.

why not? This allow users to change default from command line options. 
@bnmnetp himself wants to change... you also mention about changing the reporter...

**Comment by bnmnetp** [comment]
configuring a value for a command line option certainly felt to me like a thing "normal" users may want to do.
@schettino72 are you suggesting that in conf.py we expose this under a different name than DOIT_CONFIG?  maybe CL_PARAMETERS  or something more descriptive?   Although I have no idea how many of the command line options exposed are configurable through doit and how many are not.

**Comment by Kwpolska** [comment]
Is DOIT_CONFIG even documented by doit itself?

**Comment by bnmnetp** [comment]
I found it here:  http://pydoit.org/cmd_run.html

I don't spend my whole day looking through source :-)

File: nikola/__main__.py [hunk 4]
```diff
@@ -241,14 +241,15 @@ def load_tasks(self, cmd, opt_values, pos_args):
         if self.quiet:
             DOIT_CONFIG = {
                 'verbosity': 0,
-                'reporter': 'zero',
+                'reporter': 'zero'
```
ralsina: No need to change it, but I tend to leave the last commas in this sort of things because it makes the diff better the next time something is added :-)

ralsina:
It looks good enough for me. Add yourself to AUTHORS?

nikola/__main__.py
Kwpolska: (It also lets the next person just type in their changes without having to care about adding a comma.)

schettino72:
@Kwpolska it is documented here: http://pydoit.org/configuration.html#configuration-at-dodo-py

@bnmnetp every command line option can be configured (and also some options that are not available through the command line)

PR #1717, opened 2015-05-12
fix 1251

ralsina:
I am not 100% sure this is a good idea, but it *should* fix the thing. (No it doesn't)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
+* Ensure galleries are built after the posts task (Issue #1251)
```
Kwpolska: wrong placement

ralsina:
It's best to fix this on the gallery_directive plugin.

PR #1719, opened 2015-05-12
fix #1688

### LINKED ISSUE #1688: post-list makes posts always out-of-date
Ideally: they would only be rebuilt when the actual list of posts they display, or the content of any of those posts, changes.

Practically: they should depend on the whole timeline not changing, at least react to post addition/removal.

ralsina:
This makes the post-list directive only trigger a rebuild of a post if the timeline changes.

OTOH:

1) It makes the posts plugin a bit more dense
2) I suspect this is racy as hell in multicore builds (which I hope noone does since I never worried before ;)


Kwpolska:
I’ve heard some people do multi-core builds.  What would crash for them?

ralsina:
Not crash, but limited testing shows some times not everything will be built. That already happens with sitemap, for example, this is building from scratch using -n8:

```
.  render_tags:output/categories/cat_nikola.xml
.  render_pages:output/posts/foo16.html
.  sitemap:output/sitemap.xml
.  sitemap:output/sitemapindex.xml
.  robots_file:output/robots.txt
(nikola)ralsina@yoga:~/Desktop/proyectos/nikola/master/foo$ 
(nikola)ralsina@yoga:~/Desktop/proyectos/nikola/master/foo$ nikola build -n 8
Scanning posts....done!
.  sitemap:output/sitemap.xml
.  sitemap:output/sitemapindex.xml
(nikola)ralsina@yoga:~/Desktop/proyectos/nikola/master/foo$ nikola build -n 8
Scanning posts....done!
(nikola)ralsina@yoga:~/Desktop/proyectos/nikola/master/foo$ 
```

You can see how the 1st run left the sitemap unbuilt, and then the 2nd one does it.

PR #1720, opened 2015-05-13
Emit an event when a post/page already exists, also.

punchagan:
Creation of new_post successfully emits an event, but no event is fired
new_post fails because there is an existing post.  This event can be
used by plugins, for e.g. open the post in an editor or prompt to delete
the post, etc.


Kwpolska:
Nobody will use this anyways, but it shouldn’t hurt.  Merging.

PR #1731, opened 2015-05-19

Kwpolska:


ralsina:
Should we update this with the fixed version? Maybe >=2.7.0 instead?

Kwpolska:
did you mean `Yapsy>=1.11.123`?  test dependencies are flagged down at exact versions.

ralsina:
I meant pytest, but if we have them on fixed numbers, that's ok for me.

Yapsy let's wait for the next minor release that will have the fix and a consistent version number...

ralsina:
We could go back to 1.10.143 or something like that until the new release is out, that one works.

Or 1.11.123b :-P

PR #1734, opened 2015-05-19
New auto

### LINKED ISSUE #1729: auto doesn't rebuild in some situations
- [ ] When there's a tag collision
- [ ] When adding a new post via command line

ralsina:
This replaces livereload with a homebrew version. It seems to fix #1729 and #1757 to me, and also adds a nice thing in that it shows build failures in the web page as an alert (may be too annoying ;-)

Missing bits:

- [x] Cleanup
- [x] Support -b
- [x] Support -a
- [x] Error out on missing deps
- [x] Use re to inject JS instead of replace


Kwpolska:
suggestion:

```
mv auto.plugin ../
mv auto.py __init__.py
```

Will make things a bit cleaner.

nikola/plugins/command/auto/__init__.py
```diff
@@ -0,0 +1,302 @@
...
+            with open(LRJS_PATH) as fd:
+                start_response(b'200 OK', [(b'Content-type', mimetype)])
+                return self.inject_js(mimetype, fd.read())
+        start_response(b'404 ERR', [])
+        return ['404 {0}'.format(uri)]
+
+    def inject_js(self, mimetype, data):
+        """Inject livereload.js in HTML files."""
+        if mimetype == 'text/html':
+            data = re.sub('</head>', self.snippet, data, 1, re.IGNORECASE)
```
Aeyoun: `</body>`

Note that HTML5 do not guarantee any element. It requires doctype and title only.

nikola/plugins/command/auto/__init__.py
```diff
@@ -0,0 +1,302 @@
...
+        if os.path.isfile(f_path):
+            with open(f_path) as fd:
+                start_response(b'200 OK', [(b'Content-type', mimetype)])
+                return self.inject_js(mimetype, fd.read())
+        elif p_uri.path == '/livereload.js':
+            with open(LRJS_PATH) as fd:
+                start_response(b'200 OK', [(b'Content-type', mimetype)])
+                return self.inject_js(mimetype, fd.read())
+        start_response(b'404 ERR', [])
+        return ['404 {0}'.format(uri)]
```
Aeyoun: Why not inject live reload here also? User could be fiddling with their slugs or other path manipulation.

nikola/plugins/command/auto/__init__.py
```diff
@@ -0,0 +1,302 @@
...
+        self.logger.info("Serving on port {0}...".format(port))
+
+        # Yes, this is racy
+        if browser:
+            webbrowser.open('http://{0}:{1}'.format(host, port))
+
+        try:
+            ws.serve_forever()
+        except KeyboardInterrupt:
+            ws.server_close()
```
Aeyoun: Log this [like in the serve](https://github.com/getnikola/nikola/blob/master/nikola/plugins/command/serve.py#L123) plugin.

nikola/plugins/command/auto/__init__.py
```diff
@@ -0,0 +1,302 @@
...
+            'default': 8000,
+            'type': int,
+            'help': 'Port nummber (default: 8000)',
+        },
+        {
+            'name': 'address',
+            'short': 'a',
+            'long': 'address',
+            'type': str,
+            'default': '0.0.0.0',
```
Aeyoun: Time for `::` by default? ;)

Kwpolska: blocker: https://github.com/livereload/livereload-js/pull/39

nikola/plugins/command/auto/__init__.py
ralsina: Good idea. Adding.

nikola/plugins/command/auto/__init__.py
ralsina: I am switching to 127.0.0.1 because 0.0.0.0 acts weird in chrome: editing the path makes it a google search!

nikola/plugins/command/auto/__init__.py
ralsina: I don't understand the suggestion. How would you inject it?

ralsina:
@aeyoun @Kwpolska so ... any pending concerns with this branch?

nikola/plugins/command/auto/__init__.py
```diff
@@ -0,0 +1,320 @@
...
+try:
+    from urlparse import urlparse
+except ImportError:
+    from urllib.parse import urlparse  # NOQA
+import webbrowser
+from wsgiref.simple_server import make_server
+import wsgiref.util
+
+from blinker import signal
+import pyinotify
```
Kwpolska: This works on Linux only and crashes on other platforms:

```pytb
  File "C:\Python27\lib\site-packages\yapsy\PluginManager.py", line 484, in loadPlugins
    candidate_module = imp.load_module(plugin_module_name,None,candidate_filepath,("py","r",imp.PKG_DIRECTORY))
  File "C:\projects\nikola\nikola\plugins\command\auto\__init__.py", line 43, in <module>
    import pyinotify
ImportError: No module named pyinotify
```

You would have to find another library, like [watchdog](http://pythonhosted.org/watchdog/).

Issue #1737, opened 2015-05-20
Automate symlinks into bower


ralsina:
As bower components add/remove files, symlinks in nikola/data need adding/removing.

This is currently done manually, and it's a good idea to automate it.

Kwpolska:
fixed in #1739

PR #1739, opened 2015-05-20
Auto bower

ralsina:
Script to update and symlink bower dependencies (Issue #1737)


Kwpolska:
It should also take care of jinja themes.

ralsina:
I think jinjify should handle that, instead.

ralsina:
@Kwpolska what do you think about merging this as a 1st iteration?

scripts/update-bower.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Update all bower packages
+bower update
+
+# Link bootstrap3 theme to bootstrap
+pushd nikola/data/themes/bootstrap3/assets/js/
+ln -sf ../../../../../../bower_components/bootstrap/dist/js/*js .
+rm npm.js
+git add *js
```
Kwpolska: `git add *` can be problematic.  use `git add .` to add everything in the current working directory.

Kwpolska:
Change the `git add` command, add `git commit -am 'bower update'` and merge.

Kwpolska:
Done.  I also added Jinja asset updates for the two bootstrap themes to `jinjify.py` (done via copying).  Merging.

Issue #1740, opened 2015-05-20
`nikola status`


Aeyoun:
Just thinking out loud here …

    $ nikola status
    The last build succeeded.
    2 days and 6 hours since last deployment.
    48 files have changed (including 1 new post) since then.
    2 scheduled post (next in 58 hours), and 2 drafts.
    You’re running the latest version of Nikola.

Any interest? (I keep mistyping “git status” and though “Hey, Nikola could do something useful here.”)

ralsina:
Sounds interesting.

duck:
+1

PR #1741, opened 2015-05-21
New status command

Aeyoun:
From issue #1740.

Sample output:

    39 output files modified since last deployment 4 hours and 31 minutes ago.
    4 days and 22 hours to next scheduled post.
    78 posts in total, 1 scheduled, and 0 drafts.

Counting new posts is out. Tracking whether last build succeeded seemed less useful when I started working  on it, so I dropped that too. Checking for version should be done in version plugin.


nikola/plugins/command/status.py
```diff
@@ -0,0 +1,106 @@
...
+from dateutil.tz import gettz, tzlocal
+
+from nikola.plugin_categories import Command
+
+
+class CommandDeploy(Command):
+    """ Site status. """
+    name = "status"
+
+    doc_usage = "[[preset [preset...]]"
```
Kwpolska: not really, at least not in this instance.

nikola/plugins/command/status.py
```diff
@@ -0,0 +1,106 @@
...
+    doc_purpose = "display site status"
+    doc_description = "Show information about the posts and site deployment."
+    logger = None
+
+    def _execute(self, command, args):
+
+        self.site.scan_posts()
+
+        timestamp_path = os.path.join(self.site.config["CACHE_FOLDER"], "lastdeploy")
```
Kwpolska: On a side note, does `github_deploy` write to this file?  If not, it should.

nikola/plugins/command/status.py
```diff
@@ -0,0 +1,106 @@
...
+            with io.open(timestamp_path, "r", encoding="utf8") as inf:
+                last_deploy = datetime.strptime(inf.read().strip(), "%Y-%m-%dT%H:%M:%S.%f")
+                last_deploy_offset = datetime.utcnow() - last_deploy
+        except (IOError, Exception):
+            print("It does not seem like you’ve ever deployed the site (or cache missing).")
+
+        if last_deploy:
+
+            if last_deploy_offset.days > 0:
+                last_deploy_offsetstr = "{0} days and {1} hours".format(str(int(last_deploy_offset.days)), str(int(last_deploy_offset.seconds / 60 / 60)))
```
Kwpolska: Why do you need this `str(int())` dance?

Just use `{0:.0f} days and {1:.0f} hours` as the format string.  It will guarantee zero decimal places.  Even if not for this, str.format() does not need strings as its arguments — and it’s even better to give ints and floats so you can format them more precisely.

nikola/plugins/command/status.py
```diff
@@ -0,0 +1,106 @@
...
+                posts_drafts = posts_drafts + 1
+            if post.publish_later:
+                posts_scheduled = posts_scheduled + 1
+                post_due_offset = post.date - datetime.utcnow().replace(tzinfo=gettz("UTC"))
+                if (post_scheduled_nearest_offset is None) or (post_due_offset.seconds < post_scheduled_nearest_offset.seconds):
+                    post_scheduled_nearest_offset = post_due_offset
+
+        if posts_scheduled > 0 and post_scheduled_nearest_offset is not None:
+            if post_scheduled_nearest_offset.days > 0:
+                nearest_scheduled_timestr = "{0} days and {1} hours".format(str(int(post_scheduled_nearest_offset.days)), str(int(post_scheduled_nearest_offset.seconds / 60 / 60)))
```
Kwpolska: Deduplication: write a `display_time` function that takes `offset` as an argument and is used for generating pretty `last_deploy_offset` and `post_scheduled_nearest_offset` text

nikola/plugins/command/status.py
```diff
@@ -0,0 +1,106 @@
...
+
+            elif post_scheduled_nearest_offset.seconds / 60 / 60 > 0:
+                nearest_scheduled_timestr = "{0} hours and {1} minutes".format(str(int(post_scheduled_nearest_offset.seconds / 60 / 60)), str(int(post_scheduled_nearest_offset.seconds / 60 / 60)), str(int(((post_scheduled_nearest_offset.seconds / 60) // 60) * 60)))
+            else:
+                nearest_scheduled_timestr = "{0} minutes".format(str(int(((post_scheduled_nearest_offset.seconds / 60) // 60) * 60)))
+            print("{0} to next scheduled post.".format(nearest_scheduled_timestr))
+
+        print("{0:,} posts in total, {1:,} scheduled, and {2:,} drafts.".format(posts_count, posts_scheduled, posts_drafts))
```
Kwpolska: Sounds better if we used plain numbers (123456 vs 123,456) here.

nikola/plugins/command/status.py
Aeyoun: It does now.

PR #1745, opened 2015-05-23
fix #1710 -- add a rst2html command

### LINKED ISSUE #1710: Do a rst2html-like tool
Maybe not supporting the whole rst2html command line, BUT! If we can:

1) Make it load rst extension plugins
2) Do the basic stuff, like compiling files

Then it will be very useful. Specially if we get some/all of the sphinx compatibility to work.

Kwpolska:
First draft.  Nikola extensions (that don’t need access to the timeline) seem to work fine.

cc @ralsina

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,8 +4,9 @@ New in master
 Features
 --------
 
+* New `nikola rst2html` command (Issue #1710)
+* New `nikola status` command (Issue #1740)
 * Support [code] in wordpress importers (Issue #1186)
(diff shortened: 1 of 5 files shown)


ralsina:
Awesome!

Kwpolska:
Note that this is a very early draft that can't do much. It could use some
more features. And it's probably useless for Windows users, because it
expects a Unicode-capable terminal — that could bee fixed by a file writing
option. We also force a specific template that inlines rst.css.

-- 

PR #1750, opened 2015-05-23
give meaningful error if two posts conflict (Fix #1749)

### LINKED ISSUE #1749: No useful error when two posts have same slug
As noticed in #1748, if two posts (or pages) in the same output directory have the same slug, no good error message is given. (Usually doit will emit an error since two tasks generate the same target.)

I think it would be a good idea if after scanning all posts, a check for duplicate slugs in the same directory is performed and warnings are emitted if such situations occurr. (I'm hesitant to make this an error since depending on where -- and by whom -- things are written to in the output, the tags might not collide in the end.) What do you think about that?

ralsina:
Give a meaningful error and exit when slugs conflict.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1463,8 +1463,22 @@ class Nikola(object):
                 self.pages.append(post)
 
             for lang in self.config['TRANSLATIONS'].keys():
-                self.post_per_file[post.destination_path(lang=lang)] = post
-                self.post_per_file[post.destination_path(lang=lang, extension=post.source_ext())] = post
+                dest = post.destination_path(lang=lang)
+                src_dest = post.destination_path(lang=lang, extension=post.source_ext())
+                if dest in self.post_per_file:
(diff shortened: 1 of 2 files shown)


nikola/nikola.py
```diff
@@ -1463,8 +1463,22 @@ def scan_posts(self, really=False, ignore_quit=False, quiet=False):
...
+                        dest,
+                        self.post_per_file[dest].source_path,
+                        post.source_path))
+                    sys.exit(1)
+                if (src_dest in self.post_per_file) and self.config['COPY_SOURCES']:
+                    utils.LOGGER.error('Two posts are trying to generate {0}: {1} and {2}'.format(
+                        src_dest,
+                        self.post_per_file[dest].source_path,
+                        post.source_path))
+                    sys.exit(1)
```
Kwpolska: `quit = True` so we can show all bad posts at once

ralsina: Done.

felixfontein:
Thanks for implementing!

PR #1759, opened 2015-05-27
Smarter check

### LINKED ISSUE #1758: Nikola check has a fragile target list
As described in #1632, nikola check gets the list of target files by parsing ``nikola list --all`` which means it assumes a lot of things.

Instead, it should obtain them from doit internals.

ralsina:
Fix #1758 and hopefully breaks nothing ;-)


ralsina:
@felixfontein I'd like your review of this one

nikola/plugins/command/check.py
```diff
@@ -44,33 +45,23 @@
...
 
 
-def _call_nikola_list(l, site, arguments):
-    class NotReallyAStream(object):
-        """A massive hack."""
-        out = []
+def _call_nikola_list(site):
+    files = []
+    for task in generate_tasks('render_site', site.gen_tasks('render_site', "Task", '')):
+        files.extend(task.targets)
```
felixfontein: This won't work if any of the generated tasks has a delayed task loader, but a) probably no plugin author does that, and b) I have to go through that anyway when adjusting this for the earlytask_impl branch, so I vote for leaving this as is for the moment.

nikola/plugins/command/check.py
```diff
@@ -44,33 +45,23 @@
...
-    try:
-        l.outstream = newstream
-        l.parse_execute(arguments)
-        return newstream.out
-    finally:
-        l.outstream = oldstream
-
-
-def real_scan_files(l, site):
+def real_scan_files(site):
```
felixfontein: This function is also called from `orphans.py` and `github_deploy.py`.

nikola/plugins/command/check.py
```diff
@@ -44,33 +45,23 @@
...
     real_fnames = set([])
     output_folder = site.config['OUTPUT_FOLDER']
     # First check that all targets are generated in the right places
-    for task in _call_nikola_list(l, site, ["--all"]):
-        task = task.strip()
-        if output_folder in task and ':' in task:
-            fname = task.split(':', 1)[-1]
+    for fname in _call_nikola_list(site):
+        fname = fname.strip()
+        if output_folder in fname:
```
felixfontein: `startwith()` is better than `in` in this context, as maybe some file `cache/bla/output/bla` is generated, which also contains `output`.

nikola/plugins/command/check.py
```diff
@@ -262,7 +252,7 @@ def analyze(self, task, find_sources=False, check_remote=False):
                         self.logger.warn("Broken link in {0}: {1}".format(filename, target))
                         if find_sources:
                             self.logger.warn("Possible sources:")
-                            self.logger.warn("\n".join(_call_nikola_list(self.l, self.site, ["--deps", task])))
+                            self.logger.warn("\n".join(_call_nikola_list(self.site)))
```
felixfontein: The output of `nikola list --all --deps <taskname>` is quite different than the one from `nikola list --all`, so this is not a valid substitution. Actually, I think we need some support from doit here, if we don't want to repeat some of its code or use internal APIs. @schettino72: what do you think (or does that already exist?) about having a simple way in doit to get essentially the output of `doit list --all --deps` (including a list of targets per task) in a nice way?

nikola/plugins/command/check.py
```diff
@@ -273,14 +263,11 @@ def scan_links(self, find_sources=False, check_remote=False):
...
-            if task.split(':')[0] in (
-                    'render_tags', 'render_archive',
-                    'render_galleries', 'render_indexes',
-                    'render_pages', 'render_posts',
-                    'render_site') and '.html' in task:
-                if self.analyze(task, find_sources, check_remote):
+        # Maybe we should just examine all HTML files
+        output_folder = self.site.config['OUTPUT_FOLDER']
+        for fname in _call_nikola_list(self.site):
+            if output_folder in fname and '.html' in fname:
```
felixfontein: What about using `endswith()` here instead of (the second) `in`? Just in case someone likes to use `.html` in directory names :)

nikola/plugins/command/check.py
ralsina: Done

nikola/plugins/command/check.py
ralsina: Ack.

nikola/plugins/command/check.py
ralsina: Done

nikola/plugins/command/check.py
ralsina: Good catch. Same thing as before, I just need to check task.file_dep, no big deal.

ralsina: Done.

nikola/plugins/command/check.py
ralsina: well [-5:] but same idea :-)

nikola/plugins/command/check.py
schettino72: As I said before, the `doit list` supports a `template` param to format its output (it seems I didnt document this parameter - sorry). Currently the template can only output values for: `name`, `status` and `doc`. I guess the best option would be to extend supported values that can appear in the template. Or maybe support a json output?

There is also the `doit info` command that outputs the targets and dependencies but lacks a formatting and a `--all` options...

nikola/plugins/command/check.py
```diff
@@ -44,33 +46,26 @@
...
 
 
-def _call_nikola_list(l, site, arguments):
-    class NotReallyAStream(object):
-        """A massive hack."""
-        out = []
+def _call_nikola_list(site):
+    files = []
+    deps = defaultdict(list)
+    for task in generate_tasks('render_site', site.gen_tasks('render_site', "Task", '')):
```
Kwpolska: Is this safe?  Are we not doing it twice?

We should probably talk to our task loader anyways.

ralsina: It works anyway. To talk to our taskloader we'd have to move it out of __main__ first.

Kwpolska: You can just talk to `self.site.doit.task_loader`.  You also can save the generated tasks there, if any.

ralsina: I tried it, see no difference, and I don't understand internals well enough, maybe.

So, proposing to merge this as-is.

nikola/plugins/command/check.py
felixfontein: I'd prefer JSON output; that's easy to parse and one can also easily add more info without breaking existing clients.

felixfontein:
Before merging: what is that problem with the `nikola check -l` failure in the tests running on AppVeyor? Is it something Windows specific?

nikola/plugins/command/check.py
```diff
@@ -44,33 +46,26 @@
...
+        files.extend(task.targets)
+        for target in task.targets:
+            deps[target].extend(task.file_dep)
+    for task in generate_tasks('post_render', site.gen_tasks('render_site', "LateTask", '')):
+        files.extend(task.targets)
```
felixfontein: Don't you need the lines updating `deps[target]` here as well?

ralsina:
The output of the demo site changes because I fixed a broken link generated by the listings plugin, I think.

nikola/plugins/command/check.py
ralsina: You are correct. Good catch! Fixing...

felixfontein:
Ah ok. Then everything's fine from my POV :)

Issue #1764, opened 2015-05-28
Make galleries (and maybe slides?) work with base theme


ralsina:
It's a bit shameful to ship a "major" feature like galleries that doesn't work at all with the base theme, from which we recommend theme authors should fork.

This will probably be a lot of work, but I am calling it a blocker.
- [x] galleries WIP in #1765
- [x] lightbox
- [ ] slides TBD

ralsina:
I'll tackle this as part of #1623 starting tonight.

ralsina:
I now think slides don't belong in core, and should move to a plugin.

Issue #1623, opened 2015-02-11
Refactor gallery into a JSON+JS implementation


ralsina:
The current solution has a bunch of problems:

1) It's JQuery-dependent
2) It uses a largely abandoned library
3) It's complex code
4) It doesn't even work in the base theme

So, how to make this easier:

1) Simplify the plugin to make it just generate a json file with the correct image data.
2) Write, in the base theme, a _very_ simplistic loader of that data that lists the images
3) Write in the bootstrap theme (which guarantees JQuery) something nicer, based on whatever JQuery gallery thing we find

This plan would move the complexity out of core and into the templates, which is where we probably want it since:

1) The gallery display code is JS anyway
2) The theme authors are the ones who may want to hack that
3) Makes the gallery directive a much simpler thing

Plan:
- [x] Generate JSON file with gallery image info
- [x] Create by hand HTML file that can display said thing 
- [x] Templatize
- [x] Remove useless template inputs
- [x] Remove code that generates removed inputs

Kwpolska:
Let’s not use fancy schmancy *experiments*.  That’s what happened with this one.  Let’s look for something **tried and true**; hell, we could even make something ourselves with Bootstrap’s `<img class="img-thumbnail">` and [the tooltip JS](http://getbootstrap.com/javascript/#tooltips).

Oh, this one is also abandoned: https://github.com/codrops/GammaGallery (last commit December **2012**)

devurandom:
I'll vote for anything that stabilises the gallery_directive, so you can remove the "horrible, horrible hack" note from the plugin. :)

StyXman:
Could you include #2146 in this?

ralsina:
This is actually done!

PR #1765, opened 2015-05-28
Implement Galleries as JSON+JS for all themes

ralsina:
Working on #1764 and #1623

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1765)


ralsina:
@Aeyoun this adds JS galleries to base... is that ok with you? Base is your baby :-)

The main goal is to let all the other themes just inherit gallery.tmpl and not have to tweak it.

docs/creating-a-theme.txt
```diff
@@ -1,384 +1,767 @@
...
-        ${bootstrap.html_head()}
-        <%block name="extra_head">
-        </%block>
-        ${extra_head_data}
+    <meta charset="utf-8">
+    <meta name="description" content="This is a demo site for Nikola.">
+    <meta name="viewport" content="width=device-width">
+    <title>My Nikola Site | My Nikola Site</title>
+
+    <link href="assets/css/rst.css" rel="stylesheet" type="text/css">
```
Aeyoun: Type is not needed for text/css.

ralsina: I thought some validator or other complained if it was not there?

ralsina: In any case, I have no idea why this file is in this diff. It's exactly the same in master :-(

Kwpolska:
1. The way you load jQuery and flowr is hacky, and will not make `USE_CDN=False` users happy; it would be easier to just put the JS in the usual place for all pages (besides, I already sneaked some JS into base when no-one looked)
2. The galleries are now **completely inaccessible** to people with JavaScript disabled; previous versions had graceful fallback with the images still visible (just add the `<noscript>` bit back in)
3. Why does `base` have its own `render` function?  Why are there no such functions in `bootstrap` and `bootstrap3`?  What would happen if you did not put the function there?

ralsina:
@Kwpolska 

1. You mean putting jquery in base and using USE_CDN to load one or the other?
2. I don't much mind that usecase. Sure I can add it, but the web in general is inaccessible with JS disabled. In this specific case, sure, why not.
4. Because base has no JQuery loaded, but bootstrap has it.

ralsina:
I will take another shot at this, but this bitrot way too much.

PR #1774, opened 2015-06-01
Making ipynb support truly agnostics in agreement with Jupyter raise...

damianavila:
WIP, but essentially let you use Jupyter (the IPython evolution) generating an ipynb with configurable language agnostic kernel and also populating the notebook metadata in a nikola namespace, so we now have support for `onefile` and `twofiles` (splitted meta). In this way, since the read_metadata is implemented, then we don't have to populate the metadata manually (or with jsextension)... all is more easy now just from the command line ;-)
I have to test some things because it is too late now... I will probably complete it tomorrow, but should be enough completed for first review...


damianavila:
Note to myself... I have to make dicts for each kernel so they have the correct `name`, `language` and `display_name` in the kernelspec...

nikola/plugins/command/new_post.py
```diff
@@ -213,6 +213,14 @@ class CommandNewPost(Command):
                     'bbcode, html, textile, txt2tags',
         },
         {
+            'name': 'kernel',
```
Kwpolska: Do we have to add a new option that will not apply to most compilers?

damianavila: yes... we need to specify the kernel to build the correct notebook template with the correct kernel specification... if not... the notebook is no validated ok and will fail...
I am open to ideas about how to deal with this that, as you said, is not available for other compilers...
One option would be probably make it only available with ipynb (warn and exit if you try to use it with other compilers... but maybe they are other options...

ralsina: It should totally fail when used with other compilers. And it should say in the help that it's only important for jupyter, something like

'help': 'Kernel to use for jupyter posts, irrelevant for all other formats' or somesuch

Kwpolska: My idea would be to make it `-f ipynb@python` — this will also be extensible in the future when we have other compilers with multiple formats (or `-f pandoc@markdown`, for example)

If we do add this parameter, it must also be added in `new_page.py`.

nikola/plugins/compile/ipynb/__init__.py
```diff
@@ -82,38 +82,47 @@ def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False,
...
- "metadata": {
-  "name": ""
- },
- "nbformat": 3,
- "nbformat_minor": 0,
- "worksheets": [
+                empty = {}
+                kernel_meta.append(json.dumps(empty))
+
+            fd.write(NOTEBOOK_TEMPLATE % tuple(kernel_meta))
```
Kwpolska: please use str.format instead of old %-style formatting

nikola/plugins/command/new_post.py
damianavila: >My idea would be to make it -f ipynb@python

This is interesting... I will explore this possibility... we can probably provide the more popular formats: `ipynb@python2`, `ipynb@python3`, `ipynb@julia`, `ipynb@r` and leave the others kernels as a plugable thing (I mean not in core... there are more than 30 kernel... we can not support all of them in core).

nikola/plugins/compile/ipynb/__init__.py
damianavila: oh... yes... still I forget about the new style... I will do the change...

nikola/plugins/command/new_post.py
Kwpolska: How much work is needed to support a kernel, exactly?

damianavila: Just an `ipynb` with a different kernel spec... the thing about supporting more than the 4 main ones is that we need to explore how each of the other ~30 kernels specify their kernelspec and be awared of any change in the future...

damianavila:
@ralsina @Kwpolska this is still WIP, but I am stuck in something and it is late... probably I am missing something obvious but I can not figured out.
First, you need to modify you conf.py and add this line to the COMPILERS:
```
    "ipynb_python2": ('.ipynb',),
```
The issue is that when I run `nikola new_post -f ipynb_python2` I don't get the modifications I want to introduce overriding the set_kernel_metadata method here: https://github.com/getnikola/nikola/pull/1774/files#diff-2f082ac5337057b98fc6c75abb9c4bb4R84

It just take the parent method... not adding anything... ideas welcomed.

After that I just need to add the pyhon3, julia and R kernel metadata and it will be ready for review again...

nikola/plugins/compile/ipynb_python2.plugin
```diff
@@ -0,0 +1,10 @@
+[Core]
+Name = ipynb_python2
+Module = ipynb/ipynb_python2
```
Kwpolska: That’s probably not the correct way to write that.

Kwpolska:
You must probably return `nb_object` and use that instance instead.

I would personally prefer a sleeker implementation that would set `ipynb_kernelspec` and `ipynb_language_info` class variables which would be read by `create_post` and set on `nb`.  (I would also try to get rid of classes by using some sort of “metacompilers” that don’t have real modules, BTW)

damianavila:
>You must probably return nb_object and use that instance instead.

OK, I will try it...

>I would personally prefer a sleeker implementation that would set ipynb_kernelspec and ipynb_language_info class variables which would be read by create_post and set on nb.

Sound interesting...

>(I would also try to get rid of classes by using some sort of “metacompilers” that don’t have real modules, BTW)

Can you elaborate you idea here a little bit more?

Kwpolska:
1. `COMPILERS = {'ipynb@python3': ('.ipynb',)}`
2. split on `@` and pass `python3` as a special parameter to one big ipynb module
3. store data for each language in dicts inside the plugin
4. ???
5. profit

damianavila:
>2. split on @ and pass python3 as a special parameter to one big ipynb module

Passing the parameter as part of the metadata here: https://github.com/getnikola/nikola/blob/master/nikola/plugins/command/new_post.py#L382?
because if I create an special parameter, I guess I should update the other compilers... then I can drop that parameter from the metadata inside create post...

>3. store data for each language in dicts inside the plugin

OK

Kwpolska:
Your idea for 2. makes sense, but we should do it if and only if the compiler has an `@` sign in it.  That way, we won’t need to edit any other compilers.

damianavila:
> You must probably return nb_object and use that instance instead.

Did not work... for forget about that... I just following the last plan.
Pretty close to be done. Will push soon...

damianavila:
@ralsina @Kwpolska AFAIK, this is working OK, please review...
I probably have to add some docs and raise the version number of the plugin... and probably change the changelog... beside those things... I am missing something else? Thanks!!

Kwpolska:
Build failure due to setup.py still including ipynb as a package. Please fix that.

(Also, we should switch to an automated package finder)

damianavila:
Opps, I forgot that one... I will fix it now...

Kwpolska:
Just set it to setuptools.find_packages() and it should work (adjust the
from import?)

-- 

damianavila:
>Just set it to setuptools.find_packages() and it should work (adjust the
from import?)

let's try it...

damianavila:
There is some error in appveyor... but I don't know if it is related with the `find_packages()` thing... on linux tests passed... any hints?

Kwpolska:
It looks like we made it install our test suite in site-packages… I’ll fix
this later today (or you can try yourself); we need to limit the package
finder to the `nikola` package.

-- 

setup.py
```diff
@@ -118,17 +118,7 @@ def run(self):
...
-                'nikola.plugins.command',
-                'nikola.plugins.compile',
-                'nikola.plugins.compile.ipynb',
-                'nikola.plugins.compile.markdown',
-                'nikola.plugins.compile.rest',
-                'nikola.plugins.task',
-                'nikola.plugins.task.sitemap',
-                'nikola.plugins.template',
-                ],
+      packages=find_packages(),
```
Kwpolska:
```py
      packages=find_packages(exclude=('tests',)),
```

Also, in the future, please work on a branch in getnikola/nikola (not your own fork) so we can collaborate.

damianavila: >Also, in the future, please work on a branch in getnikola/nikola (not your own fork) so we can collaborate.

Yes, I will take into account next time... make easier to collaborate, I agree...

damianavila: In fact, closing this one and opening a new PR in the getnikola/nikola becuase probably there are other things to fix...

PR #1780, opened 2015-06-01
Fix #1776 and all related cases

### LINKED ISSUE #1776: `check -l` can’t stomach REDIRECTIONS pages
REDIRECTIONS = [ ( "test.html", "/index.html" ) ]

    WARNING: check: Broken link in output/test.html: 0; url=/index.html
    WARNING: check: Broken link in output/test.html: /index.html

ralsina:
Added basic test for #1776

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -232,6 +232,24 @@ class TranslationsPatternTest1(TranslatedBuildTest):
             outf.write(data)
 
 
+class RedirectionsTest1(EmptyBuildTest):
+    """Check REDIRECTS"""
+
+    @classmethod
+    def patch_site(self):
+        """"""
+        conf_path = os.path.join(self.target_dir, "conf.py")
+        with io.open(conf_path, "a", encoding="utf8") as outf:
+            outf.write("""\n\nREDIRECTS = [ ("posts/foo.html", "/foo/bar.html"), ]\n\n""")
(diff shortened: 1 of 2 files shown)


tests/test_integration.py
```diff
@@ -232,6 +232,50 @@ def patch_site(self):
...
+    """Check external REDIRECTS"""
+
+    @classmethod
+    def patch_site(self):
+        """"""
+        conf_path = os.path.join(self.target_dir, "conf.py")
+        with io.open(conf_path, "a", encoding="utf8") as outf:
+            outf.write("""\n\nREDIRECTS = [ ("foo.html", "http://www.example.com/"), ]\n\n""")
+
+class RedirectionsTest3(EmptyBuildTest):
```
ralsina: It's probably much cheaper (as in "runs a lot faster") to put all of these in a single test class if possible.

ralsina:
Ok, so all those tests seem to pass in travis, and they pass in my computer. They still fail in yours, @Aeyoun ?

Aeyoun:
It’s REDIRECTIONS, not REDIRECTS. In any case, I can’t see that it actually runs check -l on this generated site? Setting `REDIRECTIONS = [ ( "test.html", "should/trigger/warning.html" ), ]` does not fail either.

PR #1783, opened 2015-06-03
ipynb in all themes

ralsina:
Enable mathjax for all ipynb posts/pages.


Kwpolska:
:+1: for this bit, want to do the other half or can I take it?

damianavila:
Seem OK to me...

ralsina:
@Kwpolska feel free, I have a cold and code will be worse than usual for a couple of days :-P

Kwpolska:
The other half is now up for review (c3d5a4e + ab8765e).

Also, ipython blogs often want a change in mathjax config (to accept the ugly `$e = mc^2$` syntax for math) — we still need that as setup for ipynb (unless you can always do `\(e = mc^2\)`)

ralsina:
Looks good to me. +1

damianavila:
I little bit late, but LGTM too...

PR #1788, opened 2015-06-05
fix #1767

### LINKED ISSUE #1767: `nikola version` should check for updates
if `//rdf/RDF/Project/release/Version/revision` (from https://pypi.python.org/pypi?:action=doap&name=Nikola) differs from current version, suggest updating using pip.

ralsina:
Added a ``--check`` option to ``nikola version``


nikola/plugins/command/version.py
```diff
@@ -38,7 +47,25 @@ class CommandVersion(Command):
...
         """Print the version number."""
         print("Nikola v" + __version__)
+        if options.get('check'):
+            if requests is None:
+                req_missing(['requests'], 'check for updates')
+                exit(1)
+            data = requests.get(URL).text
+            doc = lxml.etree.fromstring(data.encode('utf8'))
+            revision = doc.findall('*//{http://usefulinc.com/ns/doap#}revision')[0].text
+            print("Latest version is: ", revision)
```
Kwpolska:
```py
if revision == __version__:
    print("Nikola is up-to-date")
else:
    print("The latest version of Nikola is v{0} -- please upgrade using `pip install --upgrade Nikola=={0}` or your system package manager".format(revision))

ralsina: Good idea.

PR #1794, opened 2015-06-08
fotorama directive for rst

humitos:
Allows to include a fotorama[1] gallery in any post.

[1] http://fotorama.io/

### DIFF
--- a/nikola/data/themes/base/templates/embedded-fotorama.tmpl
+++ b/nikola/data/themes/base/templates/embedded-fotorama.tmpl
@@ -1,0 +1,16 @@
+## -*- coding: utf-8 -*-
+
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script> <!-- 33 KB -->
+
+<link  href="http://cdnjs.cloudflare.com/ajax/libs/fotorama/4.6.4/fotorama.css" rel="stylesheet"> <!-- 3 KB -->
+<script src="http://cdnjs.cloudflare.com/ajax/libs/fotorama/4.6.4/fotorama.js"></script> <!-- 16 KB -->
+
+<div class="fotorama"
+     % for key, value in options.iteritems():
+     data-${key}="${value}"
+     % endfor
+     >
+  % for image in fotorama_content:
+    <a href="${image['url']}"><img src="${image['url_thumb']}"></a>
+  % endfor
+</div>
--- a/nikola/plugins/compile/rest/fotorama.py
+++ b/nikola/plugins/compile/rest/fotorama.py
@@ -1,0 +1,132 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © 2015 Manuel Kaufmann
+
+# Permission is hereby granted, free of charge, to any
+# person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the
+# Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the
+# Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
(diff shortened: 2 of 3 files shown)


humitos:
You can check an example here: http://argentinaenpython.com.ar/fotorama/

ralsina:
This belongs in plugins, not in core. We are moving stuff in that direction.

Unless of course, you want to propose it as a replacement for the current slides directive, in which case it will need a few changes.

nikola/data/themes/base/templates/embedded-fotorama.tmpl
```diff
@@ -0,0 +1,16 @@
+## -*- coding: utf-8 -*-
+
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script> <!-- 33 KB -->
+
+<link  href="http://cdnjs.cloudflare.com/ajax/libs/fotorama/4.6.4/fotorama.css" rel="stylesheet"> <!-- 3 KB -->
```
ralsina: You can't drop CSS link elements in the middle of the HTML, it's not standards compliant (see Notes here: https://developer.mozilla.org/en/docs/Web/HTML/Element/link ).

I'm afraid that means to support this you need to add that link element in your conf.py (EXTRA_HEAD is the option, IIRC)

nikola/data/themes/base/templates/embedded-fotorama.tmpl
```diff
@@ -0,0 +1,16 @@
+## -*- coding: utf-8 -*-
+
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script> <!-- 33 KB -->
```
ralsina: Most themes already load jquery, but they do so at the end of the body.

nikola/data/themes/base/templates/embedded-fotorama.tmpl
```diff
@@ -0,0 +1,16 @@
...
+
+<link  href="http://cdnjs.cloudflare.com/ajax/libs/fotorama/4.6.4/fotorama.css" rel="stylesheet"> <!-- 3 KB -->
+<script src="http://cdnjs.cloudflare.com/ajax/libs/fotorama/4.6.4/fotorama.js"></script> <!-- 16 KB -->
+
+<div class="fotorama"
+     % for key, value in options.iteritems():
```
ralsina: This is not compatible with python 3, just use ``items()``. Yes, it's slightly inefficient in python 2, but you won't have a million options :-)

nikola/plugins/compile/rest/fotorama.py
```diff
@@ -0,0 +1,132 @@
...
+    }
+    has_content = True
+
+    def _sanitize_options(self):
+        defaults = {
+            'nav': 'thumbs',
+            'ratio': '16/9',
+            'keyboard': 'true',
+            'thumb-width': 128,
+            'thumb-height': 128,
```
ralsina: we already have a thumbnails size setting in conf.py it would be great if you could respect it.

nikola/data/themes/base/templates/embedded-fotorama.tmpl
Kwpolska: You could also use the template hooks for that.

-- 

ralsina: indeed an example on how to do that via template hooks would be awesome :-)

Kwpolska: https://getnikola.com/extending.html#template-hooks

Kwpolska:
@humitos Please file your PR against the themes repo, this should not be in the core.

humitos:
Moved here: https://github.com/getnikola/plugins/pull/94

PR #1796, opened 2015-06-08
promote requests to requirements.txt

Kwpolska:
cc @ralsina — needs review on removed ``req_missing``

(might as well cherry-pick all but 71535a9)


ralsina:
Looks good to me!

PR #1801, opened 2015-06-09
fix #973

### LINKED ISSUE #973: Modify the way in which one-file-mode metadata are stripped
Currently, Nikola has the following code for removing metadata comments from HTML output, copy-pasted in most compilers:

```python
if not is_two_file:
    data = re.split('(\n\n|\r\n\r\n)', data, maxsplit=1)[-1]
```

This has one issue: this requires that:

 * the comment block is at the very top of the file
 * the comment block is separated by one empty line from the rest of the comment

Overall, this is kinda hacky.  It’d be good if there was a better way, possibly one that isn’t littered all over our codebase.

Via: #957, @simonvanderveldt.  CC @ralsina, @al3xandru.

ralsina:
Sadly can't do it in all the compiler plugins in the plugins repo because tht would break them for earlier versions.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
+* Cleaner splitting of metadata in onefile posts (Issue #973)
 * It's markdown extra, not extras (Issue #1799)
 
 New in v7.5.0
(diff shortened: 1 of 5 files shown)


Kwpolska:
don’t we do it in all the other compilers?  What about reST?

ralsina:
@Kwpolska we don't do it for pandoc (and maybe we should!), added it for rest.

Kwpolska:
We might try doing it for Pandoc and PHP.

ralsina:
Main problem with pandoc is that we have no idea what's inside the file.

Kwpolska:
Merge when Travis is done testing your branch then.

PR #1816, opened 2015-06-12
Fix #1810 -- use watchdog in auto

### LINKED ISSUE #1810: Use watchdog in auto
Someone else suggested it a while ago, and I (wrongly) didn't even look at it. It would make auto more crossplatform, and simpler, at the cost of a changed dependency.

Kwpolska:
This is #1810.

Fixes (some of which would be useful even without `watchdog`):

* ~~make rebuilds actually work~~ cherry-picked onto master
* rebuild only once by using a set
* tell people we’re rebuilding and show them the log when we’re done

Also, we have to watch the directory with `conf.py` and filter out events because you must watch entire directories with `watchdog`.

cc @ralsina

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,14 @@
+New in master
+=============
+
+Features
+--------
+
+* Use ``watchdog`` in ``nikola auto`` (Issue #1810)
+
+Bugfixes
+--------
+
 New in v7.5.1
 =============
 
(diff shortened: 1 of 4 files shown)


ralsina:
Yo mean rebuilds were not working?

Kwpolska:
yeah, I broke them in f1e95ef211c8e3189bf5cab33351256956a79f91 and nobody really noticed…

ralsina:
let's merge this then :)

Kwpolska:
So is pyyaml no longer a huge problem?

ralsina:
Well, at least cherrypick the fix? We can merge this later, then.

Kwpolska:
done

ralsina:
@Kwpolska let's merge this, dependencies be damned.

PR #1820, opened 2015-06-13
fix #1168

### LINKED ISSUE #1168: Feature: wordpress_import should generate redirections for tags
Hi guys

After I migrated my site to Nikola I noticed that no redirection were generated for tags. It would be helpful if the wordpress_import feature would generate them. 

For categories, it could also be done since categories of posts are transformed into tags, but there are generally less categories. 

Thanks

ralsina:
Add redirects for wordpress tags

### DIFF
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -171,11 +171,9 @@ class CommandImportWordpress(Command, ImportMixin):
         for tag in self.all_tags:
             tag = utils.slugify(tag.decode('utf8'))
             src_url = '{}tag/{}'.format(self.context['SITE_URL'], tag)
-            # FIXME: this should check for PRETTY_URLs in existing config
-            # probably reading that belongs in the Mixin
-            if True:
-                dst_url = self.context['SITE_URL'] + 'categories/' + tag + '.html'
-            self.url_map[src_url] = dst_url
+            dst_url = self.site.link('tag', tag)
+            if src_url != dst_url:
+                self.url_map[src_url] = dst_url
 
         self.write_urlmap_csv(
             os.path.join(self.output_folder, 'url_map.csv'), self.url_map)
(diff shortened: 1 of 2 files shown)


nikola/plugins/command/import_wordpress.py
```diff
@@ -165,6 +166,12 @@ def show_info_about_mising_module(modulename):
             self.extra_languages)
         self.context['REDIRECTIONS'] = self.configure_redirections(
             self.url_map)
+
+        # Add tag redirects
+        for tag in self.all_tags:
+            tag = utils.slugify(tag.decode('utf8'))
+            self.url_map['{}tag/{}'.format(self.context['SITE_URL'], tag)] = self.context['SITE_URL'] + 'categories/' + tag + '.html'
```
Kwpolska: is this guaranteed to be the path? I don’t think so, as it’s either (a) fully customizable or (b) htaccess-dependent…

And it did not take me long (a look in my RSS, in fact) to find [a blog](http://zaufanatrzeciastrona.pl/post/tag/plus-bank/) with `/post/tag/` as the path…

ralsina: But the path is nowhere in the dump that I can see. Worst case, the user can search/replace.

Kwpolska: okay, let’s merge it this way.

Kwpolska: …or let’s not.  you seem to ignore `PRETTY_URLS` here.

ralsina: Seems like the importer never even looks at the existing site's config so far.

I'll see what I can do.

ralsina:
@Kwpolska I made it do "the right thing" which is to use site.config. However, it will not *currently* do the right thing until I fix #1823 

I propose we merge this as is.

Kwpolska:
:+1: for merging, please handle this yourself (in a terminal, unfortunately).

PR #1822, opened 2015-06-13
Add print CSS

Kwpolska:
# Bootstrap & Bootstrap 3 [v2]

* Post pages look well, meta lines have (mostly) hidden URLs and the Source link is hidden altogether; comments are hidden
* ~~Helvetica is used, which may or may not be a good idea~~ v2: Switched to Garamond with fallback to Junicode
* The only thing that is missing is the blog title, which may or may not be necessary, and that would have to be added in a different way (we would have to put a special print-only title outside of navbar).
* Any comments?

image

# Base [v1]

* Very similar to Bootstrap’s, uses the same HTML 5 Boilerplate styles as base
* Blog title can be added easily, but was skipped for consistency

image

### DIFF
--- a/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
+++ b/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
@@ -179,7 +179,7 @@ article.post-micro {
 /* hat tip bootstrap/html5 boilerplate */
 @media print {
     *, *:before, *:after {
-        font-family: Junicode, Garamond, serif;
+        font-family: Garamond, Junicode, serif;
     }
 
     article .entry-title a[href]:after,
--- a/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
+++ b/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
@@ -178,6 +178,10 @@ article.post-micro {
 
 /* hat tip bootstrap/html5 boilerplate */
 @media print {
+    *, *:before, *:after {
+        font-family: Junicode, Garamond, serif;
+    }
+
     article .entry-title a[href]:after,
     article .metadata a[href]:after,
     article .tags a[href]:after {
--- a/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
+++ b/nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
@@ -182,6 +182,10 @@ article.post-micro {
         font-family: Garamond, Junicode, serif;
     }
 
+    body {
+        font-size: 12pt;
+    }
+
     article .entry-title a[href]:after,
     article .metadata a[href]:after,
     article .tags a[href]:after {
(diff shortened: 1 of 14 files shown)


ralsina:
LGTM!

nikola/data/themes/bootstrap3/assets/css/theme.css
```diff
@@ -175,3 +175,24 @@ article.post-micro {
 .entry-summary {
   margin-top: 1em;
 }
+
+/* hat tip bootstrap/html5 boilerplate */
+@media print {
+    article .entry-title a[href]:after,
```
Aeyoun: ::pseudo-elements (two colons)

nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
```diff
@@ -175,3 +175,24 @@ article.post-micro {
 .entry-summary {
   margin-top: 1em;
 }
+
+/* hat tip bootstrap/html5 boilerplate */
+@media print {
```
Aeyoun: Stick this in a separate file?

Aeyoun: Save some ink:

    * {
      background-color: #ffffff;
      color: #1c1c1c;
      font-family: Garamond, serif
    }

nikola/data/themes/bootstrap3/assets/css/theme.css
Kwpolska: Please send your complaints to [Bootstrap](https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.css#L205) and [HTML 5 Boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L219) first.

nikola/data/themes/bootstrap3-jinja/assets/css/theme.css
Kwpolska: Garamond, I’ll accept.

Everything else, please send your complaints to [Bootstrap](https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.css#195) and [HTML 5 Boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L207) first.

> Furthermore, printing in pure black, devoid of any complex color mixing, can make printing much faster, because the printer doesn’t have to mix in any red, green, or blue ink and can focus on getting the task done.

Kwpolska: Done: `Junicode, Garamond, serif` (Junicode is an open-source font which is quite similar to Garamond and probably has similar savings in store).

Using multiple files is not really helpful.  Bundles are still the default, and even if they weren’t, that’s a wasted HTTP request.

Kwpolska: (actually, let’s put Garamond first.)

Kwpolska:
Added some print CSS for base, stolen from HTML 5 Boilerplate and our Bootstrap customizations.

@ralsina, @Aeyoun: please review, both in code and in display (screenshots in the top post)

Kwpolska:
Nobody felt like reviewing for the past 8 days, so I’m merging.  If anything breaks, we can just fix it later.

Issue #1823, opened 2015-06-14
Importers don't read config


ralsina:
In some cases, when importing into an existing site, it will be useful to use the site's current config.

Example: doing redirects in importers requires knowing the PRETTY_URLs setting, or, even better, using the path/link methods that use it.

However, `self.site.config` contains the default config instead of the existing one.

PR #1824, opened 2015-06-15
Fix part of 1823

ralsina:
This makes Nikola initialize LocaleBorg when called without a config file, which is needed to use the default config from importers and should have no bad side effects.

As a result of this, if you import using wordpress like this:

```
nikola import_wordpress foo.xml .
```

The config file **will** be loaded. However, there is still a case where this is broken:

```
nikola import_wordpress foo.xml myoldsite
```

In that case, the config will **NOT** be loaded, which means it will do bad things. So #1823 is not completely fixed by this.


Kwpolska:
We could just just make the `conf.py` file loading part a separate function and use it from within the compilers to solve part 2 (or even do it for both parts).

ralsina:
Yes, sounds like a refactor that gives us value :-)

Kwpolska:
Guess we can merge for now.

PR #1834, opened 2015-06-18
Agnostic IPynb, supersede 1774

damianavila:
Now in the nikola repo to allow easy collaboration...


damianavila:
To know previous discussions, see #1774 
ping @Kwpolska who was reviewing the other PR...

nikola/plugins/command/new_post.py
```diff
@@ -268,6 +268,10 @@ def _execute(self, options, args):
             onefile = self.site.config.get('ONE_FILE_POSTS', True)
 
         content_format = options['content_format']
+        content_subformat = None
+
+        if "@" in content_format:
+            content_format, content_subformat = tuple(content_format.split("@"))
```
Kwpolska: `tuple()` is unnecessary and weird

nikola/plugins/compile/ipynb.py
```diff
@@ -85,35 +84,123 @@ def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False,
...
         if onefile:
-            raise Exception('The one-file format is not supported by this compiler.')
+            nb["metadata"]["nikola"] = metadata
+
+        if kernel is not None:
+            nb["metadata"]["kernelspec"] = ipython_kernel_spec[kernel]
+            nb["metadata"]["language_info"] = ipython_language_info[kernel]
+
         with io.open(path, "w+", encoding="utf8") as fd:
-            if not content.startswith("Write your"):
```
Kwpolska: This was an intentional feature: users could create an ipynb outside of Nikola and just import it by using the `-f` option.  And we want to keep it in place.

You will have to refactor this function.  Due to the `onefile` changes, you might have to load the notebook, add meta (and *nothing else*) and save it back.

nikola/plugins/compile/ipynb.py
```diff
@@ -85,35 +84,123 @@ def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False,
...
-     "outputs": []
-    }
-   ],
-   "metadata": {}
-  }
- ]
-}""")
+                nbformat.write(nb, fd, 'ipynb')
+
+# python2 nb metadata info
```
Kwpolska: I would recommend just writing it as one large dict of dicts and not messing with all these variables:

```python
IPYNB_DATA = {
    'python2': {
        'kernelspec': {…},
        'codemirror_mode': {…},
    },
    …
}
```

Kwpolska: Also: can’t we ask ipython for this information?  It really sucks if *we* have to keep it in our codebase and *we* have to keep it up-to-date…

nikola/plugins/command/new_post.py
damianavila: you are right... will change

nikola/plugins/compile/ipynb.py
damianavila: > This was an intentional feature: users could create an ipynb outside of Nikola and just import it by using the -f option. And we want to keep it in place.

In fact, rethinking this, I would like the users to use `nikola new_post --import path/to/ipynb`, but I have to investigate a little bit how `--import` works

> You will have to refactor this function. Due to the onefile changes, you might have to load the notebook, add meta (and nothing else) and save it back.

Will look into it...

Kwpolska: `nikola new_post --import foo.ipynb`:

* loads the `foo.ipynb` file
* sets `content` to the contents of the file
* does nothing else to tell the compiler about it for compatibility reasons (hence the `"Write your"` hack)
* does not touch the file in any way

nikola/plugins/compile/ipynb.py
damianavila: >I would recommend just writing it as one large dict of dicts 

OK

>can’t we ask ipython for this information? 

No, because we are want to start with the ipynb with the correct kernel... and IPython does not know which kernel start until it gets the info from the ipynb (or goes to the default kernel if there is no info available....

>It really sucks if we have to keep it in our codebase and we have to keep it up-to-date…

Yep, sucks... but I am only adding the main kernels, which I think will be pretty stable so far...
I probably can make so more research to cut all this info to the minimal set of specification to start the correct kernel... then IPython fills the holes... I will research on that...

Kwpolska: Please report an issue upstream (getting a list of available kernels/as-small-as-possible kernel data).  You could also try looking at how ipynbs are created. Maybe they can help…

damianavila: getting the list of available is possible... mmm... new ideas... more work but I think it will work nice... thanks for the brainstorming. I'll be back!

nikola/plugins/compile/ipynb.py
damianavila: OK, thanks for the description... will fix this too.

Kwpolska:
I just cleaned up a lot of things and made `python(2|3)` the default kernel.

Still needs to be done:

- [ ] support importing posts (should not be hard)
- [ ] replace IPYNB_KERNELS with something that gets data out of the local system.

cc @damianavila

Kwpolska:
@damianavila I introduced a lot of changes here today.  Please test and review.

damianavila:
@Kwpolska thanks for working on this... I was just starting to do it by myself and I thought, I have to check if Chris did not do it before proceed... I will review in a little bit!

damianavila:
@Kwpolska, I just push some minor fixes... still there seems to be an error in the TravisCI run:
```
=================================== FAILURES ===================================
______________________ InvariantBuildTest.test_invariance ______________________
self = <tests.test_integration.InvariantBuildTest testMethod=test_invariance>
    def test_invariance(self):
        """Compare the output to the canonical output."""
        if sys.version_info[0:2] != (2, 7):
            pytest.skip('only python 2.7 is supported right now')
        good_path = os.path.join(os.path.dirname(__file__), 'data', 'baseline{0[0]}.{0[1]}'.format(sys.version_info))
        if not os.path.exists(good_path):
            pytest.skip('no baseline found')
        with cd(self.target_dir):
            try:
                diff = subprocess.check_output(['diff', '-ubwr', good_path, 'output'])
                self.assertEqual(diff.strip(), '')
            except subprocess.CalledProcessError as exc:
                print('Unexplained diff for the invariance test. (-canonical +built)')
                print(exc.output.decode('utf-8'))
>               self.assertEqual(exc.returncode, 0, 'Unexplained diff for the invariance test.')
E               AssertionError: Unexplained diff for the invariance test.
/home/travis/build/getnikola/nikola/tests/test_integration.py:545: AssertionError
----------------------------- Captured stdout call -----------------------------
Unexplained diff for the invariance test. (-canonical +built)
diff -ubwr /home/travis/build/getnikola/nikola/tests/data/baseline2.7/assets/css/nikola_ipython.css output/assets/css/nikola_ipython.css
--- /home/travis/build/getnikola/nikola/tests/data/baseline2.7/assets/css/nikola_ipython.css	2015-06-20 08:34:47.000000000 +0000
+++ output/assets/css/nikola_ipython.css	2015-06-20 13:41:26.194658000 +0000
@@ -52,10 +52,6 @@
   margin: 0 0 0px;
 }
 
-h1, h2, h3, .metadata p {
-margin-left: 15px;
-}
-
 .highlight .hll { background-color: #ffffcc }
 .highlight  { background: #f8f8f8; }
 .highlight .c { color: #408080; font-style: italic } /* Comment */
```
Ideas about the diff?

Otherwise, I tested the PR and everything works as expected, AFAIK...

Kwpolska:
This is expected, I removed that in https://github.com/getnikola/nikola/commit/f5717266cda2241cf55af452cfa6773ee7931b67 and thus made the site output different.  Just ignore it, it will disappear as soon as I merge and rebuild the invariant master.

Merging!

damianavila:
Thanks for your help!! Nice to this merged!

PR #1840, opened 2015-06-21
Fix #1837 -- nikola new_post --available-compilers

### LINKED ISSUE #1837: Add an option to `nikola new_post` to list available compilers
Right now, we have this to tell users what to do:

```
  -f ARG, --format=ARG      Markup format for the post, one of rest, markdown, wiki, bbcode, html, textile, txt2tags.
```

This is misleading, contains ancient compilers that now reside in the plugins repo, misses new important compilers (ipynb) and does not say anything about COMPILERS/POSTS.  Instead, it should be replaced with a `--available-formats` option to print a list.  For example:

```
Available formats:


 rest      reStructuredText          .rst, .txt
!markdown  Markdown                  .md, .mdown, .markdown
!ipynb     IPython Notebook/Jupyter  .ipynb
!html      HTML                      .html
!php       PHP                       .php
!pandoc    Pandoc                    (disabled: no extensions declared)

! Compilers marked with an exclamation mark require additional configuration.
  Read more: https://getnikola.com/handbook.html#configuring-other-input-formats
```

Kwpolska:
This is #1837.

Also accessible as -F, and in nikola new_page.

Sample output (from my own blog):

```
$ nikola new_post --available-compilers
Available input formats:

 NAME      EXTENSIONS

 html      .html, .htm
 rest      .txt, .rst
!ipynb     .ipynb
!markdown  .md, .mdown, .markdown
~pandoc    (disabled: not in COMPILERS)
~php       (disabled: not in COMPILERS)

More compilers are available in the Plugins Index.

Compilers marked with ! and ~ require additional configuration:
    ! not in the PAGES/POSTS tuples (unused)
    ~ not in the COMPILERS dict (disabled)
Read more: https://getnikola.com/handbook.html#configuring-other-input-formats
```

(there is no “friendly name” field, cannot do that)

### DIFF
--- a/nikola/plugins/command/new_post.py
+++ b/nikola/plugins/command/new_post.py
@@ -118,8 +118,8 @@ def print_compilers(compilers_raw, post_pages, compiler_objs):
 
     print("Available input formats:\n")
 
-    name_width = max([len(i[0]) for i in parsed_list] + [4]) # 4 == len('NAME')
-    fname_width = max([len(i[1]) for i in parsed_list] + [11]) # 11 == len('DESCRIPTION')
+    name_width = max([len(i[0]) for i in parsed_list] + [4])  # 4 == len('NAME')
+    fname_width = max([len(i[1]) for i in parsed_list] + [11])  # 11 == len('DESCRIPTION')
 
     print((' {0:<' + str(name_width) + '}  {1:<' + str(fname_width) + '}  EXTENSIONS\n').format('NAME', 'DESCRIPTION'))
 
--- a/nikola/plugins/command/new_post.py
+++ b/nikola/plugins/command/new_post.py
@@ -30,6 +30,7 @@ import datetime
 import os
 import sys
 import subprocess
+import operator
 
 from blinker import signal
 import dateutil.tz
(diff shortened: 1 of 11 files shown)


ralsina:
We can add a friendly name field for plugins, then :-)

ralsina:
Also, we should classify plugins per type, I think. It can be done inspecting the code.

Kwpolska:
What sort of “type” do you mean?

nikola/plugins/command/new_post.py
```diff
@@ -37,16 +38,22 @@
...
 from nikola import utils
 
+COMPILERS_DOC_LINK = 'https://getnikola.com/handbook.html#configuring-other-input-formats'
 POSTLOGGER = utils.get_logger('new_post', utils.STDERR_HANDLER)
 PAGELOGGER = utils.get_logger('new_page', utils.STDERR_HANDLER)
 LOGGER = POSTLOGGER
 
 
-def filter_post_pages(compiler, is_post, compilers, post_pages, compiler_names):
+def filter_post_pages(compiler, is_post, compilers, post_pages, compiler_names, compilers_raw=None):
```
Kwpolska: @ralsina: are we okay with an API change here (drop `=None`)?

Kwpolska:
Friendly names added.  Note that the API for `filter_post_pages` changed substantially.

ralsina:
Compiler/task/rest extension/etc.

Kwpolska:
I'm not sure I understand… This plugin adds a new feature that handles
installed compilers only. What do you want to be done?

-- 

ralsina:
@Kwpolska this sends the user to the plugins site looking for compiler plugins. We have enough plugins that actually finding the compilers takes a little while.

Kwpolska:
We could just add a new field to `.plugin` files and let users filter by that.  Should not be hard to implement (but I’m not taking it, at least for now)

PR #1847, opened 2015-06-27
Fixing little things in WordPress importer (WIP)

felixfontein:
When importing a WordPress blog whose blog description was not set, I got `null` inserted into `conf.py` instead of `"PUT DESCRIPTION HERE"`, which caused `nikola build` to fail. This is fixed in 5781644609e3aac850c41d4e71a08220934bb52d.

Then, there are two more problems with the WordPress importer which make it fail for me; see comments here: https://github.com/getnikola/nikola/commit/b6cbfcbc

### DIFF
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -550,7 +550,7 @@ def get_text_tag(tag, name, default):
     if tag is None:
         return default
     t = tag.find(name)
-    if t is not None:
+    if t is not None and t.text is not None:
         return t.text
     else:
         return default
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -184,7 +184,11 @@ class CommandImportWordpress(Command, ImportMixin):
 
         # Add tag redirects
         for tag in self.all_tags:
-            tag = tag.decode('utf8')
+            # In python 2, path is a str. slug requires a unicode
+            # object. According to wikipedia, unquoted strings will
+            # usually be UTF8
+            tag_str = tag.decode('utf8') if isinstance(tag, utils.bytes_str) else tag
+            tag = utils.slugify(tag_str)
             src_url = '{}tag/{}'.format(self.context['SITE_URL'], tag)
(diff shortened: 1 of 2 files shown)


felixfontein:
Parts will be fixed in #1824.

Missing at the moment: exception raised by `tag.decode('utf8')` under Python 3.

felixfontein:
The currently failed checks are due to a disk full error on Travis, and a missing import `markdown`.

nikola/plugins/command/import_wordpress.py
```diff
@@ -169,7 +169,11 @@ def show_info_about_mising_module(modulename):
 
         # Add tag redirects
         for tag in self.all_tags:
-            tag = utils.slugify(tag.decode('utf8'))
+            # In python 2, path is a str. slug requires a unicode
+            # object. According to wikipedia, unquoted strings will
+            # usually be UTF8
```
Kwpolska: Link to that wikipedia page in code comments.

nikola/plugins/command/import_wordpress.py
```diff
@@ -169,7 +169,11 @@ def show_info_about_mising_module(modulename):
...
+            tag_str = tag.decode('utf8') if isinstance(tag, utils.bytes_str) else tag
```
Kwpolska:
```py
try:
    tag_str = tag.decode('utf8')
except AttributeError:
    tag_str = tag
```

Kwpolska:
Please make this mergeable and use try/except for decoding, and you can merge as soon as Travis builds.

nikola/plugins/command/import_wordpress.py
felixfontein: This piece is essentially a copy'n'paste from somewhere else in the same source file. The original comment has no link to Wikipedia, either.

felixfontein:
Should I modify both occurences in this case? (The second is here: https://github.com/getnikola/nikola/blob/master/nikola/plugins/command/import_wordpress.py#L419)

felixfontein:
Ok, modified both. Now merging.

PR #1848, opened 2015-06-27
Allowing HTTP authentication for downloading WordPress blog content

felixfontein:
Also providing error messages when not downloading instead of storing error message in destination file.

### DIFF
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -84,6 +84,13 @@ class CommandImportWordpress(Command, ImportMixin):
             'help': "Do not try to download files for the import",
         },
         {
+            'name': 'download_auth',
+            'long': 'download-auth',
+            'default': None,
+            'type': str,
+            'help': "Specify username and password for HTTP authentication (separated by ':')",
+        },
+        {
             'name': 'separate_qtranslate_content',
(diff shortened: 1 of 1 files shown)


nikola/plugins/command/import_wordpress.py
```diff
@@ -131,6 +138,14 @@ def _execute(self, options={}, args=[]):
         self.exclude_drafts = options.get('exclude_drafts', False)
         self.no_downloads = options.get('no_downloads', False)
 
+        self.auth = None
+        if options.get('download_auth', None) is not None:
```
Kwpolska: You don’t need to explicitly specify `, None`, it’s the default.

nikola/plugins/command/import_wordpress.py
```diff
@@ -131,6 +138,14 @@ def _execute(self, options={}, args=[]):
...
+            username_password = options.get('download_auth')
+            self.auth = tuple(username_password.split(':', 1))
```
Kwpolska: tuple() is unnecessary here AFAIK.

felixfontein: Both the requests documentation and source code explicitly mention tuple, but not list. So I'd prefer to stick to a tuple here.

PR #1853, opened 2015-07-01
Fix 1830

ralsina:
Make nikola auto work correctly with python3.


Kwpolska:
It looks sensible. If you tested everything you changed, with Python 2.7 and 3.4, please merge.

Issue #1850, opened 2015-06-30
Spurious "scheduled to be published in the future" warning


lordmauve:
Using Nikola 7.5.1 on Python 3, I get messages of the form

    [2015-06-30T18:03:48Z] NOTICE: Nikola: path/to/post.rst is scheduled to be published in the future (2015-06-30 18:00:00+00:00)

As can be seen from the timestamps, this message isn't even self-consistent.

I tried different ways of expressing the time zone (UTC, local time etc) and nothing works except setting the publication time over an hour into the past in UTC. My local time is currently UTC+1; perhaps Nikola is applying the timezone offset twice?

Kwpolska:
This sounds really weird.  Please open `nikola console`, run the following commands and paste the transcript of your session:

```py
post = site.timeline[0]
post.title # if it’s wrong, please fiddle with the index above until you get it right
post.date
post.current_time
post.date >= post.current_time
post.meta['en']['date']
post.config['TIMEZONE']
post.config['__tzinfo__']
import nikola.utils
nikola.utils.to_datetime(post.meta['en']['date'])
nikola.utils.to_datetime(post.meta['en']['date'], post.config['__tzinfo__'])
```

Kwpolska:
On a side note, I can’t reproduce with `.. date: 2015-06-30 18:00:00 UTC` on Python 3.4.3 with 'UTC' as the Nikola timezone and with UTC+02:00 local time.

habilain:
I'm also experiencing this issue. One workaround is to set TIMEZONE = None in the conf.py, which seems to correct things.

As to what's going on, I used the tried and tested debug prints. On my system, which is set to BST (UTC+1), I created a new post at 22:44:45. The post was faithfully tagged with the time 2015-07-02 22:44:45+01:00, but during building the website, Nikola had the time 2015-07-02 21:48:09.397857+01:00. 

In any case, long story short, the problem is that Nikola wasn't passing the timezone from the config when getting the current time, and dateutils wasn't getting the current timezone - maybe a linux thing.

I've created a pull request with the fix, but it's so small that the pull request may be overkill. Not really used to Git's workflow.

Kwpolska:
Pull Requests are never overkill, they are the correct way of doing this.  You don’t even need to mess with Git in this case, you could create a PR by opening that file and clicking the Edit button.

This fix makes perfect sense, but you might have some timezone misconfiguration on your system if `tzlocal()` does not work properly.

[...]

PR #1858, opened 2015-07-02
Fix for #1850

habilain:
Fix for problem where posts may be tagged as being in the future when due to time zone configuration not being used and the fallback not working

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1858)


Kwpolska:
Thanks!

PR #1861, opened 2015-07-04
Fix #1856 -- make tag listing pages per-language

### LINKED ISSUE #1856: Tag listing pages should use only tags that exist in that language
If you have different tags in the languages, tag lists show tags from all languages instead of posts from only one language. To fix this, we would need to replace `posts_per_tag` with `tags_per_lang` or something like that.

Problematic code: [tags.py](https://github.com/getnikola/nikola/blob/master/nikola/plugins/task/tags.py#L182,L207) → [nikola.py](https://github.com/getnikola/nikola/blob/master/nikola/nikola.py#L1460)

via [nikola-discuss](https://groups.google.com/forum/#!topic/nikola-discuss/QueyOeFVwdU)/cc @tekNico

Kwpolska:
This change will display only tags that exist in this language’s posts, honoring `SHOW_UNTRANSLATED_POSTS`.  However, the current behavior of per-tag pages is retained to help people who jump between languages by changing URLs.

cc @tekNico

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1861)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,6 @@
 [run]
 source = nikola
-omit = /tmp/*, _*, nikola/packages, nikola/data, nikola/winutils
+omit = /tmp/*, _*, nikola/packages*, nikola/data*, nikola/winutils*
 [report]
 exclude_lines =
     pragma: no cover
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,6 @@
 [run]
 source = nikola
-omit = /tmp/*, _*
+omit = /tmp/*, _*, nikola/packages, nikola/data, nikola/winutils
 [report]
 exclude_lines =
     pragma: no cover
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,6 +7,7 @@ Features
 Bugfixes
 --------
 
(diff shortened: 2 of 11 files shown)


teknico:
Tested and it works exactly as I hoped, thanks a lot!

Brown paper bag moment: can you please resync the Italian translation from Transifex? I botched a pretty important term ("English" -> "Italiano"), fixed now.

Regarding "the current behavior of per-tag pages", pages of tags in all languages are accessible by URLs in all languages (not only their own), is that what you meant? That's fine by me.

Kwpolska:
> Brown paper bag moment: can you please resync the Italian translation from Transifex? I botched a pretty important term ("English" -> "Italiano"), fixed now.

That would normally be done on release, but I’ll handle it now.

> Regarding "the current behavior of per-tag pages", pages of tags in all languages are accessible by URLs in all languages (not only their own), is that what you meant? That's fine by me.

Yup.

teknico:
> That would normally be done on release, but I’ll handle it now.

Unfortunately the v.7.6.0 release happened in the meantime, and its Italian translation is similarly botched. I apologize. :-(

PR #1864, opened 2015-07-05
Unifying compiler extensions

felixfontein:
In c65be6e24a5daaa3d1b7526ae9dcce7daf017f64 Kwpolska added a new plugin class, `CompilerExtension`, and made `MarkdownExtension` and `RestExtension` subclasses of this new plugin class. This PR moves the plugin loading code from the Rest and Markdown plugins into a common place in `PageCompiler`, which can be used by all page compiler plugins.

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1864)


Kwpolska:
Broken: https://ci.appveyor.com/project/Kwpolska/nikola/build/1.0.1865#L1681
Yapsy probably does not think of them as CompilerExtensions, it looks like we have to activate all three names in `nikola/nikola.py`.

felixfontein:
Yapsy does find them. There's another problem, and I'm currently working on it.

felixfontein:
We have to activate all three names, but not because yapsy won't identify derived classes of `RestExtension` and `MarkdownExtension` as `CompilerExtension` plugins, but because yapsy would identify the imported `RestExtension` or `MarkdownExtension` class definition as a plugin and would often ignore the real plugin defined in the file.

felixfontein:
Part of the problem is I think that many plugins in Nikola ignore the advice here: http://yapsy.sourceforge.net/Advices.html#plugin-class-detection-caveat

Kwpolska:
We can’t fix it now, there would be so many things to change.

felixfontein:
I assume that the failed invariance test is part of another PR. The rest seems to work fine now.

tests/base.py
```diff
@@ -239,5 +242,17 @@ def __init__(self):
         self.template_system = self
         self.name = 'mako'
 
+    def _activate_plugins_of_category(self, category):
```
Kwpolska: This is code duplication, is it possible to import `nikola.nikola` and do `FakeSite._activate_plugins_of_category = nikola.nikola.Nikola._activate_plugins_of_category` instead?

felixfontein: Seems to work locally for me, now let's see what Travis and AppVeyor say :)

felixfontein: Doesn't seem to work. What about moving the method to `nikola.utils` and calling it explicitly with a `site` instance?

Kwpolska: Nah, let’s just keep it duplicated.  Merging.

(Invariance failures are caused by this being branched out before rst.css and aria fixes.)

felixfontein:
Thanks!

PR #1867, opened 2015-07-06
Extending and refactoring WordPress importer

felixfontein:
Beginning with some refactoring and extending of the WordPress importer.
Done so far:
 * Feeding all posts together with post format through transform_content, and let it decide extension.
 * No longer using extension `.wp` for WordPress-to-markdown converted posts.
 * Exporting wp-status and excerpt (if available).
 * Allowing to exclude private items and include empty items.
 * Exporting attachment-post/page relation as JSON file.
 * Exporting categories (if there's only one per post) including category hierarchy.
 * Exporting comments.
 * Add support for WordPress page compiler plugin (both to make new site use it, or to use it to convert WordPress posts to `.html` files).

[Review on Reviewable](https://reviewable.io/reviews/getnikola/nikola/1867)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,21 @@ New in master
 Features
 --------
 
+* Several improvements to WordPress importer (PR #1867):
+  * Allowing to export categories and category hierarchy with
+    --export-categories-as-categories
+  * Allowing to exclude private posts, and allowing to include empty posts
+  * Allowing to use HTTP authentication for downloads with --download-auth
+    (PR #1848)
+  * Allowing to export comments with --export-comments
+  * Allowing to use WordPress page compiler to directly convert posts
+    to HTML on import with --transform-to-html
+  * Allowing to use WordPress page compiler on imported site instead of
+    converting posts to markdown with --use-wordpress-compiler
+  * Allowing to automatically install the WordPress page compiler when
+    needed with --install-wordpress-compiler
+  * Exporting information on attachments per post as JSON
+  * Exporting post status and excerpt
 * New ‘pagekind’ variable available to identify different kind of pages from theme templates
 * Add ``--no-server`` option to ``nikola auto`` (Issue #1883)
 * Always return unicode in slugify (Issue #1885)
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -1908,15 +1908,47 @@ the following:
 
   This is also useful for DISQUS thread migration!
 
-* Will try to convert the content of your posts. This is *not* error free, because
-  WordPress uses some unholy mix of HTML and strange things. Currently we are treating it
-  as markdown, which does a reasonable job of it.
+* Allows you to export your comments with each post
+* Exports information on attachments per post
+* There are different methods to transfer the content of your posts:
 
-  You will find your old posts in ``new_site/posts/post-title.wp`` in case you need to fix
-  any of them.
+  - You can convert them to HTML with the WordPress page compiler plugin
+    for Nikola. This will format the posts including supported shortcodes
+    the same way as WordPress does. Use the ``--transform-to-html`` option
+    to convert your posts to HTML.
+
+    If you use this option, you do not need to install the plugin
+    permanently. You can ask Nikola to install the plugin into the subdirectory
+    ``plugins`` of the current working directory by specifying
+    the ``--install-wordpress-compiler`` option.
+
+  - You can leave the posts the way they are and use the WordPress page
+    compiler plugin to render them when building your new blog. This also
+    allows you to create new posts using the WordPress syntax, or to manually
+    add more shortcode plugins later. Use the ``--use-wordpress-compiler``
+    option to not touch your posts.
+
+    If you want to use this option, you have to install the plugin permanently.
+    You can ask Nikola to install the plugin into your new site by specifying
+    the ``--install-wordpress-compiler`` option.
+
+  - You can let Nikola convert your posts to Markdown. This is *not* error
+    free, because WordPress uses some unholy mix of HTML and strange things.
+    This is the default option and requires no plugins.
+
+  You will find your old posts in ``new_site/posts/post-title.html`` in the first case,
+  ``new_site/posts/post-title.wp`` in the second case or ``new_site/posts/post-title.md``
+  in the last case if you need to edit or fix any of them.
+
+  Please note that the page compiler currently only supports the ``[code]`` shortcode,
+  but other shortcodes can be supported via plugins.
+
+  Also note that the WordPress page compiler is licensed under GPL v2 since
+  it uses code from WordPress itself, while Nikola is licensed under the more
+  liberal MIT license.
 
 This feature is a work in progress, and the only way to improve it is to have it used for
-as many sites as possible and make it work better each time, so I am happy to get requests
+as many sites as possible and make it work better each time, so we are happy to get requests
 about it.
 
 .. [#] The dump needs to be in 1.2 format. You can check by reading it, it should say
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -550,10 +550,16 @@ class CommandImportWordpress(Command, ImportMixin):
         else:
             return content
 
-    def transform_content(self, content, post_format):
+    def transform_content(self, content, post_format, attachments):
         if post_format == 'wp':
             if self.transform_to_html:
-                content = self.wordpress_page_compiler.compile_to_string(content)
+                additional_data = {}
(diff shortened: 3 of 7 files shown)


Kwpolska:
> Add support for WordPress page compiler plugin (both to make new site use it, or to use it to convert WordPress posts to .html files).

Should be simple to implement by just talking to the `plugin` plugin and loading the new plugin (can you call `collectPlugins` twice without everything going down in flames?).  Make sure to add a command-line option to pick the usage mode, and if there is no value specified, make sure to ask — please include a note about the compiler’s license for anti-GPL zealots.

nikola/plugins/command/import_wordpress.py
```diff
@@ -536,17 +601,45 @@ def process_item(self, item):
...
+            if post_type == 'post':
+                out_folder_slug = self.import_item(item, wordpress_namespace, 'posts')
+            else:
+                post_type = 'page'
+                out_folder_slug = self.import_item(item, wordpress_namespace, 'stories')
+            # If post was exported, store data
+            if out_folder_slug:
+                self.posts_pages[post_id] = (post_type, out_folder_slug[0], out_folder_slug[1])
+
+    def write_attachments_info(self, path, attachments):
```
Kwpolska: Who would use those files? Would it actually be useful to give Nikola some real post attachments support, btw?

Kwpolska:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/getnikola/nikola/1867)

Kwpolska:
<sup>**[nikola/plugins/command/import_wordpress.py, line 443 \[r1\]](https://reviewable.io:443/reviews/getnikola/nikola/1867#-JtbhjIHB75Ne-WiSATG)** ([raw file](https://github.com/getnikola/nikola/blob/a334aa395d89e9a828c3a8ea5bcb2a8c34ef81cb/nikola/plugins/command/import_wordpress.py#L443)):</sup>
Is there any other post format available? Would that be HTML, or what?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/getnikola/nikola/1867)

nikola/plugins/command/import_wordpress.py
felixfontein: These files could be used by plugins for the WordPress page compiler. For example, if you include attachments into a post as a gallery, you have something like
```[gallery columns="3" link="file" ids="52,53,54,55,56,57,58"]```
inside the post -- here 52 up to 58 are IDs for attachments. If you have this `.json` file, you can map the IDs to actual images.

Comment (felixfontein) [inline nikola/plugins/command/import_wordpress.py]: About giving Nikola proper post attachments functionality: I don't really care enough about that feature to implement/want it.

**Comment by felixfontein** [comment]
Regarding the other post formats: I don't know. That was already implemented before I touched it; apparently there's a markdown plugin for WordPress. (I looked it up: #484)

**Comment by felixfontein** [comment]
Ok, I now looked at the WordPress plugin which allows to use different text formats, and added more explicit support, including an error when an unsupported format is encountered.

**Comment by felixfontein** [comment]
Regarding calling `collectPlugins()` more than once: the documentation doesn't explicitly disallow this, and the source seems to have no problem with it. Yapsy has a map for every category in which is indexed by the plugin info file's filename (`PluginManager._category_file_mapping`), so if you call the function again, it will go through all plugins again, load them, go through all categories to which they fit, and ignore them (assuming neither the plugin nor the categories changed since the last call of `collectPlugins()`.

So installing a plugin and then calling `collectPlugins()` a second time will not hurt, but find the new plugin.

File: nikola/plugins/command/import_wordpress.py [hunk 2]
```diff
@@ -148,6 +179,53 @@ def _execute(self, options={}, args=[]):
...
+                cat_name = get_text_tag(cat, '{{{0}}}cat_name'.format(wordpress_namespace), None)
+                cat_path = [cat_name]
+                if cat_parent_slug in cat_map:
+                    cat_path = cat_map[cat_parent_slug] + cat_path
+                cat_map[cat_slug] = cat_path
+            self._category_paths = dict()
+            for cat, path in cat_map.items():
+                self._category_paths[cat] = utils.join_hierarchical_category_path(path)
+
+    def _adjust_config_template(self, channel, rendered_template):
```
Kwpolska: huge :-1:

Instead, you should change the input to the template and the template itself, just like I did in https://github.com/getnikola/nikola/commit/0f4b1e2596077177f2e81b16036424a13318915c (relevant lines: 57/216).  Don’t mess with regex.  (It was done before, this code is probably ancient)

felixfontein: Yes, I took the old code and modified it. I've removed it now and did it the way you wanted. (Actually I prefer it that way, too :) )

felixfontein:
Ok, I think I'm done for now.

CHANGES.txt
```diff
@@ -7,6 +7,20 @@ Features
...
+    --export-categories-as-categories
+  * Allowing to exclude private posts, and allowing to include empty posts
+  * Allowing to use HTTP authentication for downloads with --download-auth
+  * Allowing to export comments with --export-comments
+  * Allowing to use WordPress page compiler to directly convert posts
+    to HTML on import with --transform-to-html
+  * Allowing to use WordPress page compiler on imported site instead of
+    converting posts to markdown with --use-wordpress-compiler
+  * Allowing to automatically install the WordPress page compiler when
+    needed with --install_wordpress_compiler
```
Kwpolska: use `-` instead of `_`
also, you should advertise the compiler if the user does not ask for it

Kwpolska:
1. Merge master (or rebase and force-push this branch).
2. Put your changes at the top of the Features section.
3. Test with some wordpress dumps if it works properly, if you haven’t already.

docs/manual.txt
```diff
@@ -1908,13 +1908,23 @@ the following:
 
   This is also useful for DISQUS thread migration!
 
+* Allows you to export your comments with each post
+* Exports information on attachments per post
 * Will try to convert the content of your posts. This is *not* error free, because
```
Kwpolska: put info about your compiler here, drop this paragraph (you should really push the wordpress compiler)

nikola/plugins/command/import_wordpress.py
```diff
@@ -49,10 +52,27 @@
...
 from nikola.plugins.command.init import SAMPLE_CONF, prepare_config, format_default_translations_config
+from nikola.plugins.command import plugin
 
 LOGGER = utils.get_logger('import_wordpress', utils.STDERR_HANDLER)
 
 
+def install_plugin(site, plugin_name, output_dir=None, show_install_notes=False):
+    LOGGER.notice("Installing plugin '{0}'".format(plugin_name))
+    plugin_installer = plugin.CommandPlugin()
+    plugin_installer.set_site(site)
```
Kwpolska: We use our plugins as singletons.  Take the instance from `plugin_installer = site.plugin_manager.getPluginByName('plugin', 'Command').plugin_object`

nikola/plugins/command/import_wordpress.py
```diff
@@ -107,15 +141,54 @@ class CommandImportWordpress(Command, ImportMixin):
...
     ]
     all_tags = set([])
 
-    def _execute(self, options={}, args=[]):
-        """Import a WordPress blog from an export file into a Nikola site."""
-        if not args:
-            print(self.help())
-            return
+    def _find_wordpress_compiler(self):
+        for plugin_info in self.site.plugin_manager.getPluginsOfCategory('PageCompiler'):
```
Kwpolska: `self.site.plugin_manager.getPluginByName('wordpress', 'PageCompiler')` — returns plugin info or `None`

CHANGES.txt
felixfontein: Sorry, that was a copy'n'paste typo. Fixed.

docs/manual.txt
felixfontein: I rewrote the documentation a bit. Can you take another look?

nikola/plugins/command/import_wordpress.py
felixfontein: Thanks! Fixed.

nikola/plugins/command/import_wordpress.py
felixfontein: Thanks as well. Fixed.

felixfontein:
Ok, all tests successfull (with two of my WordPress dumps and different settings, and automatic ones). Squashing a bit, and merging.

PR #1871, opened 2015-07-08
Add -d, --detach to nikola serve

Kwpolska:
This is a clone of `jekyll serve --detach`.  The feature will return control to your terminal immediately, letting you work on your website with the server running in the background.  This is accomplished by
forking right before the server is initialized. All log messages are discarded.

    $ nikola serve -d
    [2015-07-08T08:20:25Z] INFO: serve: Serving HTTP on 0.0.0.0 port 8000...
    [2015-07-08T08:20:25Z] INFO: serve: Detached with PID 14594. Run `kill 14594` to stop the server.

This is functionally equivalent to `nikola serve 2> /dev/null &`, but is shell-independent and easier to type.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Add ``-d``, ``--detach`` option to ``nikola serve`` (Issue #1871)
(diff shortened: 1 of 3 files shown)


Kwpolska:
Merged due to no response in 2 days.

Issue #1874, opened 2015-07-08
demo site has some problems


obestwalter:
Today I started trying nikola (v7.6.0) and ran into some problems with the demo site right away. I can't find an issue about this and I am completely new so I just report my experience here.

I followed [All You Need to Know](https://getnikola.com/handbook.html#id17) to the letter installing nikola[extras] and just pressed enter to use all the defaults in the build wizard.

### broken links in ``1.rst``

    * `See a demo listing </stories/listings-demo.html>`__
    * `See a demo slideshow </stories/slides-demo.html>`__
    * `See a demo of the Bootstrap theme </stories/bootstrap-demo.html>`__

should be:

    * `See a demo listing </stories/listings-demo/index.html>`__
    * `See a demo slideshow </stories/slides-demo/index.html>`__
    * `See a demo of the Bootstrap theme </stories/bootstrap-demo/index.html>`__

### Local devserver URL inconsistency

``nikola serve -b`` opens http://0.0.0.0:8000 instead of http://127.0.0.1:8000 as mentioned [here](https://getnikola.com/handbook.html#id17) or http://localhost:8000 as mentioned in ``README.txt`` of the demo project.

### bash centric install command

This is not directly a nikola related problem, but while I am at it I should mention this also:

in zshell brackets are automatically interpreted by the shell, so calling ``pip install nikola[extras]`` fails with ``zsh: no matches found: nikola[extras]`` - which is not a helpful error message and might throw new users as well. It would be more portable to use ``pip install 'nikola[extras]'`` in the docs as this works in bash and zshell.

obestwalter:
Only the devserver inconsistency left now, but I don't know what would be preferable. I would use 127.0.0.1:8000 as the "safest" choice.

obestwalter:
I keep forgetting that windows exists, but I am pretty sure that the quotes will be o.k. there as well :)

obestwalter:
*sigh* windows does not like single quotes, but zsh is ok with double quotes, so now the install commands should be portable.

Kwpolska:
Thanks for caring to report all those problems!

> broken links in `1.rst`

This is caused by a very recent change in default site settings, and I forgot those links exist.  And our tests couldn’t catch it, because they use the `-q` settings, which do not include `PRETTY_URLS=True` yet…  ~~I’ll just use our `:doc:` directive to make it work properly in both setups.~~ Which is actually **broken** for PRETTY_URLS (#1878), luckily our magic `link://`s still work.

[...]

PR #1875, opened 2015-07-08
fix links

obestwalter:
as mentioned in https://github.com/getnikola/nikola/issues/1874#issue-93891351


Kwpolska:
Not necessarily.  See #1874 for details.

PR #1880, opened 2015-07-09
Fix #1879 -- use user’s HTML for custom teaser text

### LINKED ISSUE #1879: Custom INDEX_READ_MORE_LINK not working
I am adding an additional class for the "Read more" links using

    INDEX_READ_MORE_LINK = '<p class="more"><a class="label label-primary" href="{link}">{read_more}…</a></p>'

but it is not being applied to the output files. Using Nikola 7.6.0

Kwpolska:
This is #1879.

Are there any places where this could break?  This could bite, we might need some other solution (checking if INDEX_READ_MORE_LINK makes sense?)

cc @notfoss, @ralsina

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,8 @@ New in master
 Features
 --------
 
+* Use provided teaser format (``*_READ_MORE_LINK``) with custom teaser text
+  (Issue #1879)
 * Delete old ``bootstrap`` theme (use ``bootstrap3`` instead)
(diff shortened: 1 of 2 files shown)


notfoss:
> checking if INDEX_READ_MORE_LINK makes sense?

Can you elaborate on that? Do you want to do away with this option?

Kwpolska:
No, not at all.  The “check” would be to make sure the `read more…` link can render something sensible (it’s not empty).  That would be a corner case; we can probably just ignore it.

Merging.

PR #1886, opened 2015-07-11
Fix #1885 -- always return unicode in slugify

### LINKED ISSUE #1885: utils.slugify sometimes returns a str, sometimes a unicode
Debian Sid, Python 2.7.10.

    >>> import nikola.utils
    >>> a = nikola.utils.slugify(u'aaaaa')
    >>> type(a)
    <type 'str'>
    >>> a = nikola.utils.slugify(u'aaa-aa')
    >>> type(a)
    <type 'unicode'>

It seems, that if `slugify` does not replace anything, it will return a `str` instead of an `unicode`. This breaks the wordpress import because it calls `slugify` twice:

            for tag in self.all_tags:
                try:
                    tag_str = tag.decode('utf8')
                except AttributeError:
                    tag_str = tag
                tag = utils.slugify(tag_str)
                src_url = '{}tag/{}'.format(self.context['SITE_URL'], tag)
                dst_url = self.site.link('tag', tag)
                if src_url != dst_url:
                    self.url_map[src_url] = dst_url

(`site.link` calls `slugify` too)

Kwpolska:
This is #1885.

cc @evgeni

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Always return unicode in slugify (Issue #1885)
 * Remove logging handlers (Issue #1797)
 * Add ``-d``, ``--detach`` option to ``nikola serve`` (Issue #1871)
 * Use provided teaser format (``*_READ_MORE_LINK``) with custom teaser text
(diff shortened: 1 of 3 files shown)


evgeni:
+1

PR #1892, opened 2015-07-16
Fix getnikola/nikola-themes#49 -- https→http fallback

Kwpolska:
This is getnikola/nikola-themes#49.

cc @chacha92, @ralsina, @arusahni

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -33,6 +33,8 @@ Features
 Bugfixes
 --------
 
+* Use ``http`` as fallback in plugin/theme installers if a SSL error occurs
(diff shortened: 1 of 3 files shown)


ralsina:
I think before falling back to insecure HTTP, we should warn the user. Otherwise why bother with HTTPS in the first place?

Kwpolska:
I’d just add `press ^C to abort` — some people just don’t care, and the security-conscious will race to their keyboard anyway.

Kwpolska:
Even with `time.sleep(1)`. Makes sense?

ralsina:
Sure, that's enough.

PR #1897, opened 2015-07-17
Mock 1.1.4

ralsina:
Fiix a bunch of tests that were broken and passed by chance, fix a feature which HAD tests, but they passed by chance, too.


Kwpolska:
Guess we can merge.

PR #1918, opened 2015-07-30
added translation for bosnian [bs]

saleone:
Hi,
I added full translation for bosnian. Later I noticed later that this should have been done on [transifex](https://www.transifex.com/projects/p/nikola/) and I will go there and do it again (just so you can confirm this is valid translation) and I hope you can merge it then if not now.

### DIFF
--- a/translations/nikola.messages/bs.po
+++ b/translations/nikola.messages/bs.po
@@ -1,0 +1,133 @@
+# Messages in Nikola
+# Copyright (C) 2012-2013
+# This file is distributed under the same license as the Nikola package.
+# 
+# Translators:
+# Translators:
+# Translators:
+# Translators:
+# Translators:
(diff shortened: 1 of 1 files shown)


ralsina:
I'd rather you go through transifex, we have a script to update the code from there.

BTW: I noticed you asked for a huge number of translations, are they all needed?

Kwpolska:
Bosnian and Serbian (Latin) were accepted.  Any other languages do not make sense IMO and were deleted because they don’t make any sense (we hate country specifiers):

* English (United States)
* English (United Kingdom)
* Croatian (Croatia)
* Serbian (Serbia)
* Serbian (Latin) (Serbia)
* Bosnian (Bosnia and Herzegovina)
* Serbian (Ijekavian) — are these really relevant?
* Serbian (Ijekavian Latin)

Please upload your translations.

PR #1920, opened 2015-07-31
Fix #1900 -- render ipynb in listings

### LINKED ISSUE #1900: ipynb listings
It would be nice if I could throw an `.ipynb` file in the listings directory and have it render.

Issues:
* would anyone ever want to show off the raw (json) file?
* strange behavior, compared to other formats (including html)

Kwpolska:
This is #1900.

cc @ralsina, @damianavila

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,10 +1,16 @@
-New in Master
+New in master
 =============
+
+Features
+--------
+
+* Render Jupyter notebooks (ipynb) in listings (Issue #1900)
 
 Bugfixes
 --------
 
-* Handle non-integer shutter speeds in WP importer (Issue #1917)
+* Handle non-integer shutter speeds and other variables in WordPress
+  importer (Issue #1917)
 
 
 New in v7.6.2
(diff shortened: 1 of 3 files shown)


ralsina:
Well, it looks ok, and I trust you that it works but it's just special casing :-(

I am +0 but go ahead and merge it if you want.

damianavila:
Do you have somewhere an example to take a look? I did not use listing before, but IIRC, it puts the code snippet in your post an create a link to a page inside the listing folder... so I suspect that render OK on the listing page, but what about the snippet inside the post and how it renders if the notebook have markdown cells inside the snippet?

ralsina:
I havenot tried this but I suspect the listings directive will just show nicely highlighted json (or not even highlighted, depending on how it guesses). Which is ok, it just means you can't embed notebooks that way, just like before.

Kwpolska:
Yeah: listings **in posts** are going to show JSON, and it does not guess anything (you have to specify `json` yourself).  Listings can also be displayed standalone, and this is what this changes.

Standalone:
image

In post/page:
image

damianavila:
OK, and what about the start `start-after` and `end-before`, how that plays here? (sorry to not test by myself, just commenting on the fly, in the middle of some stuff).

Kwpolska:
Those don’t apply in standalone listings, only in posts (where you would just cut JSON in parts at random)

Issue #1922, opened 2015-08-02
Default to non-relative URLs?


Kwpolska:
Relative URLs can create problems, eg. missing assets. We could switch `URL_TYPE` to `full_path` (root-relative) for new sites to work around them.

Comments, opinions, counter-arguments?

@Aeyoun
cc @ralsina

Aeyoun:
Just to be clear: `<base href="${canonical}">`. This produces the same as using the web browser’s default fallback base URL even when the document is loaded from a path or even a host different from the canonical link (that is the point where resources and links start breaking).

* Relative URLs need a base element to compensate for over permissive path-matching on servers
* Root-relative only really benefit from having a base element when loaded from a different host (like localhost or a scraper)
* Relative and root-relative both load resources and have working links when documents are either downloaded to local computer OR served from a different server than it’s origin. (Minor.)
* Absolute URLs everywhere never see any benefit from having a base element

Recommendation alternative 1:
Keep current relative URLs and add a canonical link as base element. Smallest change.

Recommendation alternative 2:
Leave out base element and

Nikola’s canonical URL is already well-tested on all page types and are perfectly suited to be used as is in templates (well, except galleries because I don’t use them.)

Examples of URLs where all resources and links break without picking one of the above (hosted from Apache, nginx, or ISS default setup):
* /posts//example.html
* /errors/404.html loaded as /where/is/this/anyway
* /posts/pretty-url//

W3C’s link checker can be used for extensive automated testing of all anchors and links for anyone who want to test out setting a base element  on their site:
https://validator.w3.org/checklink

PS: People complaining that setting base break thing usually set the base to the canonical link’s dirname and not the canonical link itself.

ralsina:
"Relative and root-relative both load resources and have working links when documents are either downloaded to local computer OR served from a different server than it’s origin. (Minor.) "

That's "nikola serve" and "nikola auto", so, not minor.

I have not seen yet an example of where this is currently broken.

Aeyoun:
https://chriswarrick.com//fo/ur/oh/fo/ur
https://chriswarrick.com/blog/2015/07/06/upass//

ralsina:
The second one violates RFC 1630 AFAICS :-)

Aeyoun:
@ralsina, maybe. But then again http://apple.com/..///.././ipad//.//../

Aeyoun:
[...]

PR #1928, opened 2015-08-05
Comment out <base> elements when serving from auto

Aeyoun:
Restores local resource loading and URL resolution in `nikola auto` regardless of the outcome of #1922.

`nikola serve` is a different problem because of the inflexible Content-Length implementation there; although the solution would essentially look the same as this right here. Starting with `auto` before proceeding with `serve`.


ralsina:
LGTM add docstrings (see travis run) and merge at will :-)

PR #1932, opened 2015-08-06
Fix #1931

### LINKED ISSUE #1931: Yearly archivs not rebuilt when their posts change

ralsina:
Add reasonable (I think) config dependencies to the archives so they don't miss rebuilds.


Aeyoun:
Seems to work. :)

Issue #1906, opened 2015-07-23
Auto don’t like double-reload


Aeyoun:
1. Serve with auto
2. Open in browser and press ctrl+r, ctrl+r (or get impatient with auto’s iniated reload while rebuilding and hit ctrl+r just once as it also starts refreshing)

```pytb
Traceback (most recent call last):
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 138, in run
        self.finish_response()
      File "/usr/local/lib/python3.4/dist-packages/ws4py/server/wsgirefserver.py", line 71, in finish_response
        SimpleHandler.finish_response(self)
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 180, in finish_response
        self.write(data)
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 274, in write
        self.send_headers()
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 332, in send_headers
        self.send_preamble()
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 255, in send_preamble
        ('Date: %s\r\n' % format_date_time(time.time())).encode('iso-8859-1')
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 453, in _write
        self.stdout.write(data)
      File "/usr/lib/python3.4/socket.py", line 391, in write
        return self._sock.send(b)
    BrokenPipeError: [Errno 32] Broken pipe
    ----------------------------------------
    Exception happened during processing of request from ('127.0.0.1', 45759)
    Traceback (most recent call last):
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 138, in run
        self.finish_response()
      File "/usr/local/lib/python3.4/dist-packages/ws4py/server/wsgirefserver.py", line 71, in finish_response
        SimpleHandler.finish_response(self)
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 180, in finish_response
        self.write(data)
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 274, in write
        self.send_headers()
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 332, in send_headers
        self.send_preamble()
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 255, in send_preamble
        ('Date: %s\r\n' % format_date_time(time.time())).encode('iso-8859-1')
      File "/usr/lib/python3.4/wsgiref/handlers.py", line 453, in _write
        self.stdout.write(data)
[...]

PR #1935, opened 2015-08-09
Fix NameError caused by failed import in auto plugin.

okin:
After upgrading to 7.6.3 I got the following error message:

```
[2015-08-09T18:01:02Z] ERROR: yapsy: Unable to import plugin: /Users/niko/.virtualenvs/nikola/lib/python3.4/site-packages/nikola/plugins/command/auto
  File "/Users/niko/.virtualenvs/nikola/lib/python3.4/site-packages/yapsy/PluginManager.py", line 485, in loadPlugins
    candidate_module = imp.load_module(plugin_module_name,None,candidate_filepath,("py","r",imp.PKG_DIRECTORY))
  File "/Users/niko/.virtualenvs/nikola/lib/python3.4/imp.py", line 245, in load_module
    return load_package(name, filename)
  File "/Users/niko/.virtualenvs/nikola/lib/python3.4/imp.py", line 217, in load_package
    return methods.load()
  File "<frozen importlib._bootstrap>", line 1220, in load
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/Users/niko/.virtualenvs/nikola/lib/python3.4/site-packages/nikola/plugins/command/auto/__init__.py", line 443, in <module>
    f = WebSocketWSGIHandler.finish_response
NameError: name 'WebSocketWSGIHandler' is not defined
```

This was introduced through this fix for #1906: https://github.com/getnikola/nikola/commit/80a49adb373b0762560c473cff00daa0c50e502b
The introduced `WebSocketWSGIHandler` may not be available.

This PR avoids crashing the plugin.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Bugfixes
 --------
 
 * Ignore sliced multibyte characters when reading metadata for sitemaps
+* Fix NameError caused by failed import in auto plugin.
 
 New in v7.6.3
 =============
--- a/nikola/plugins/command/auto/__init__.py
+++ b/nikola/plugins/command/auto/__init__.py
@@ -447,7 +447,6 @@ try:
     else:
         EX = IOError
 
-
     def finish_response(self):
         """Monkeypatched finish_response that ignores broken pipes."""
         try:
@@ -459,4 +458,3 @@ try:
 except NameError:
     # In case there is no WebSocketWSGIHandler because of a failed import.
     pass
-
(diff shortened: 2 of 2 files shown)


ralsina:
Eh? But why would that be available? It's part of a dependency.

okin:
I do not have ws4py installed in this environment and this makes the import fail and get caught.

To trigger the error I just did a "nikola build" and this then seems to fail because the monkey patch is done on module level. 

Am 10. August 2015 05:03:03 MESZ, schrieb Roberto Alsina <notifications@github.com>:

ralsina:
Ah, ok then.

PR #1937, opened 2015-08-10
change wording in teaser comments

Wayneoween:
I was confused that this setting may change some other stuff too.
imho this describes the effect better.

### DIFF
--- a/nikola/conf.py.in
+++ b/nikola/conf.py.in
@@ -546,7 +546,7 @@ IMAGE_FOLDERS = {'images': 'images'}
 #     ("icon", "/icon_128x128.png", "128x128"),
 # )
 
-# Show only teasers in the index pages? Defaults to False.
+# Activate teasers in the index pages? Defaults to False.
 # INDEX_TEASERS = False
 
 # HTML fragments with the Read more... links.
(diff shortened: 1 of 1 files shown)


Kwpolska:
I'm not sure about this... "Show only teasers" makes sense IMO, where "Activate" is vague.

Wayneoween:
Ok, it was just a thought anyways and English is not my first language, so there may be that.

"Show only" sounds to me like it excludes teasers from somewhere else. Maybe "Use teasers" is better?

ralsina:
I agree with @Kwpolska ... if any native english speakers see this and can comment, much appreciated :-)

Kwpolska:
I believe we should phrase it like this:




# Show teasers (instead of full posts) in indexes/feeds? Defaults to False.




This makes it more clear, and also fixes awkward syntax (“Show only”).




PS. I actually have a very fancy certificate that proves I speak English on a near-native level (CEFR C2).





-- 

ralsina:
@Kwpolska hey, one of my favourite english writers was a pole :)

Kwpolska:
Who is it?


Also, would you mind implementing the revised description above?





-- 

ralsina:
@Kwpolska Jseph Conrad! Done.

PR #1942, opened 2015-08-14
Make Nikola Faster 1

ralsina:
* Add a plugin filtering step so we can NOT load some of them.
* Avoid loading compiler plugins we don't need because of configuration.
* Avoid loading plugins listed in DISABLED_PLUGINS
* Avoid loading compiler extensions for compilers we don't load

At this point this branch brings "nikola help" from 0.65 seconds to 0.31 in average of 100 runs.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in Master
 Bugfixes
 --------
 
+* Make Nikola startup faster by not loading useless plugins (Issue #1825)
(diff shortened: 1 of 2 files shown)


Kwpolska:
Does `nikola new_post -F` still work? All plugins should be disabled there...



-- 

ralsina:
@Kwpolska it works better when you are not in a site, strangely enough. I can fix it.

PR #1944, opened 2015-08-15
Refactor newpost (Fix #1943)

### LINKED ISSUE #1943: Nikola new_post -F works weird
Because of the selective plugin disabling it now misses a few formats when called with a conf.py

Plan:

- [ ] Keep the data about disabled compilers in the Nikola class
- [ ] Refactor new_post so all the standalone functions that always get the same arguments go away

ralsina:
There are some  minor changes, in that more plugins are not loaded, so for them friendlyName is not available.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in Master
 Bugfixes
 --------
 
+* Refactor new_post to match lazy plugin loading (Issue #1943)
 * Make Nikola startup faster by not loading useless plugins (Issue #1825)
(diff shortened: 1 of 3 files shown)


Kwpolska:
444-446 are old docs.

How does it look now, and could we patch in friendly names for the built-in compilers?

ralsina:
@Kwpolska we'd have to put friendly names in the plugin metadata, which makes sense.

Kwpolska:
Could you take care of this?



-- 

ralsina:
@Kwpolska sure

PR #1946, opened 2015-08-15
Add the option to configure TEASER_REGEXP

yamila-moreno:
For those who migrate from wordpress (thanks for the importer!), we have the "" tag for the "Read more" link. But the new posts will have ".. TEASER_END" as recommended.

So I'll end with two different "strings" to detect as TEASER_REGEXP.

For that reason, I suggest those changes:

- by default the behaviour is the same
- but if you need to use two (or more!!) different ways, you can create the regexp in the conf.py

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -86,6 +86,7 @@
 * `Tolu Sonaike <https://github.com/tolusonaike>`_
 * `Troy Toman <https://github.com/troytoman>`_
 * `Udo Spallek <https://github.com/udono>`_
+* `Yamila Moreno <https://github.com/yamila-moreno>`_
 * `Yaşar Arabacı <https://github.com/yasar11732>`_
 * `Yasuhiko Shiga <https://github.com/quoth>`_
 * `Zhaojun Meng <https://github.com/zhaojunmeng>`_
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,11 @@
 New in Master
 =============
 
+Features
+--------
+
+* New option to use custom, and several TEASER_ENDs
+
 Bugfixes
 --------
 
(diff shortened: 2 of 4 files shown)


ralsina:
Changing this to a regexp makes me nervous :-)
We could make it configurable but a string,  and you can just use "read more" ... would that be bad?

yamila-moreno:
Hi! I tried to make as few changes as possible. But I can change it, or sure. If I understood this correctly, you want something like:

TEASER_ENDS = '|'.join(['TEASER_END', 'more'])

and use this within the regular expression. Is that what you meant?

Thanks for the interest!! :D

ralsina:
Ah, no, I was just wrong :-)

Your change is totally ok, let's  merge it!

Can you add yourself to AUTHORS.txt and write a one-liner for CHANGES.txt?

yamila-moreno:
Ok! In a moment I'll push -f :D #yay

yamila-moreno:
Here you are the changes :+1:

ralsina:
Checking the code a bit more, you are setting TEASER_REGEXP in conf.py but the code in post.py will still use the one defined in post.py line 75.

You will also need to remove that line 75 and use self.config['TEASER_REGEXP'] instead of TEASER_REGEXP in Post.text() so you use the one you configured.

yamila-moreno:
Not sure if I understand what you mean, sorry.

In nikola/post.py I've added:

```
teaser_regexp = self.config.get('TEASER_REGEXP', TEASER_REGEXP)
```

with this line, I choose the conf.py value (if exists) or set the default value (line 75). From this point minimun changes to use the new teaser_regexp variable.

Where is still used the original TEASER_REGEXP?

ralsina:
You are right... not my best day to review code :-) Merging...

yamila-moreno:
hahahaha, better to be sure!! :dancer: 
Thanks for accepting the suggestion!

ralsina:
You're welcome, and thank you for helping the project :-)

PR #1957, opened 2015-08-20
fix #1956

### LINKED ISSUE #1956: handle include tag in mako templates
Currently templates used via include tags are not considered dependencies. It's not hard.

ralsina:
Add dependency on included files in mako templates (Fix #1956)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Features
 Bugfixes
 --------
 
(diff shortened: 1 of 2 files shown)


Kwpolska:
:+1:

Invariance needs a push, mind handling that and merging?

ralsina:
I did the invariance push already, it was because of a previous fix, this changes nothing. merging...

PR #1959, opened 2015-08-20
WIP: Add media extension for markdown

ralsina:
This doesn't really work at this point. Python Markdown requires ElementTree elements as output of the extension, and etree fails to parse some of micawber's data (like the youtube example in the docstring)

Leaving it here for when I or someone else figures it out.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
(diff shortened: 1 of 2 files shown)


ralsina:
Hey, this already exists! https://github.com/rennat/python-markdown-oembed

PR #1961, opened 2015-08-21
use https where appropriate

polyzen:


nikola/plugins/command/import_wordpress.py
```diff
@@ -801,7 +801,7 @@ def import_postpage_item(self, item, wordpress_namespace, out_folder=None, attac
         content = get_text_tag(
             item, '{http://purl.org/rss/1.0/modules/content/}encoded', '')
         excerpt = get_text_tag(
-            item, '{http://wordpress.org/export/1.2/excerpt/}encoded', None)
+            item, '{https://wordpress.org/export/1.2/excerpt/}encoded', None)
```
Kwpolska: Wrong.

nikola/plugins/command/import_wordpress.py
```diff
@@ -917,7 +917,7 @@ def import_postpage_item(self, item, wordpress_namespace, out_folder=None, attac
     def _extract_item_info(self, item):
         """Extract information about an item."""
         # The namespace usually is something like:
-        # http://wordpress.org/export/1.2/
+        # https://wordpress.org/export/1.2/
```
Kwpolska: Should not be.

nikola/plugins/task/galleries.py
```diff
@@ -587,7 +587,7 @@ def make_url(url):
             description='',
             lastBuildDate=datetime.datetime.utcnow(),
             items=items,
-            generator='http://getnikola.com/',
+            generator='https://getnikola.com/',
```
Kwpolska: There should be one more in `nikola/nikola.py`

Kwpolska: https://github.com/getnikola/nikola/blob/master/nikola/nikola.py#L1786

PR #1998, opened 2015-09-02
Fix 1983

ralsina:
Make tabcompletion work outside sites.


nikola/__main__.py
```diff
@@ -359,7 +360,7 @@ def run(self, cmd_args):
                     LOGGER.info('Did you mean "{}" or "{}"?'.format('", "'.join(best_sugg[:-1]), best_sugg[-1]))
             return 3
 
-        if sub_cmds[args[0]] is not Help and not isinstance(sub_cmds[args[0]], Command):  # Is a doit command
+        if not sub_cmds[args[0]] in (Help, Command, TabCompletion):
```
Kwpolska: This won’t work, `in` won’t catch Command subclasses.

PR #2005, opened 2015-09-03
Ensure link path to be URL encoded

masayuko:
This PR allow to use characters not permitted in URL. For example, in the case of tag, category and author not sluged appear in a generated path.

### DIFF
--- a/nikola/data/themes/base/templates/post_helper.tmpl
+++ b/nikola/data/themes/base/templates/post_helper.tmpl
@@ -15,7 +15,7 @@
         <ul itemprop="keywords" class="tags">
         %for tag in post.tags:
           % if tag not in hidden_tags:
-            <li><a class="tag p-category" href="${_link('tag', tag)}" rel="tag">${tag}</a></li>
+            <li><a class="tag p-category" href="${_link('tag', tag)}" rel="tag">${tag|h}</a></li>
           % endif
         %endfor
         </ul>
@@ -41,7 +41,7 @@
 
 <%def name="open_graph_metadata(post)">
 %if use_open_graph:
-    <meta property="og:site_name" content="${blog_title|striphtml}">
+    <meta property="og:site_name" content="${blog_title|striphtml,h}">
     <meta property="og:title" content="${post.title()[:70]|h}">
     <meta property="og:url" content="${abs_link(permalink)}">
     %if post.description():
@@ -55,14 +55,14 @@
     <meta property="og:type" content="article">
 ### Will only work with Pintrest and breaks everywhere else who expect a [Facebook] URI.
 ###    %if post.author():
-###    <meta property="article:author" content="${post.author()}">
+###    <meta property="article:author" content="${post.author()|h}">
 ###    %endif
     %if post.date.isoformat():
     <meta property="article:published_time" content="${post.formatted_date('webiso')}">
     %endif
     %if post.tags:
         %for tag in post.tags:
-           <meta property="article:tag" content="${tag}">
+           <meta property="article:tag" content="${tag|h}">
         %endfor
     %endif
 %endif
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -40,8 +40,9 @@ import natsort
 import mimetypes
 try:
     from urlparse import urlparse, urlsplit, urlunsplit, urljoin, unquote
+    from urllib import quote
 except ImportError:
-    from urllib.parse import urlparse, urlsplit, urlunsplit, urljoin, unquote  # NOQA
+    from urllib.parse import urlparse, urlsplit, urlunsplit, urljoin, unquote, quote  # NOQA
 
 try:
     import pyphen
@@ -493,6 +494,7 @@ class Nikola(object):
             'USE_FILENAME_AS_TITLE': True,
             'USE_OPEN_GRAPH': True,
             'USE_SLUGIFY': True,
+            'USE_URLENCODING': False,
             'TIMEZONE': 'UTC',
             'WRITE_TAG_CLOUD': True,
             'DEPLOY_DRAFTS': True,
@@ -1355,9 +1357,10 @@ class Nikola(object):
                 index_len = len(self.config['INDEX_FILE'])
                 if self.config['STRIP_INDEXES'] and \
                         link[-(1 + index_len):] == '/' + self.config['INDEX_FILE']:
-                    return link[:-index_len]
-                else:
-                    return link
+                    link = link[:-index_len]
+                if self.config['USE_URLENCODING']:
+                    link = quote(link.encode('utf-8'))
+                return link
             else:
                 return os.path.join(*path)
         except KeyError:
--- a/nikola/conf.py.in
+++ b/nikola/conf.py.in
@@ -1042,6 +1042,10 @@ UNSLUGIFY_TITLES = True
 # sometimes crash Nikola, your web server, or eat your cat.
 # USE_SLUGIFY = True
 
+# If you use URL encoding rather than UTF-8 encoding for URLs, you should
+# set this to true.
+# USE_URLENCODING = False
+
 # Templates will use those filters, along with the defaults.
 # Consult your engine's documentation on filters if you need help defining
 # those.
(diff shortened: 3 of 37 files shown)


nikola/data/themes/base/templates/post_helper.tmpl
```diff
@@ -41,7 +41,7 @@
 
 <%def name="open_graph_metadata(post)">
 %if use_open_graph:
-    <meta property="og:site_name" content="${blog_title|striphtml}">
+    <meta property="og:site_name" content="${blog_title|striphtml,h}">
```
Kwpolska: Is `,h` really necessary alongside `striphtml` and is it valid syntax?

nikola/nikola.py
```diff
@@ -39,9 +39,9 @@
 import natsort
 import mimetypes
 try:
-    from urlparse import urlparse, urlsplit, urlunsplit, urljoin, unquote
+    from urlparse import urlparse, urlsplit, urlunsplit, urljoin, unquote, quote
```
Kwpolska: `quote` is not available in Python 2.

Kwpolska:
Is URL encoding really needed? In 2015, where UTF-8 is the only encoding that exists, and we explicitly disallow `"` etc. in URLs (even with slugify disabled)?

nikola/nikola.py
masayuko: Oops, urllib in python 2.

nikola/data/themes/base/templates/post_helper.tmpl
masayuko: Yes. It is valid syntax. http://docs.makotemplates.org/en/latest/filtering.html
striphtml call markupsafe.Markup.striptags().
https://github.com/mitsuhiko/markupsafe/blob/master/markupsafe/__init__.py#L148
This method only remove |<[^>]*> blocks and unescape.

masayuko:
At least as far as I know, In Japanese environment with UTF-8, URL encoding surely has been used.
And I'd like to use slugify disabled name for author. ex. /output/authors/日本語 名前/. In this case, URL encoding is needed.

Kwpolska:
It should just work without encoding (or it should be handled by someone else), are you sure this is necessary? Please test using a Japanese link in Chrome and Firefox, with slugify disabled and without your patch.

masayuko:
I know modern browsers can treat Japanese link correctly.
If I enter "日本語.html" in the address bar, browser will request to server by URL encoding.
urlencodeja
Search engines treat URLs with utf-8 and URLs with URL encoding separately. This is bad for SEO.
URL encoding has been used to prevent this.

Is it should configurable in the conf.py?

masayuko:
I made it configurable. Please check this again.
I need this feature for japanese site.

Kwpolska:
I really don’t know if we should merge this, especially with the switch…

@Aeyoun: what’s your opinion?  Should this go in the core, or should this be something users would just do themselves (in custom templates)?

Aeyoun:
Adding `|h` in more places is probably good and will avoid issues like titles like `Get your <body> on a diet!`. Are we consistently either |h or |striphtml everywhere?

All URLs in Nikola are already compiling with “RFC-3986 standard for URIs, the RFC-3987 standard for IRIs, and the XML standard.” ([sitemap FAQ](http://www.sitemaps.org/faq.html#faq_xml_encoding)).

Nikola uses the same URL fromat in the sitemap as well as for link canonicalization so there is no confusion in the terms of search engines and duplicate content. So the SEO argument is something I wouldn’t worry about at all in this case. We’re already following best-practices. I’m not sure if any particular servers would do better with one format over the other. This is the only possible argument for changing this. This would have ether  be a problem everyone or no one, so there should be no need for an option to optionally break URLs.

From the attached screenshot, everything part of the browser (and Nikola) seem to be doing the correct thing. In summary, besides `|h`-ing all the things, I don’t see where this branch accomplishes anything.

Aeyoun:
TL;DR; unless we need to change URLs because of either RFC-3986 or RFC-3987, we shouldn’t change a thing with the URLs.

masayuko:
Please refer to https://support.google.com/webmasters/answer/35653?hl=en

Aeyoun:
That article covers the same RFCs as I referred to.

Aeyoun:
Maybe I’m misunderstand the issues here. I believe we already do use RFC-3986 / RFC-3987 (plus XML reserved-characters escaping) already and everywhere. If this is not the case, than that should be fixed and made non-optional. Updating these links should be done internally and not require any changes to any templates.

@masayuko, is Nikola not meeting RFC-3986 / RFC-3987 everywhere? nowhere? somewhere? what about sitemaps? atom/rss? If not, the link utilities that produced the links should be considered broken and need fixing. The link issue should not require template changes at all as Nikola should already be producing standard links.

We could add a new method `utils.urlencode()` and push every link through it before sending links be used in templates, feeds, sitemaps, etc.

masayuko:
The attached screenshot means the follow process is occured:
Request with UTF-8 encoded URL (user) -> URL encoding (brower) -> decode from URL encoding to UTF-8 (server)  -> UTF-8 (filesystem) -> Response
I know Nikola is already compatible with these RFC.

I hope the below part:
> Below is that same URL, UTF-8 encoded (for hosting on a server that uses that encoding) and URL escaped:

>  http://www.example.com/%C3%BCmlat.html&q=name

masayuko:
@Aeyoun At least, I hope sitemap, RSS/Atom, link tag for canonical is encoded with URL encoding. They have something to do with search engines.

Aeyoun:
The following method should produce correctly encoded links for anything that Nikola produces. (Limited to netloc and paths, because that’s all that Nikola controls.) I’ll make a new branch, stick this in `utils.py`, and make sure anything that produces a URL/IRI calls this method before returning it to a template, feed, sitemap, etc. It can be called multiple times, so I’ll add it to all existing `abs_link()` and other generic purpose link methods.

Opinions? @Kwpolska? @masayuko?

    import urllib.parse
    from collections import OrderedDict

    def encodelink(iri):
        link = OrderedDict(urllib.parse.urlparse(iri).__dict__)
        link['path'] = urllib.parse.quote(urllib.parse.unquote(link['path']))
        try:
            link['netloc'] = link['netloc'].encode('utf-8').decode('idna').encode('idna').decode('utf-8')
        except UnicodeDecodeError:
            link['netloc'] = link['netloc'].encode('idna').decode('utf-8')
        encoded_link = urllib.parse.urlunparse(link.values())
        return encoded_link.encode('utf-8')



Some examples, all producing the expected result. Why a method fulfilling this purpose isn’t already provided by `urllib.parse` is beyond me.

    >>> encodelink("/test")
    b'/test'
    >>> encodelink("/æøå")
    b'/%C3%A6%C3%B8%C3%A5'
    >>> encodelink("/test-æøå")
    b'/test-%C3%A6%C3%B8%C3%A5'
    >>> encodelink("/日本語.html?ninja")
    b'/%E6%97%A5%E6%9C%AC%E8%AA%9E.html?ninja'
    >>> encodelink("/test-%C3%A6%C3%B8%C3%A5")
    b'/test-%C3%A6%C3%B8%C3%A5'
    >>> encodelink("/æøå?test")
    b'/%C3%A6%C3%B8%C3%A5?test'
    >>> encodelink("http://æøå.no/a & b?1&2")
    b'http://xn--5cab8c.no/a%20%26%20b?1&2'
    >>> encodelink("http://æøå.no/æøå")
    b'http://xn--5cab8c.no/%C3%A6%C3%B8%C3%A5'
    >>> encodelink("http://elg.no/stein")
    b'http://elg.no/stein'
    >>> encodelink("http://xn--5cab8c.no/%C3%A6%C3%B8%C3%A5")
    b'http://xn--5cab8c.no/%C3%A6%C3%B8%C3%A5'

masayuko:
It's great and  thank you for understanding. I agree.

masayuko:
There is one thing that maybe need more encode('utf-8').

> link['path'] = urllib.parse.quote(urllib.parse.unquote(link['path']))

to

link['path'] = urllib.parse.quote(urllib.parse.unquote(link['path']).encode('utf-8'))

PR #2013, opened 2015-09-04
Fix #2011

### LINKED ISSUE #2011: Improve archive
When listing years or months, show the count of posts.

Example:

* 2004 (25)
* 2005 (23)

Or

* January (4)
* February (8)

jjconti:


Kwpolska:
what about _generate_posts_task for daily archives?

nikola/data/themes/base/templates/list.tmpl
```diff
@@ -8,8 +8,11 @@
     </header>
     %if items:
     <ul class="postlist">
-    % for text, link in items:
+    % for text, link, count in items:
```
Kwpolska: other lists might not contain the count thing, you should look into that (in `self.site.generic_post_list_renderer`?)

jjconti: I looked for uses of list.tmpl

Kwpolska: Okay, let’s merge this.

PR #2015, opened 2015-09-04
Fix KeyError occurred while creating sitemap.xml with non ascii filenames on python2

masayuko:


Kwpolska:
I’ll trust you blindly and hope this fixes it and does not introduce any new problems.

PR #2016, opened 2015-09-04
Make japanese catalog messages more appropriate

masayuko:


ralsina:
Sorry, but all translation work is done in transifex, could you redo this there?

masayuko:
OK. I have just submitted it in transifex.

Kwpolska:
Thanks! Why is *uncategorized* not translated?

masayuko:
If I translated it, *uncategorized* was 未分類.
The *Uncategorized* is used for the slug string of the default section name when POSTS_SECTION_FROM_META is True. URLs with UTF-8 encoded japanese string are not sutable for SEO. In Japanese, It is generally slugified to a string consist of ascii characters using ROMAJI.
For instance, *未分類* is slugified to *mibunrui* using ROMAJI. 
But *mibunrui* are unintelligible rather than *uncategorized* for Japanese.
So I didn't translate it.

My english is not good enough. I wonder if I can explain well.

Kwpolska:
Understood.

PR #2025, opened 2015-09-04
WIP Idea to avoid repeating ourselves so much in templates

ralsina:
1) add themes PARENT folders to template lookup
2) Use ``inherits parent-theme/templates/whatever.tmpl `` in ``whatever.tmpl``
3) If the template already inherits, it can namespace it and wrap trivial calls
4) Voilá, no need to repeat all of the parent theme's whatever.tmpl in the child theme.

This branch shows the simplified base_helper.tmpl in bootstrap3

You only have to do the silly redefinition-calling-parent for functions that are called from within that same file.

Problems:

* ~~jinjify needs to catch the reference to parent theme "base/" in this case and patch it to "base-jinja/"~~
* ~~In internationalized sites the link to the translations is styled wrong, no idea why~~
* jinjified themes should build

Looks like I have run into something where mako is just too different from jinja to make jinjify work :-(

There seems to be no way to make this work in Jinja2, because blocks are not callable!

```
{% import 'base-jinja/templates/post.tmpl' as basepost with context %}
{% extends 'base.tmpl' %}

{% block extra_head %}
{{ basepost.extra_head() }}
{% endblock %}
```

So, this means we should use macros instead of blocks, which means defs instead of blocks in Mako. May still be worth doing...

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,12 +5,17 @@ Features
 --------
 
 * Allow inheriting templates with theme name (Issue #1814)
+* Made TAG_PATH translatable (Issue #1914)
+* Made CATEGORY_PATH translatable (Issue #1914)
 * Display post counts for archive links (Issue #2011)
 * Document link/path handlers (Issue #2008)
+* Made DATE_FORMAT and JS_DATE_FORMAT translatable (Issue #2032)
 
 Bugfixes
 --------
 
+* Multi-lingual and multi-level directories confused section slug
+  detection (Issue #2023)
 * Use Unicode strings for WordPress comment headers (Issue #2019)
 * Don't add stories to author pages (Issue #2007)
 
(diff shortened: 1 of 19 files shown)


Kwpolska:
> In internationalized sites the link to the translations is styled wrong, no idea why

Maybe there was some special styling that is now missing?

ralsina:
@kwpolska yes, probably. If we decide this line of work is worth pursuing, I will surely fix that.

ralsina:
Apparently this will not work properly outside Mako and/or making it work right with Mako and Jinja means poking most of the templates in complex ways.

ralsina:
Oh, well, it works for def-heavy templates, that's much more than nothing :-)

Aeyoun:
What do you mean "def-heavy"? Would this be more useful if there were more defs in the base theme? I'm kind of planning of making more defs in the base theme, if that makes a difference. In a few weeks time, most likely. Testing out some stuff with my own site and thought I'd clean it up and introduce `base2` in some weeks time.

ralsina:
@Aeyoun for files that are a bunch of defs this is very usedul (see the change in nikola/data/themes/bootstrap3/templates/base_helper.tmpl), basically, only have to keep the functions you change, the rest will be "inherited".

For blocks, that doesn't work, so this technique doesn't help avoid repetition (fairly: it works in Mako, doesn't work in Jinja).

So, the more templates we have that are collections of defs, the more this avoids stupid repetition.

PR #2027, opened 2015-09-05
Optimize nikola check

felixfontein:
When `nikola check` is run with more than one flag, it runs `_call_nikola_list` several times to retrieve the same data. Adds a (optional) cache to avoid unnecessary calls.

This also fixes failure return in `_execute()`.

### DIFF
--- a/nikola/plugins/command/check.py
+++ b/nikola/plugins/command/check.py
@@ -168,12 +168,13 @@ class CommandCheck(Command):
             self.logger.level = 1
         else:
             self.logger.level = 4
+        failure = False
         if options['links']:
-            failure = self.scan_links(options['find_sources'], options['remote'])
+            failure |= self.scan_links(options['find_sources'], options['remote'])
         if options['files']:
-            failure = self.scan_files()
+            failure |= self.scan_files()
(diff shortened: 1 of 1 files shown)


nikola/plugins/command/check.py
```diff
@@ -46,7 +46,10 @@
 from nikola.utils import get_logger, STDERR_HANDLER
 
 
-def _call_nikola_list(site):
+def _call_nikola_list(site, cache=None):
+    if cache is not None:
+        if 'files' in cache and 'deps' in cache:
```
Kwpolska: Is this check necessary?

felixfontein: Yes, because this checks if the values are in the cache. (Just checking one of the two is fine as well.)

Kwpolska: I would just trust that a correct cache is given, but okay.

felixfontein: Somehow the cache must be filled when the values are not in there yet.

Kwpolska: I misunderstood the implementation slightly.

PR #2028, opened 2015-09-06
Fixing atom feed renderer for empty post lists.

felixfontein:
When the atom feed renderer is called with a empty `post` list, it crashes with an exception as it tries to call `formatted_date` of the last `post` object. Since it doesn't care about the `post` object, I moved most of the code of `Post.formatted_date` to `utils` and left a wrapper in `Post` which substitutes the post's date if it wasn't given.


Aeyoun:
LGTM.

Issue #1683, opened 2015-05-04
Max 49 998 files per sitemap file


da2x:
Theoretical issue, but at some point in time someone will run into this much sooner because of issue #1610. Sitemaps should include no more than 50 000\* URLs**.

Luckily, we already have a sitemapindex so we can just create more sitemaps.

The problem right now is that we don’t know how many URLs will be included in the sitemap. So we don’t know how many sitemaps/tasks should be created. The `urlset` is only available to the sitemap plug-in after the `_scan_locs` task has run, but this is too late for task creation.

\* In practice it’s 49 998. Don’t ask.
** Not `<url>` elements, but any kind of URL. Including schema, or localized link alternative, etc. Really should keep a running count  `://` in the sitemaps as they’re created to get it right.

ralsina:
Ugh. We can work around that by using a calculated dependency: 

http://pydoit.org/dependencies.html#calculated-dependencies

ralsina:
Aeyoun: would it be good enough to parse the sitemap using lxml, count the links and just warn people that Nikola's sitemap support is limited to 49998 links?

Kwpolska:
Wouldn’t it be good enough to just count links when we add them in `sitemap.py`?

ralsina:
Noone is going to fix this until someone actually runs into it, and then pays someone to fix it or does it himself:

* It's difficult
* It's a theoretical problem so far

PR #2030, opened 2015-09-06
Create directory-specific sitemaps for directories with 200+ files

Aeyoun:
Folders with 200+ HTML files get their own sitemaps that are included in the main sitemapindex.xml. These are not included in the main sitemap.xml file as they already exists in the folder-sitemaps. As all URLs for a folder reside within the same folder, they’re valid sitemaps in terms of search engines. Works nicely for sites with many tags (TAG_PATH fills up quickly) and should work for sectioned sites as well. The number 200 was entirely arbitrarily chosen in an attempt to keep the sitemaps somewhat small without having to create one for every folder with just a few files.

The motivation here was trying to make sitemaps smaller so sitemap consumers don’t need to download the entire growing main sitemap.xml file every time I add one page. It eats a lot of bandwidth. Plus it mitigates #1683 somewhat.

New sitemapindex follows best practices and passes sitemap tests at Bing, Yandex, Google, and Baidu. Well, at least I think it passes at Baidu. Big green checkmark and Chinese text means it’s okay, right?

### DIFF
--- a/nikola/plugins/task/sitemap/__init__.py
+++ b/nikola/plugins/task/sitemap/__init__.py
@@ -155,7 +155,7 @@ class Sitemap(LateTask):
                     sitemap_path = os.path.join(root, "sitemap.xml")
                 else:
                     sitemap_path = os.path.join(output_path, "sitemap.xml")
-                if not sitemap_path in urlset:
+                if sitemap_path not in urlset:
                     urlset[sitemap_path] = {}
                     sitemapindex[sitemap_path] = sitemap_format.format(self.site.abs_link(os.path.relpath(sitemap_path, output_path)), self.get_lastmod(sitemap_path))
 
--- a/nikola/plugins/task/sitemap/__init__.py
+++ b/nikola/plugins/task/sitemap/__init__.py
@@ -140,6 +140,9 @@ class Sitemap(LateTask):
         urlset = {}
 
         def scan_locs():
+            # No need to run this more than once
+            if urlset:
+                return True
             """Scan site locations."""
             for root, dirs, files in os.walk(output, followlinks=True):
                 if not dirs and not files and not kw['sitemap_include_fileless_dirs']:
@@ -290,6 +293,8 @@ class Sitemap(LateTask):
 
             return {'file_dep': file_dep}
 
+        scan_locs()
+
         yield {
             "basename": "_scan_locs",
             "name": "sitemap",
@@ -297,8 +302,6 @@ class Sitemap(LateTask):
         }
 
         yield self.group_task()
-
-        scan_locs()
 
         for sitemap_path in sorted(urlset.keys()):
             # TODO: Fix issue #1683
--- a/nikola/plugins/task/sitemap/__init__.py
+++ b/nikola/plugins/task/sitemap/__init__.py
@@ -135,7 +135,6 @@ class Sitemap(LateTask):
 
         output_path = kw['output_folder']
         sitemapindex_path = os.path.join(output_path, "sitemapindex.xml")
-        sitemap_path = os.path.join(output_path, "sitemap.xml")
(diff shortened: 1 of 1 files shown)


Aeyoun:
@ralsina @Kwpolska, line 301 of nikola/plugins/task/sitemap/__init__.py is probably a performance regression. Not sure how to generate the tasks before having scanned all of the directories?

ralsina:
@Aeyoun you can make your task depend scan_locs_task like write_sitemap does:

```         
"calc_dep": ["_scan_locs:sitemap"],
```

ralsina:
If I understand correctly, here is what we need to do.

1) Run a bunch of tasks that generate files
2) Run scan_locs that creates a list of all files in a folder, including those generated by those tasks
3) From that list, find folders with > 200 files, and for each of those create a task that will generate
    a local sitemap
4) Run those tasks, generate all the small local sitemaps
5) Done

I have looked at the code, and I don't know how to make this work correctly:

1) Just running scan_locs before generating the tasks is not good enough, because it runs *before* the other tasks run, so you will miss files.

2) Using scan_locs as a calc_dep is not good enough, because you won't know how many or what tasks to create.

So, I can see 2 ways to make this work:

1) A "task that generates more tasks", so that we could make taskA calc_dep in scan_locs, it returns a list of sitemaps to generate, then taskB takes that list and yields more tasks, one for each sitemap

2) Using doit's getargs in targets, and have one task that uses scan_locs data to decide what sitemaps we need, and then generates exactly those sitemaps.

@schettino72 any of those things possible?

felixfontein:
With the `earlytask_impl` branch, this is a piece of cake. Just use a task with a high enough stage number, its tasks will only be created when all other tasks with lower stage numbers are already done.

schettino72:
The easiest solution is to handle this logic yourself inside the task.

On step 4 "Run those tasks, generate all the small local sitemaps".
Instead of trying to dynamically create tasks only for folders that has more than 200 files,
create tasks for *all* folders. In these tasks' action check the logic (200 files or more) to see
if a local sitemap should really be created or not.

Remember that inside your action you can manipulate the task metadata.
http://pydoit.org/tasks.html#keywords-on-actions passing `task`.
You might need to change `targets` at run time in the action 
(and maybe more stuff to keep the folders with less than 200 files from running all the time).

And as @felixfontein said, early_task might be another way to solve this.

ralsina:
Ahhh changing targets inside one task that generates all the sitemaps looks like the easy way out.

Yes, early tasks would fix this but that's a backwards-incompatible change for us so it has to wait for v8 :-(

Aeyoun:
Well, we actually don’t know what all the folders will be either.

ralsina:
@Aeyoun just have one task, calc_dep on scan_locs, decide there what folders need sitemaps, and have it set its own targets list accordingly.

Aeyoun:
Hm. updating targets after running actions doesn’t really work. nikola.gen_tasks() as used by check.py is unaware of metadata that is updated from within a task’s action. So check.py will break. Other than that, it seem to work to update the metadata as part of running the task. Though it also introduced some build instability as scan_locs() and tasks that depend on it is unaware of any sitemap files by design.

felixfontein:
Changing the metadata later probably also break `nikola run <name of target>` for the later inserted targets.

ralsina:
So, @Aeyoun is this stalled? Impossible?

Aeyoun:
I can’t figure out how to do it, at least. The task is always created too early to know what other files will exist to accurately say which sitemaps will be generated when. I’ve tried some variations from the original idea but they all run into the same problem at some time or another.

Aeyoun:
This would have to generate it’s tasks and run only after everything else have already finished.

ralsina:
Ok, so basically this never worked, and it's for a minor corner-feature (if that's a thing). I say we close it.

PR #2031, opened 2015-09-06
Avoiding Nikola trying to cache posts which aren't written anyway.

felixfontein:
If `SHOW_UNTRANSLATED_POSTS` is set to `False` and a post's translation doesn't exist, Nikola nonetheless tries to create the cached post every time `nikola build` is run. Since it never creates it anyway (as it notices that the translation isn't available and `SHOW_UNTRANSLATED_POSTS` is set to `False` -- there's identical code to the one I inserted in `Post.compile` which simply returns in this condition), there's no need to schedule a task for it.

### DIFF
--- a/nikola/plugins/task/posts.py
+++ b/nikola/plugins/task/posts.py
@@ -77,6 +77,8 @@ class RenderPosts(Task):
             deps_dict = copy(kw)
             deps_dict.pop('timeline')
             for post in kw['timeline']:
+                if not post.is_translation_available(lang) and not self.site.config['SHOW_UNTRANSLATED_POSTS']:
+                    return
                 # Extra config dependencies picked from config
                 for p in post.fragment_deps(lang):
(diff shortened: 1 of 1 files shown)


nikola/plugins/task/posts.py
```diff
@@ -77,6 +77,8 @@ def tl_ch():
             deps_dict = copy(kw)
             deps_dict.pop('timeline')
             for post in kw['timeline']:
+                if not post.is_translation_available(lang) and not self.site.config['SHOW_UNTRANSLATED_POSTS']:
+                    return
```
Kwpolska: That won’t do, it will kill the task generator for **all** posts.  Try `continue`.

felixfontein: Yikes, that's what I wanted to do. Thanks for catching this!

PR #2034, opened 2015-09-07
Basic tag/category case sanitizing.

### LINKED ISSUE #2014: Wordpress allows case insensitive tags. nikola breaks when it tries to build them
I had `books` as well as `Books` in my tags and nikola frequently kept crashing while building the site.

felixfontein:
By default, uses first lower/uppercase spelling of a tag/category name it finds. Also allows to convert all to lower case.

See discussion in #2014.

(fixes #2014)


ralsina:
LGTM merging

PR #2047, opened 2015-09-09
Making path handlers of indexes plugin work before get_tasks() is run.

### LINKED ISSUE #2040: indexes plugin results in exception if index link is created too early.
The `indexes` plugin needs `self.number_of_pages` and `self.number_of_pages_section` to be initialized for its path handlers to work. If some other task plugin tries to create a link to a (section) index before `gen_tasks()` of the `indexes` plugins is called, it will raise an exception.

felixfontein:
Fixes #2040.

### DIFF
--- a/nikola/plugins/task/indexes.py
+++ b/nikola/plugins/task/indexes.py
@@ -55,12 +55,18 @@ class Indexes(Task):
         return super(Indexes, self).set_site(site)
 
     def _get_filtered_posts(self, lang, show_untranslated_posts):
+        """Return a filtered list of all posts for the given language.
+
+        If show_untranslated_posts is True, will only include posts which
+        are translated to the given language. Otherwise, returns all posts.
+        """
         if show_untranslated_posts:
             return self.site.posts
         else:
             return [x for x in self.site.posts if x.is_translation_available(lang)]
 
     def _compute_number_of_pages(self, filtered_posts, posts_count):
+        """Given a list of posts and the maximal number of posts per page, computes the number of pages needed."""
         return min(1, (len(filtered_posts) + posts_count - 1) // posts_count)
 
     def gen_tasks(self):
@@ -242,7 +248,7 @@ class Indexes(Task):
                                                                     self.site.config['INDEX_PATH'],
                                                                     index_file] if _f],
                                                      name,
-                                                     utils.get_displayed_page_number(name, self.number_of_pages, self.site),
+                                                     utils.get_displayed_page_number(name, number_of_pages, self.site),
                                                      lang,
                                                      self.site,
                                                      extension=extension)
--- a/nikola/plugins/task/indexes.py
+++ b/nikola/plugins/task/indexes.py
@@ -83,7 +83,6 @@ class Indexes(Task):
         }
 
         template_name = "index.tmpl"
-        posts = self.site.posts
         for lang in kw["translations"]:
             def page_link(i, displayed_i, num_pages, force_addition, extension=None):
                 feed = "_atom" if extension == ".atom" else ""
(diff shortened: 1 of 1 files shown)


nikola/plugins/task/indexes.py
```diff
@@ -46,12 +46,23 @@ class Indexes(Task):
...
         """Set Nikola site."""
+        self.number_of_pages = dict()
+        self.number_of_pages_section = {lang: dict() for lang in site.config['TRANSLATIONS']}
         site.register_path_handler('index', self.index_path)
         site.register_path_handler('index_atom', self.index_atom_path)
         site.register_path_handler('section_index', self.index_section_path)
         site.register_path_handler('section_index_atom', self.index_section_atom_path)
         return super(Indexes, self).set_site(site)
 
+    def _get_filtered_posts(self, lang, show_untranslated_posts):
```
Kwpolska: this function name should be more informative.  Also, using _underscores is not a valid reason to skip documenting things.

nikola/plugins/task/indexes.py
```diff
@@ -230,11 +233,16 @@ def index_path(self, name, lang, is_feed=False):
...
+            number_of_pages = self.number_of_pages[lang]
+        else:
+            number_of_pages = self._compute_number_of_pages(self._get_filtered_posts(lang, self.site.config['SHOW_UNTRANSLATED_POSTS']), self.site.config['INDEX_DISPLAY_POST_COUNT'])
+            self.number_of_pages[lang] = number_of_pages
         return utils.adjust_name_for_index_path_list([_f for _f in [self.site.config['TRANSLATIONS'][lang],
                                                                     self.site.config['INDEX_PATH'],
                                                                     index_file] if _f],
                                                      name,
-                                                     utils.get_displayed_page_number(name, self.number_of_pages[lang], self.site),
+                                                     utils.get_displayed_page_number(name, self.number_of_pages, self.site),
```
Kwpolska: Are you sure about this, in particular about the `self.` part?

felixfontein: Ah, `self` should go. But then, the arguments aren't needed anyway since for the given arguments, `utils.adjust_name_for_index_path_list` simply returns the first argument.
(In particular, computing the page numbers isn't needed for sections.)

nikola/plugins/task/indexes.py
```diff
@@ -253,11 +261,17 @@ def index_section_path(self, name, lang, is_feed=False):
             index_file = os.path.splitext(self.site.config['INDEX_FILE'])[0] + extension
         else:
             index_file = self.site.config['INDEX_FILE']
+        if name in self.number_of_pages_section[lang]:
+            number_of_pages = self.number_of_pages_section[lang][name]
+        else:
+            posts = [post for post in self._get_filtered_posts(lang, self.site.config['SHOW_UNTRANSLATED_POSTS']) if post.section_slug(lang) == name]
+            number_of_pages = self._compute_number_of_pages(posts, self.site.config['INDEX_DISPLAY_POST_COUNT'])
+            self.number_of_pages_section[lang][name] = number_of_pages
         return utils.adjust_name_for_index_path_list([_f for _f in [self.site.config['TRANSLATIONS'][lang],
```
felixfontein: Here, the arguments aren't needed for `utils.adjust_name_for_index_path_list` since for the given arguments, it simply returns the first argument. (In particular, computing the page numbers isn't needed for sections.)

nikola/plugins/task/indexes.py
felixfontein: Ah, that `pep257` tool doesn't complain about missing docstrings for underscore names apparently. Good to know.

felixfontein:
The failed check is due to some invariance breaking, which seems to not be the fault of this PR.

nikola/plugins/task/indexes.py
Aeyoun: Page numbers is indeed needed for sections.

nikola/plugins/task/indexes.py
Kwpolska: It believes they are private, which often makes sense, but it’s good to throw in a docstring nevertheless.

Kwpolska:
@Aeyoun, mind reviewing and merging?

nikola/plugins/task/indexes.py
felixfontein: Yes, but `utils.adjust_name_for_index_path_list` doesn't need it for the first page, and this function only provides a link to the first page :)

felixfontein:
Anyone minds if I merge?

felixfontein:
Thanks!

Issue #656, opened 2013-09-05
Handle bundling of plugin assets


ralsina:
Some plugins (localsearch) bring their own assets. Those are currently ignored by the bundling mechanism unless the user creates a custom theme.

ralsina:
Considering that bundling is no longer recommended with HTTP/2 and that users can always just create theirown bundle files, no need to actually *work* on this.

PR #2049, opened 2015-09-09
WIP Bundle assets from plugins (Issue #656)

ralsina:
Recovering an ancient branch I totally forgot about. If it can be made to work, cool.


ralsina:
Not needed.

PR #2053, opened 2015-09-10
Update bootswatch example with existing theme

mr337:
Working through the handbook applying a bootswatch theme results in a site that has no CSS. This is due to spruce theme not being available on bootswatch and downloading a GH 404 page rather than css file.

This PR simply updates the handbook to an existing theme for the next poor soul.


ralsina:
Thanks for the fix!

PR #2063, opened 2015-09-11
showoff a little using Dr. Nikola's Vendetta

ralsina:
This branch tries to show off some of the per-story / per-site customizations Nikola allows, using
a longish text shown as an e-book.

* Use custom template to make it book-like
* Use hyphenation to make text layout better
* Use JS animations to make it scroll smoothly
* Use FlowType to keep fonts optimally readable 
* 2/1 column adaptive layout (try to resize window)
* Google Fonts for nicer look

Looks better if you have pyphen installed and enable typogrify filters for HTML.

It's an example, not a real ebok reader. It's good enough :-)

Still to do:

- [x] Write a document explaining how this was done
- [ ] Announce it in a nice place
- [ ] Tune scrollbar (pain!)
- [x] Remove the fixed viewport height
- [x] The second image sometimes loses aspect ratio.
- [ ] On clicking on the book, make sure page scrolls so it's fully visible.
- [ ] On window resizing, make sure we always display on a page boundary

Add more TODOs if you notice anything / want anything


Kwpolska:
> Tune scrollbar (pain!)

What do you plan to do?

> Use JS animations to make it scroll smoothly

NO NO NO
Don’t fuck with users’ scrolling.  Instead, go to `chrome://flags` and enable smooth scrolling in your web browser.

ralsina:
@Kwpolska mostly just make the scrollbar less conspicuous. Maybe just use overflow tricks to hide it.

I only meant making it scroll smoothly when the user clicks to "turn the page", it's already there, check it out.

Kwpolska:
As of right now:

1. The title font looks ugly IMO
2. There should be a next/previous page button, clicking there is not immediately obvious
3. Images get squished at smaller window sizes
4. The project gutenberg links leads to a 403 forbidden error (PS. they use CherryPy!)
5. I think we might need to add some ©/PG info there
6. There should be more details on how this page works

ralsina:
@Kwpolska 

1. Yes, it needs a better title font, this was to remind me to get one (although it was something this sort of pulp would be typeset with!)

2. well, it's how all ebook readers work, but sure, why not :-)

3. They shoudn't after last commit

4. That's probably temporary. It used to work, and it's done in [the way they suggest](https://www.gutenberg.org/wiki/Gutenberg:Information_About_Linking_to_our_Pages#Canonical_URLs_to_Books_and_Authors)

5. Maybe, I did keep all the PG frontmatter

6. Totally, I want to write a whole tutorial

Kwpolska:
2\. My Kindle 4 doesn’t, it has actual physical buttons that I can press!
3\. They still do.

ralsina:
Oh, a fellow kindle-with-buttons fan! I am tempted to get a voyage just to get those back but with a better screen :-)

Kwpolska:
(The Kindle 4 is awesome, but it’s painful with a 23-character WiFi password…)

Kwpolska:
What’s the status of this?

ralsina:
After we publish the tutorial in the site, or at least decide it's final, I can backport it here.

PR #2070, opened 2015-09-11
Fix #2069 -- proper locale support for section links

### LINKED ISSUE #2069: post.section_link ignores languages
In [11]: site.timeline[30].section_link('pl')
Out[11]: '/nieskategoryzowane/'

The correct output should be '/pl/nieskategoryzowane/'. I’m using sections from metadata, but I think this is a problem with the path-based implementation, too.  I’ll attempt to fix it by myself.

Kwpolska:
This is #2069.

cc @Aeyoun, @ralsina.

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -745,11 +745,16 @@ class Post(object):
 
     def section_link(self, lang=None):
         """Return the link to the post's section."""
+        if lang is None:
+            lang = nikola.utils.LocaleBorg().current_lang
+
         slug = self.section_slug(lang)
+        t = os.path.normpath(self.translations[lang])
+        if t == '.':
+            t = ''
+        link = '/' + '/'.join(i for i in (t, slug) if i) + '/'
         if not self.pretty_urls:
-            link = urljoin('/' + slug + '/', self.index_file)
-        else:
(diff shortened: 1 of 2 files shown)


nikola/post.py
```diff
@@ -745,11 +745,16 @@ def section_color(self, lang=None):
 
     def section_link(self, lang=None):
```
felixfontein: Why isn't this simply using the `index_section_path` path handler of the `indexes` plugin?

Kwpolska: That’s a good question. I just fixed the existing code (by @Aeyoun). I looked a bit further, and it seems that the Post  class does not have access to the Nikola site object, and therefore cannot talk to `site.link()` nor to the plugin manager.

felixfontein: Maybe we should deprecate this function and encourate people to use `site.link()` instead. A `Post` object should know as little as possible about what other plugins are producing (well, except the `pages` and the `posts` plugins).

ralsina: +1 to just using link. We don't need to have the post access the Nikola object either, tests just need to be resolved after they are written in the document, when copying to output.

PR #2071, opened 2015-09-11
Making LocaleBorg thread-safe.

### LINKED ISSUE #2046: Problems when using parallel processing with threads and multiple languages
There's a problem with Nikola if your site has more than one language and if you use `nikola run -n4 -P thread` (run four tasks in parallel using threads instead of forking, see here: http://pydoit.org/cmd_run.html#parallel-execution). Namely, when rendering a page/..., `LocaleBorg` is storing the current language in global storage which is then used by `messages("...")` and `formatted_date()` (among others). When two tasks run in parallel for two different languages, one will overwrite the `LocaleBorg` setup of the other, and with high probability one of the output files ends up with wrong strings or wrongly formatted dates.
(When using forking, this should not be a problem, but that doesn't always work -- it might be problematic under Windows, and totally breaks for the `earlytask_impl` branch.)

I don't know how easy/complicated fixing this would be. There are two aspects:

  1. If we store the current language not in a global state, but a thread-local state, at least `messages("...")` will always use the correct language.

  2. This might not help for `formatted_date` and `get_month_name`, since they have to set the correct locale and then call some Python system function -- Python might not store the locale thread-local, but global. (Or does it store it thread-local? If that is guaranteed for both Python 2 and 3, using thread-local storage would be enough.) But this could be solved by using a global lock which avoids that `formatted_date` and `get_month_name` run in parallel.

What do yout think? While standard Nikola not run in parallel mode is totally not affected by this (and after all we don't advertise that you can run several tasks in parallel), I think that making Nikola more thread-safe is always a good idea.

felixfontein:
Fixes #2046 as discussed there.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -1078,6 +1078,8 @@ class LocaleBorg(object):
         cls.initialized = True
 
     def __get_shared_state(self):
+        if not self.initialized:
+            raise LocaleBorgUninitializedException()
         shared_state = getattr(self.__thread_local, 'shared_state', None)
         if shared_state is None:
             shared_state = {'current_lang': self.__initial_lang}
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -1074,8 +1074,15 @@ class LocaleBorg(object):
             encodings[lang] = encoding
 
         cls.encodings = encodings
-        cls.__shared_state['current_lang'] = initial_lang
+        cls.__initial_lang = initial_lang
         cls.initialized = True
+
+    def __get_shared_state(self):
+        shared_state = getattr(self.__thread_local, 'shared_state', None)
+        if shared_state is None:
+            shared_state = {'current_lang': self.__initial_lang}
+            self.__thread_local.shared_state = shared_state
+        return shared_state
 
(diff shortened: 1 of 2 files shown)


felixfontein:
The failed test is due to a gist not being loaded. I've seen this error several times appearing and disappearing over the last hours (for different branches), so I assume it's not my fault ;)

felixfontein:
I did some more testing, it seems to work fine. I also did some timing tests with one of my smaller blogs; the new code with a thread local variable and a lock didn't result in a noticeable performance drop (11.1 vs. 11.2 seconds for `master` vs `threadsafe-localeborg` in single-threaded mode), and this performance drop was the same when I replaced the `Lock()` object with an object with empty `__enter__` and `__exit__` methods and the thread-local storage with a simple shared state; so it is probably more the fault of `formatted_date` calling `__set_locale` and/or `current_lang` calling `__get_shared_state` instead of directly accessing it.

ralsina:
LGTM let's merge it and see.

felixfontein:
Thanks for merging!

PR #2072, opened 2015-09-11
Making sure all parts of an Atom feed is rendered for the correct language.

felixfontein:
For example, sometimes tags for a wrong language were inserted.


Aeyoun:
LGTM. Didn’t think about the fact that multilingual sites has different tags per language. I think I’m the only one who don’t have any multilingual sites built in Nikola.

felixfontein:
You definitely weren't the only one until two weeks ago ;-)

PR #2073, opened 2015-09-12
HTML escape in more places

masayuko:


nikola/plugins/template/mako.py
```diff
@@ -130,8 +129,3 @@ def template_deps(self, template_name):
                 deps += self.template_deps(fname)
             self.cache[template_name] = tuple(deps)
         return list(self.cache[template_name])
-
-
-def striphtml(text):
```
Aeyoun: Leave this in place. Even if we don’t use it by default. It can be useful in some contexts and we don’t know exactly what every theme author will want to do. Also, removing it will break all existing themes. Yikes.

Aeyoun:
Nice that we consistently use `|h` everywhere.

masayuko:
Thank you for checking.

Kwpolska:
I think we can merge this, any objections?

PR #2075, opened 2015-09-12
add new normalize_html filter, make typogrify call it, fixes #2064

### LINKED ISSUE #2064: The "filters" metadata typogrify doesn't work?
I tried to apply typogrify that way and could not. Either it doesn't work or it needs better docs.

So, according to the code, this should use the typogrify filter:

```
.. filters: filters.typogrify
```

And ... it does! Sort of.

1) In cache/posts/foo.html -> the post has smart quotes
2) In output/posts/foo/index.html? Nope.
3) site.timeline[0].text()? No quotes

~~Found the guilty party: LXML~~ NOPE

Now for real. This is what happens:

``smartypants('hello "world"')`` works.

But!

``smartypants('hello &quot;world&quot;')`` DOES NOT WORK.

ralsina:


docs/manual.txt
```diff
@@ -1351,6 +1351,10 @@ typogrify_sans_widont
 minify_lines
    **THIS FILTER HAS BEEN TURNED INTO A NOOP** and currently does nothing.
 
+normalize_html
```
Kwpolska: missing `_`

nikola/filters.py
```diff
@@ -302,3 +305,15 @@ def jsminify(data):
...
+def _normalize_html(data):
+    """Pass HTML through LXML to clean it up, if possible."""
+    try:
+        data = lxml.html.tostring(lxml.html.fragment_fromstring(data), encoding='unicode')
+    except:
+        pass
+    return data
+
+
+normalize_html = apply_to_text_file(_normalize_html)
```
Kwpolska: why isn’t that a decorator?

Kwpolska: Oh, wait. I get it now.  Disregard both comments.

Kwpolska:
:+1:

docs/manual.txt
ralsina: No, _normalize_html is internal, works on strings, the one without the underscore is a proper filter that works on files :-)

Issue #2095, opened 2015-09-17
Include meta['previewimage'] in Atom and RSS


Aeyoun:
Better Feedly and Apple News integration. Investigate best format.

PR #2100, opened 2015-09-18
New option FEED_PREVIEWIMAGE includes post previewimage meta

Aeyoun:
Issue #2095.

Checks to see if the previewimage is found in the content before including it. Avoids duplicate imagery. `<figure></figure>` is recommended [by Flipboard](https://about.flipboard.com/rss-spec/) and supported by Apple News and Feedly. There is very little we can do about optimizing here as every feed consumer recommends different file formats, sizes, markup, and whatnot.

* Deprecated RSS_PLAIN in favor of FEED_PLAIN for Atom and RSS
* Deprecated RSS_TEASERS in favor of FEED_TEASER for Atom and RSS


Kwpolska:
What about RSS_READ_MORE_LINK?

-- 

Kwpolska:
Also, why is this an option? We could probably just force it on everyone.
Most people won't mind.

-- 

Aeyoun:
> What about RSS_READ_MORE_LINK?

Considering removing it again. Last year, all the big sites were doing this. Now, no one is doing it anymore and I can’t figure out why they stopped.

> Also, why is this an option?

Some sites may include an `img` in their teasers already, depending on their site designs. Thought I’d give folks the options. Can flip the default to True, though.

Kwpolska:
Let’s:

* leave both in
* default to `FEED_PREVIEWIMAGE = True`
* rename `RSS_READ_MORE_LINK` to `FEED_READ_MORE_LINK` and make sure it works in Atom, too

Aeyoun:
It does work with Atom. I’ll rename and change defaults. See I set the wrong default for FEED_PLAIN too.

PR #2108, opened 2015-09-21
Fix 2106

ralsina:
Correct usage of get_asset_path, fixes false positive orphan reports. Also fixed a couple of related integration tests.


CHANGES.txt
```diff
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
+* Fix false orphan report for code.css and robots.yxy (Issue #2106)
```
Kwpolska: txt*

Kwpolska:
:+1: works for me

PR #2110, opened 2015-09-22
Fixing #2109 (bug in get_asset_path).

### LINKED ISSUE #2109: Possible bug in `get_asset_path`?
As mentioned in #2106, I think `nikola.util.get_asset_path` might have a bug, or it is used incorrectly. Or (third possiblity), I understand it wrongly what it is supposed to do.

I understand it as follows: if you want to know whether `output/bla/test.txt` is created by either the themes or by copying files (or already created by another plugin or by magic when called in `gen_tasks`), you do `get_asset_path('bla/test.txt', themes, files_folders)` and see if it returns `None` or a path.

I looked at how it is used for example in the `robots` plugin to detect whether `output/robots.txt` is created, and it seems to confirm my idea.

If that's correct, then it will fail in the following situation: assume that you have `file_folders = {'a': 'b'}` and have a file `a/bla/test.txt`. This will be copied to `output/b/bla/test.txt`. On the other hand, `get_asset_path('bla/test.txt', themes, files_folders)` will return `a/bla/test.txt`, so I assume that `output/bla/test.txt` will be created -- which it won't.

So -- do I understand it wrong what `get_asset_path` does, or is there a bug in its implementation? (Or am I somehow else confused? :) )

felixfontein:
Fixes #2109. Also fixes a bug in the docstring, and adds one more example output to make clear what it really does.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -1001,7 +1001,7 @@ def get_asset_path(path, themes, files_folders={'files': ''}, _themes_dir='theme
         if os.path.isfile(candidate):
             return candidate
     for src, rel_dst in files_folders.items():
-        relpath = os.path.relpath(path, rel_dst)
+        relpath = os.path.normpath(os.path.relpath(path, rel_dst))
         if not relpath.startswith('..' + os.path.sep):
             candidate = os.path.abspath(os.path.join(src, relpath))
             if os.path.isfile(candidate):
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -986,8 +986,11 @@ def get_asset_path(path, themes, files_folders={'files': ''}, _themes_dir='theme
     >>> print(get_asset_path('nikola.py', ['bootstrap3', 'base'], {'nikola': ''}))
     /.../nikola/nikola.py
 
-    >>> print(get_asset_path('nikola/nikola.py', ['bootstrap3', 'base'], {'nikola':'nikola'}))
+    >>> print(get_asset_path('nikola.py', ['bootstrap3', 'base'], {'nikola': 'nikola'}))
     None
+
+    >>> print(get_asset_path('nikola/nikola.py', ['bootstrap3', 'base'], {'nikola': 'nikola'}))
+    /.../nikola/nikola.py
 
     """
     for theme_name in themes:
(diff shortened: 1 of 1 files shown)


nikola/utils.py
```diff
@@ -986,9 +986,12 @@ def get_asset_path(path, themes, files_folders={'files': ''}, _themes_dir='theme
     >>> print(get_asset_path('nikola.py', ['bootstrap3', 'base'], {'nikola': ''}))
     /.../nikola/nikola.py
 
-    >>> print(get_asset_path('nikola/nikola.py', ['bootstrap3', 'base'], {'nikola':'nikola'}))
+    >>> print(get_asset_path('nikola.py', ['bootstrap3', 'base'], {'nikola': 'nikola'}))
     None
 
+    >>> print(get_asset_path('nikola/nikola.py', ['bootstrap3', 'base'], {'nikola': 'nikola'}))
+    /.../nikola/nikola.py
```
ralsina: /.../ ???

nikola/utils.py
```diff
@@ -998,9 +1001,11 @@ def get_asset_path(path, themes, files_folders={'files': ''}, _themes_dir='theme
         if os.path.isfile(candidate):
             return candidate
     for src, rel_dst in files_folders.items():
-        candidate = os.path.abspath(os.path.join(src, path))
-        if os.path.isfile(candidate):
-            return candidate
+        relpath = os.path.relpath(path, rel_dst)
+        if not relpath.startswith('..' + os.path.sep):
```
ralsina: If you are going to check this, you may also want to add a normpath call there.

nikola/utils.py
felixfontein: `/.../` stands for "some random path on my hdd which I don't want to paste here", like in the already existing examples.

Kwpolska: that’s a doctest ellipsis, not a real path

nikola/utils.py
felixfontein: Good point.

nikola/utils.py
ralsina: The others are actually doctests.

felixfontein: Isn't the new one a doctest, too? Or is some special magic needed to make it a doctest as well?

Kwpolska: This is a doctest, as it begins with `>>>` — py.test should pick it up (whether or not doctests are good is another thing)

felixfontein: And it passes; if I modify the path, `doit doctest` fails.

ralsina: Awsome then :-)

ralsina:
Thanks for fixing this Felix!

Issue #2086, opened 2015-09-15
Deprecate `Post.section_link()`


Kwpolska:
`site.link()` aka `link://` should be used instead.

(via #2069)

PR #2111, opened 2015-09-22
WIP: Deprecate `Post.section_link()`.

felixfontein:
WIP for #2086.

Where should this go in `CHANGES.txt`? Is this a feature or a bug? :)

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -745,6 +745,8 @@ class Post(object):
 
     def section_link(self, lang=None):
         """Return the link to the post's section (deprecated)."""
+        utils.LOGGER.warning("Post.section_link is deprecated. Please use " +
+                             "site.link('index_section_path', post.section_slug()) instead.")
         if lang is None:
             lang = nikola.utils.LocaleBorg().current_lang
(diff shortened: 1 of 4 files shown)


nikola/post.py
```diff
@@ -744,7 +744,7 @@ def section_color(self, lang=None):
         return utils.colorize_str_from_base_color(slug, base)
 
     def section_link(self, lang=None):
-        """Return the link to the post's section."""
+        """Return the link to the post's section (deprecated)."""
         if lang is None:
```
Kwpolska: could use a deprecation warning (`logger.warning` if we can access that)

Kwpolska:
Add a different changelog section for this, at the end, named *Deprecations*, I guess?

Also, please make a commit with `fix #2086` in the message.

nikola/post.py
felixfontein: `utils.LOGGER` is available.

Kwpolska:
Thanks!

felixfontein:
You're welcome.

PR #2117, opened 2015-09-23
Fix #2114

### LINKED ISSUE #2114: .. template: directive is not used in translated pages (.en.rst)
I have a multilanguage site and I have some pages (under `stories/`) which are translated. For example:

* home.rst
* home.en.rst

(the repo is here: https://github.com/humitos/argentinaenpython.com.ar)

If I use `.. template: englishpage.tmpl` in the `home.en.rst` doesn't work. Nikola just use the same template to render both files.

ralsina:
Totally untested. @humitos can check if this works ;-)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
+* Support setting template in metadata in translated posts (Issue #2114)
 * Fix false orphan report for code.css and robots.txt (Issue #2106)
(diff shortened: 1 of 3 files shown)


humitos:
I doesn't work.

For some reason "lang" is never "en" or other thing different than "es".

I printed all the `template_name` and it never appears the one used in the `.en.rst` file ;(

humitos:
It worked! ;D

Thanks!

PR #2129, opened 2015-10-01
Deprecation warning for Python 2

ralsina:


nikola/__main__.py
```diff
@@ -64,6 +64,20 @@
 # DO NOT USE unless you know what you are doing!
 _RETURN_DOITNIKOLA = False
 
+PY2_WARNING = """Nikola is going to deprecate python 2 support in 2016. You already have python 3
+available in your system. Why not switch?
+
+Please check https://getnikola.com/blog/env-survey-results-and-the-future-of-python-27.html for details."""
```
Kwpolska: We could use a shorter link.

Also, Python* should be uppercase

nikola/__main__.py
```diff
@@ -64,6 +64,25 @@
 # DO NOT USE unless you know what you are doing!
 _RETURN_DOITNIKOLA = False
 
+PY2_WARNING = """Nikola is going to deprecate Python 2 support in 2016. You already have Python 3
```
Kwpolska: > You most likely have Python 3 available on your system already! (And if you don’t, you can install it really easily.)

ralsina: I actually intended to check if it's available, and use a different message if it's  not.

nikola/plugins/task/py3_switch.py
```diff
@@ -0,0 +1,105 @@
...
+
+PY2_AND_NO_PY3_WARNING = """Nikola is going to deprecate Python 2 support in 2016. Your current
+version will continue to work, but please consider upgrading to Python 3.
+
+Please check http://bit.ly/1FKEsiX for details.
+"""
+PY2_WARNING = """Nikola is going to deprecate Python 2 support in 2016. You already have Python 3
+available in your system. Why not switch?
+
+Please check http://bit.ly/1FKEsiX for details.
```
Carreau: I would avoid shortened URL here. People will copy-past anyway, and you never know if bit.ly will keep it infinitely.
If you keep it you can probably make a custom name like `bit.ly/drop-Python-2` which is easier to remember.

nikola/plugins/task/py3_switch.py
```diff
@@ -0,0 +1,105 @@
...
+    name = "switch to py3"
+
+    def gen_tasks(self):
+        """Beg the user to switch to python 3."""
+        def give_warning():
+            if sys.version_info[0] == 3:
+                return
+            if has_python_3():
+                LOGGER.warn(random.choice(PY2_BARBS))
+                LOGGER.warn(PY2_WARNING)
```
Carreau: Want to sleep N second with `N = f(timedelta since python 3 release)` ?

nikola/plugins/task/py3_switch.py
Kwpolska: Worst case scenario, we could create a redirect on `getnikola.com/py2` (probably via nginx config, not `conf.REDIRECTIONS`) and use that.

PR #2135, opened 2015-10-07
Improve searchability (SEO) of Nikola on PyPI.

schettino72:


ralsina:
Thanks @schettino72 !

PR #2151, opened 2015-10-20
Ignore .cache dir

bameda:


Kwpolska:
We don’t have that directory, but maybe your editor creates it.  Let’s just get rid of it.

PR #2152, opened 2015-10-21
Update ghp import

topiaruss:
In my use on OSX, python3, there were encoding issues in setup.py for the older ghp-import during pip install.   ghp-import2 is supposed to fix that and other packaging issues so I'm hoping you can upgrade.

I've made the appropriate change in requirements-extras.txt, and tried to run tests. I immediately hit a problem related to parsing the results of `locale -a`, which are different on OSX, so fixed that issue.

I see other test failures due to temporary directories ('/var/folders') and hyphenator. 

Can you test this PR in your environment, and merge it if it works for you.  Sorry, don't have time to fix anymore tests at the moment.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -41,7 +41,7 @@ def task_locale():
             locales = []
             languages = set()
             for line in out.splitlines():
-                if line.endswith('.utf8') and '_' in line:
+                if (line.endswith('.utf8') or line.endswith('.UTF-8')) and '_' in line:
                     lang = line.split('_')[0]
                     if lang not in languages:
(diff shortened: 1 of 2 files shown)


Kwpolska:
We don’t have any tests for `ghp-import`. Assuming nothing has changed with the invocation of `ghp-import2`, Nikola should work fine.  Are you aware of any changes?

topiaruss:
I'm not aware of changes, but neither have I exercised any related code. Perhaps you can find someone who uses github pages, and get them to run some tests.

PR #2157, opened 2015-10-26
fix #2156

### LINKED ISSUE #2156: ${messages("xxx")} not working in post-list template
When I use ${messages("xxx")} tag inside post-list template. I get an error during site compilation.

Error message is the following : File "/opt/nikola/projects/konilabs-web/cache/.mako.tmp/post_list_thumb.tmpl.py", line 72, in render_content __M_writer(unicode(messages("Read More"))) TypeError: 'Undefined' object is not callable

ralsina:
Pass messages to the post list template

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -15,6 +15,7 @@ Features
 Bugfixes
 --------
 
+* Pass messages to post-list template (Issue #2156)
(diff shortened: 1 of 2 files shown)


ralsina:
Confirmed fix by Koni on IRC.

PR #2159, opened 2015-10-27
Rename RSS_FEED to feed, in order to avoid confusion

lekum:
**RSS_FEED** could be interpreted as a valid variable, so I recommend using *feed* or a similar lowercase word (and it is even more technology-agnostic).


Kwpolska:
feeds* even

PR #2172, opened 2015-11-15
Better support for german special characters characters in slugify

akrist:


Kwpolska:
:-1:

This **cannot** be merged as it stands. The letters ä/ö/ü are present in languages other than German. And those other languages don’t use (or even forbid) `ae`/`oe`/`ue` as the replacements. They might be fine with `a`/`o`/`u`, or have their own replacements.

The next idea would be “check the current language and act accordingly”. And this is complicated, because we use `slugify` a lot, and we might depend on its value not changing between runs (for example if we need to slugify a file name, and don’t have a list of slugifications made, which we don’t, especially over multiple runs of Nikola). Which would make things even worse.

I’m afraid I have to reject this.

(PS. ß is already slugified as `ss`.)

Aeyoun:
There is already a setting for using Unicode in slugs. Use that instead.

felixfontein:
How about modifying Nikola that `lang` is passed to all calls of `slugify` (with by default getting it from `LocaleBorg`)? It might be harder to implement this, but I think it's worth trying.

Kwpolska:
How about no? It’s a really minor thing most people don’t care about, and we could still fuck up if we did not notice something using locale-dependent `slugify` where it’s inappropriate.

PR #2178, opened 2015-11-21
Fix #1938 -- Jupyter compatibility

### LINKED ISSUE #1938: The `IPython.nbconvert` package has been deprecated.
IPython 4.0 was just released (http://blog.jupyter.org/2015/08/12/first-release-of-jupyter/).  The `IPython.nbconvert` package has been deprecated. The [`nbconvert`-based ipynb compiler](https://github.com/getnikola/nikola/blob/15217bc93f0af0d70ffe33e0ea067d81ddf32403/nikola/plugins/compile/ipynb.py) should be updated where necessary to reflect any API changes in the 4.0 release.

Kwpolska:
This is #1938.  cc @damianavila, @ltiao.

### DIFF
--- a/requirements-extras.txt
+++ b/requirements-extras.txt
@@ -7,7 +7,7 @@ pygal>=2.0.0
 typogrify>=2.0.4
 phpserialize>=1.3
 webassets>=0.10.1
-ipython[notebook]>=2.0.0
+jupyter>=1.0.0
 ghp-import2>=1.0.0
 ws4py==0.3.4
 watchdog==0.8.3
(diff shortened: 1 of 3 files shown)


requirements-extras.txt
```diff
@@ -7,7 +7,7 @@ pygal>=2.0.0
 typogrify>=2.0.4
 phpserialize>=1.3
 webassets>=0.10.1
-ipython[notebook]>=2.0.0
+jupyter>=1.0.0
```
damianavila: `notebook` should be the dependency... jupyter "metapackage" contain a lot of stuff which is not a dependency for the functionality...

damianavila:
Changes seems ok... I would say that sooner than later we need to deprecate the old API... but let's wait a little bit to do that...

requirements-extras.txt
Kwpolska: Fixed.

PR #2184, opened 2015-12-04
Make PEP8 compliant

habi:
As part of http://24pullrequests.com/ and to test the waters with the
Nikolai team

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -11,7 +11,8 @@ DOIT_CONFIG = {
 
 
 def recursive_glob(path, pattern):
-    """recursively walk path directories and return files matching the pattern"""
+    """recursively walk path directories and return files matching the
+    pattern"""
     for root, dirnames, filenames in os.walk(path, followlinks=True):
         for filename in fnmatch.filter(filenames, pattern):
             yield os.path.join(root, filename)
@@ -23,6 +24,7 @@ def task_flake8():
         'name': os.path.join(os.getcwd(), 'nikola'),
         'actions': ['flake8 nikola/'],
     }
+
 
 def task_pep257():
     """pep257 -- static check for docstring style"""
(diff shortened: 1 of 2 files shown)


dodo.py
```diff
@@ -11,7 +11,8 @@
 
 
 def recursive_glob(path, pattern):
-    """recursively walk path directories and return files matching the pattern"""
+    """recursively walk path directories and return files matching the
```
Kwpolska: That does not match PEP257 though, which is another important standard.

dodo.py
```diff
@@ -41,7 +43,8 @@ def set_nikola_test_locales():
             locales = []
             languages = set()
             for line in out.splitlines():
-                if (line.endswith('.utf8') or line.endswith('.UTF-8')) and '_' in line:
+                if (line.endswith('.utf8') or line.endswith('.UTF-8')) and \
```
Kwpolska: We prefer to use (parentheses) for continuations.

setup.py
```diff
@@ -90,7 +90,8 @@ def expands_symlinks_for_windows():
     sys.path = oldpath
     del sys.modules['winutils']
     if failures != -1:
-        print('WARNING: your working copy is now dirty by changes in samplesite, sphinx and themes')
+        print('WARNING: your working copy is now dirty by changes in'
+              'samplesite, sphinx and themes')
```
Kwpolska: Missing a space between the two lines..

Kwpolska:
Also, we don’t honor the 79 character limit.

habi:
I've fixed the missing space and reverted changes in dodo.py.
Hope that still helps.

Kwpolska:
Yeah, thanks for your contribution! :clap:

habi:
Very welcome.
You mention that you don't honour the 79 characters limit.
I'll try to push some changes for `pep8 */*.py --ignore=E501` if that's OK.

Kwpolska:
Actually, we [already do that](https://github.com/getnikola/nikola/blob/master/setup.cfg#L5).

habi:
Didn't see that yet. There are bunch of PEP8 errors in the `tests` folder. I'll fix them and make another PR.

PR #2188, opened 2015-12-18
fix #2166

### LINKED ISSUE #2166: Use a FS cache for Jinja to speed it up a bit
http://jinja.pocoo.org/docs/dev/api/#bytecode-cache

ralsina:
Use a FS cache for jinja bytecode like we do for Mako.

### DIFF
--- a/nikola/plugins/template/jinja.py
+++ b/nikola/plugins/template/jinja.py
@@ -52,9 +52,14 @@ class JinjaTemplates(TemplateSystem):
         """Initialize Jinja2 environment with extended set of filters."""
         if jinja2 is None:
             return
-        # Not ideal because it's not using our cache structure
-        # but we don't have that folder at init time
-        cache = jinja2.FileSystemBytecodeCache()
+
+    def set_directories(self, directories, cache_folder):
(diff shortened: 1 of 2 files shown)


nikola/plugins/template/jinja.py
```diff
@@ -52,7 +52,10 @@ def __init__(self):
         """Initialize Jinja2 environment with extended set of filters."""
         if jinja2 is None:
             return
-        self.lookup = jinja2.Environment()
+        # Not ideal because it's not using our cache structure
```
Kwpolska: Where is it held then? If this is the case, couldn’t we handle this in `create_lookup` or `set_site`?

ralsina: According to the jinja docs "system temp folder". It looks like it needs to be set on env creation. Maybe we can delay that until set_site?

ralsina: I'll see if I can fix this.

ralsina: Done I think.

Kwpolska:
Merge if this does not break anything (test it yourself)

Issue #2189, opened 2015-12-20
Enclosures seem broken


ralsina:
This is supposed to create an enclosure in the RSS feed, and doesn't:

```
.. title: The R-Podcast Episode 14: Tips and Tricks for using R-Markdown
.. slug: the-r-podcast-episode-14-tips-and-tricks-for-using-r-markdown
.. date: 2015-11-18 14:07:21
.. tags: Podcast
.. description: 
.. wp-status: publish
.. enclosure: https://archive.org/download/trp_audio_014/rpodcast014.mp3

I just did a fresh nikola deploy and the feed still isn't showing this episode's mp3 file.  Here's the default feed in case it helps: http://dev.r-podcast.org/rss.xml .  
```

ralsina:
Fixed by 497d83eb20accc102af673e8955138d3e6a79d2c

ralsina:
Fixed.

PR #2191, opened 2015-12-21
Fix enclosure support in tags plugin

thercast:
I commented on this issue within the [mailing list](https://groups.google.com/d/msg/nikola-discuss/qYTaOoek-ic/SpIDqEHNBQAJ), but this pull request simply imports the enclosure plugin to ensure any RSS feeds created for specific tags/categories also contains enclosures.  See #2189 for background on how this was fixed for the default site RSS feed.

### DIFF
--- a/nikola/plugins/task/tags.py
+++ b/nikola/plugins/task/tags.py
@@ -38,6 +38,7 @@ except ImportError:
 
 from nikola.plugin_categories import Task
 from nikola import utils
+from nikola.nikola import _enclosure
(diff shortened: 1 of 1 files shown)


ralsina:
It probably also means that this same thing will manifest in a different place, like sections, categories, or whatever generates a RSS feed out of a list of posts.

Merging this because it fixes a real problem, but maybe the real solution is to just always use _enclosure in the feed generator.

PR #2193, opened 2015-12-23
Shortcodes

ralsina:
Implement hugo-style shortcodes to help markups reach feature-parity with RST.

### DIFF
--- a/docs/extending.txt
+++ b/docs/extending.txt
@@ -549,3 +549,44 @@ arguments.  Example use:
         endswith=endswith) + t
 
     site.template_hooks['page_header'].append(greeting, True, u'Nikola Tesla', endswith=u'!')
+
+Shortcodes
+==========
+
+Some (hopefully all) markup compilers support shortcodes in these forms::
+
+    {{% foo %}}  # No arguments
+    {{% foo bar %}}  # One argument, containing "bar"
+    {{% foo bar baz=bat %}}  # Two arguments, one containing "bar", one called "baz" containing "bat"
+    
+    {{% foo %}}Some text{{% /foo %}} # one argument called "data" containing "Some text"
+
+So, if you are creating a plugin that generates markup, it may be a good idea to register it as a shortcode
+in adition of a restructured text directive or markdown extension, thus making it available to all markups.
+    
+To implement your own shortcodes from a plugin, you can call ``Nikola.register_shortcode(name, f)`` with the following
+arguments:
+
+name:
+    name of the shortcode (foo in the examples above)
+f:
+    A function that will handle the shortcode
+    
+The shortcode handler should take *at least* the following named arguments:
+
+site:
+    An instance of the Nikola class, to access site state
+    
+data:
+    If the shortcut is used as opening/closing tags, it will be the text between them, otherwise None
+    
+If the shortcode tag has arguments of the form "foo=bar" they will be passed as named arguments. Everything else
+will be passed as positional arguments in the function call.
+
+So, for example::
+
+    {{% foo bar baz=bat beep %}}Some text{{% /foo %}}
+    
+Will cause a call like this::
+
+     foo_handler("bar", "beep", baz="bat", data="Some text", site=whatever)
--- a/docs/extending.txt
+++ b/docs/extending.txt
@@ -562,7 +562,7 @@ Some (hopefully all) markup compilers support shortcodes in these forms::
     {{% foo %}}Some text{{% /foo %}} # one argument called "data" containing "Some text"
 
 So, if you are creating a plugin that generates markup, it may be a good idea to register it as a shortcode
-in adition of a restructured text directive or markdown extension, thus making it available to all markups.
+in addition of a restructured text directive or markdown extension, thus making it available to all markups.
     
 To implement your own shortcodes from a plugin, you can call ``Nikola.register_shortcode(name, f)`` with the following
 arguments:
--- a/nikola/shortcodes.py
+++ b/nikola/shortcodes.py
@@ -33,6 +33,7 @@ except ImportError:
 
 from .utils import LOGGER
 
+
 def apply_shortcodes(data, registry, site=None):
     """Apply Hugo-style shortcodes on data.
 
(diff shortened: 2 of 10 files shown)


nikola/shortcodes.py
```diff
@@ -0,0 +1,121 @@
...
+    [['foo', ([], {}), 0, 11], ['bar', ([], {}), 11, 22]]
+
+    >>> list(_find_shortcodes('{{% foo bar baz=bat fee=fi fo fum %}}'))
+    [['foo', (['bar', 'fo', 'fum'], {'fee': 'fi', 'baz': 'bat'}), 0, 37]]
+
+    >>> list(_find_shortcodes('{{% foo bar bat=baz%}}some data{{% /foo %}}'))
+    [['foo', (['bar'], {'bat': 'baz', 'data': 'some data'}), 0, 43]]
+    """
+
+    # FIXME: this is really space-intolerant
```
Kwpolska: y u no regex

nikola/shortcodes.py
```diff
@@ -0,0 +1,122 @@
...
+        else:  # Tag with closer
+            args[1]['data'] = data[end + 3:close - 1]
+            end = close + len(close_tag) + 1
+        pos = end
+        yield [name, args, start, end]
+
+
+class SCParser(HTMLParser):
+    """Parser for shortcode arguments."""
+    # Because shortcode attributes are HTML-like, we are abusing the HTML parser.
```
Kwpolska: This is overkill IMO. Mind replacing with a really simple Python parser?

ralsina: I would not mind, but after this branch is merged?

docs/extending.txt
```diff
@@ -549,3 +549,44 @@ arguments.  Example use:
...
     site.template_hooks['page_header'].append(greeting, True, u'Nikola Tesla', endswith=u'!')
+
+Shortcodes
+==========
+
+Some (hopefully all) markup compilers support shortcodes in these forms::
+
+    {{% foo %}}  # No arguments
+    {{% foo bar %}}  # One argument, containing "bar"
+    {{% foo bar baz=bat %}}  # Two arguments, one containing "bar", one called "baz" containing "bat"
```
Kwpolska: could use a note about "quotes" being accepted

docs/extending.txt
```diff
@@ -549,3 +549,44 @@ arguments.  Example use:
...
+    
+    {{% foo %}}Some text{{% /foo %}} # one argument called "data" containing "Some text"
+
+So, if you are creating a plugin that generates markup, it may be a good idea to register it as a shortcode
+in adition of a restructured text directive or markdown extension, thus making it available to all markups.
```
Kwpolska: in addition to a reStructuredText directive or Markdown extension, thus making it available to all input formats.

docs/extending.txt
```diff
@@ -549,3 +549,44 @@ arguments.  Example use:
         endswith=endswith) + t
 
     site.template_hooks['page_header'].append(greeting, True, u'Nikola Tesla', endswith=u'!')
+
+Shortcodes
+==========
+
+Some (hopefully all) markup compilers support shortcodes in these forms::
```
Kwpolska: Do we support it in HTML and ipynb?  (ipynb might need some extra help, I’d rather not run this over Python cells/a JSON file)

ralsina: So far, only markdown.

ralsina: About ipynb ... this runs over everything pretty dumbly. In fact, now that you mention it, adding it to the rest compiler would break the docs, now that they mention shortcodes ;-)

Kwpolska: Create a `_raw` shortcode.

Input:

{{% _raw foo bar baz=bat %}}

Output:

{{% foo bar baz=bat %}}

ralsina:
What do you say @Kwpolska ? Merge?

Kwpolska:
I would merge it, but I think that `data` should be a guaranteed argument (to make things easier for devs) — am I thinking right?

ralsina:
Why not? Done in the next revison.

Kwpolska:
Merge if the newest commit passes tests.

Also, here’s a reason to drop HTMLParser:

    {{% test "hello world" %}}
    → ('"hello', 'world"')

(I’ll send a PR for my parser after this gets merged)

ralsina:
Well, AFAIK that's not valid "html-like", is it? But sure, looking forward to the parser :-)

docs/extending.txt
ralsina: And I can add that one when I add a bunch of builtin template-based shortcodes in the next branch. GOOD.

felixfontein:
Out of curiosity: how is `{{% test "%}}" %}}` handled?

Kwpolska:
```py
nikola.shortcodes.apply_shortcodes('hello {{% test "%}}" %}} world', {'test': lambda a, **kw: ':' + a + ':'}, None)
'hello ::" %}} world'

**Comment by felixfontein** [comment]
Hmm, that's not what I would expect as an end-user (with programming experience). Is this really intended? If it is, shouldn't the parser complain about the not well-formed statement?

**Comment by Kwpolska** [comment]
Please write a better parser then.

**Comment by felixfontein** [comment]
It has to be a combined parser for the regex part and the argument splitting, then. I'll work on that later.

**Comment by felixfontein** [comment]
There are some more problems with the existing parsers:
```
from nikola import shortcodes

#print(shortcodes.apply_shortcodes('==> {{% <==', {}))
#print(shortcodes.apply_shortcodes('==> {{% a %%} <==', {}))
#print(shortcodes.apply_shortcodes('==> {{% a %}} {{% /a %%} <==', {}))
#print(shortcodes.apply_shortcodes('==> {{% test %}} {{% bla %}} {{% /bla %}} {{% /test %%} <==', {}))
#print(shortcodes.apply_shortcodes('==> {{% test %}} {{% bla %}} {{% /bla %}} {{% /bla %%} <==', {}))
```
Uncomment any of the lines to get an infinite loop.

felixfontein:
See #2200 for a new parser.

PR #2194, opened 2015-12-24
Use a custom shortcode parser

Kwpolska:
This replaces the cheat involving the HTML parser with a custom character-based parser, written in pure Python.

Advantages:
* Support for positional arguments with spaces through quotes and backslashes
* Does not use a hack that works by luck

Disadvantages:
* the code is a bit ugly

### DIFF
--- a/nikola/shortcodes.py
+++ b/nikola/shortcodes.py
@@ -125,7 +125,6 @@ def parse_sc(data):
     cvalue = ''
     qc = ''
     for char in elements[1]:
-        print(char, flag)
         if flag & 0b100 and flag & 1:
             # Backslash in value: escape next character, no matter what
             cvalue += char
--- a/nikola/shortcodes.py
+++ b/nikola/shortcodes.py
@@ -26,11 +26,6 @@
 
 """Support for Hugo-style shortcodes."""
 
-try:
-    from html.parser import HTMLParser
-except ImportError:
-    from HTMLParser import HTMLParser
-
 from .utils import LOGGER
 
 
@@ -84,7 +79,6 @@ def _find_shortcodes(data):
     """
     # FIXME: this is really space-intolerant
 
-    parser = SCParser()
     pos = 0
     while True:
         start = data.find('{{%', pos)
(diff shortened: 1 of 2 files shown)


nikola/shortcodes.py
```diff
@@ -106,28 +100,84 @@ def _find_shortcodes(data):
...
+            cname += char
+            flag -= 0b100
+        elif char == '=' and flag == 0:
+            # Equals sign inside unquoted name: switch to value
+            flag = 1
+        elif char == ' ' and flag == 0:
+            # Space inside unquoted name: save as positional argument
+            args.append(cname)
+            cname = cvalue = qc = ''
+        elif char == ' ' and flag == 1:
```
felixfontein: If you use binary notation above, why not here (and in the following cases) as well?

ralsina:
LGTM

nikola/shortcodes.py
```diff
@@ -106,28 +100,84 @@ def _find_shortcodes(data):
...
+    # 7 value inside quotes and backslash
+    flag = 0
+    cname = ''
+    cvalue = ''
+    qc = ''
+    for char in elements[1]:
+        if flag & 0b100 and flag & 1:
+            # Backslash in value: escape next character, no matter what
+            cvalue += char
+            flag -= 0b100
```
felixfontein: Why do you use `+` and `-` for bit operations instead of `&`, `|` and `~`?

nikola/shortcodes.py
Kwpolska: I’ll drop the binary notation then; it felt more readable to me in this case.

nikola/shortcodes.py
Kwpolska: Because I was too lazy to come up with the binary operations, and because ± is good enough for this case. Fixed on master.

Kwpolska: Because I was too lazy to come up with the bitwise operations needed, and because ± is good enough for this case. Fixed on master.

nikola/shortcodes.py
felixfontein: Why not use binary notation everywhere? That makes it more readable for everyone!

Kwpolska: A bit too much cruft for me. If you really want that, you can make everything binary.

PR #2195, opened 2015-12-24
Builtin shortcodes

ralsina:
* Adds nikola/data/shortcodes/templateenginename for "builtin" shortcodes
* Loads those
* Adds a ``raw`` shortcode
* Use jinjify to translate shortcodes


nikola/nikola.py
```diff
@@ -1294,19 +1294,23 @@ def url_replacer(self, src, dst, lang=None, url_type=None):
...
+        for sc_dir in sc_dirs:
+            if not os.path.isdir(sc_dir):
+                continue
+            for fname in os.listdir(sc_dir):
+                name, ext = os.path.splitext(fname)
+                if ext == '.tmpl':
+                    with open(os.path.join(sc_dir, fname)) as fd:
+                        template_data = fd.read()
+
+                    def render_shortcode(t_data=template_data, **kw):
```
Kwpolska: Could this be better with `functools.partial`?

nikola/data/shortcodes/mako/raw.tmpl
```diff
@@ -0,0 +1 @@
+${data}
```
Kwpolska: jinjify this

Kwpolska:
:+1:

PR #2199, opened 2015-12-27
Fix #2186 -- commit to source branch automatically

### LINKED ISSUE #2186: github_deploy should commit to the source branch automatically
I am using the xxx.github.io repo on github, so i set the GITHUB_SOURCE_BRANCH ='source', and GITHUB_DEPLOY_BRANCH='master". and the OUTPUT_FOLDER is default.
so, basically the source branch has folder structure like this:

source
├── README.txt
├── __pycache__
├── cache
├── conf.py
├── files
├── galleries
├── images
├── listings
├── output
├── plugins
├── posts
├── stories
└── templates

and then i run `nikola github_deploy`, i have generated a mater branch which contains the output results, it is folder structure is like below:

master
├── 2015
├── __pycache__
├── archive.html
├── assets
├── authors
├── cache
├── categories
├── favicon.ico
├── galleries
├── images
├── index.html
├── listings
├── output
├── posts
├── robots.txt
├── rss.xml
├── sitemap.xml
├── sitemapindex.xml
├── stories
└── zh_cn

everything works fine, but i have another output folder in this master branch, which contains exactly the same content as its parent folder... am i doing something wrong? or it suppose to be like this?
if so, i think it is more reasonable to just remove this output folder (and the cache folders as well).

ps: everything on github.com is fine, there is no output folder or cache folders in the master branch, so locally if you run `git status` you have:

      Untracked files:
      (use "git add <file>..." to include in what will be committed)
	.doit.db.db
	__pycache__/
	cache/
	output/
	posts/test.rst

Kwpolska:
This is #2186. It will be the default for new sites and will significantly simplify github_deploy setup and usage.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,7 +4,12 @@ New in master
 Features
 --------
 
+* Commit to source branch automatically in ``github_deploy``
+  if ``GITHUB_COMMIT_SOURCE`` is set to True (Issue #2186)
 * Hugo-like shortcodes (Issue #1707)
+
+Bugfixes
+--------
 
 New in v7.7.4
 =============
@@ -1690,3 +1695,5 @@ Bugfixes
 * Colorbox support in restructured text figures
 * Fix for content displaying too wide
 * Changelog
+
+vim: tw=70 syntax=rst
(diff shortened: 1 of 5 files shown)


Kwpolska:
cc @shelper (original request)

PR #2200, opened 2015-12-28
Completely rewrote shortcode parser.

felixfontein:
This PR fixes some shortcomings in the current shortcode parsers:

  * `{{% a "%}}" %}}` is parsed as expected;
  * better error messages are produced for various malformed situations;
  * no more infinite loops in cases such as `==> {{% <==`;
  * strange constructs such as `{{% a "b"c"d" %}}` aren't allowed anymore.

### DIFF
--- a/nikola/shortcodes.py
+++ b/nikola/shortcodes.py
@@ -244,19 +244,24 @@ def _split_shortcodes(data):
             if pos > len(data) or data[end_start:pos] != '%}}':
                 raise ParsingError("Syntax error: '{{{{% /{0}' must be followed by ' %}}}}' ({1})!".format(name, _format_position(data, end_start)))
             result.append((_SHORTCODE_END, data[start:pos], start, name))
+        elif name == '%}}':
+            raise ParsingError("Syntax error: '{{{{%' must be followed by shortcode name ({0})!".format(_format_position(data, start)))
         else:
             # This is an opening shortcode
-            pos, args = _parse_shortcode_args(data, name_end, name=name, start_pos=start)
+            pos, args = _parse_shortcode_args(data, name_end, shortcode_name=name, start_pos=start)
             result.append((_SHORTCODE_START, data[start:pos], start, name, args))
     return result
 
 
-def apply_shortcodes(data, registry, site=None, filename=None):
+def apply_shortcodes(data, registry, site=None, filename=None, raise_exceptions=False):
     """Apply Hugo-style shortcodes on data.
 
     {{% name parameters %}} will end up calling the registered "name" function with the given parameters.
     {{% name parameters %}} something {{% /name %}} will call name with the parameters and
     one extra "data" parameter containing " something ".
+
+    If raise_exceptions is set to True, instead of printing error messages and terminating, errors are
+    passed on as exceptions to the caller.
 
     The site parameter is passed with the same name to the shortcodes so they can access Nikola state.
 
@@ -309,6 +314,9 @@ def apply_shortcodes(data, registry, site=None, filename=None):
                 result.append(res)
         return empty_string.join(result)
     except ParsingError as e:
+        if raise_exceptions:
+            # Throw up
+            raise e
         if filename:
             LOGGER.error("Shortcode error in file {0}: {1}".format(filename, e))
         else:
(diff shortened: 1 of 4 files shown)


nikola/shortcodes.py
```diff
@@ -27,6 +27,221 @@
...
+
+
+class ParsingError(Exception):
+    """Used for forwarding paring error messages to apply_shortcodes."""
+
+    pass
+
+
+def _format_position(data, pos):
+    """Return position formatted as line/column."""
```
Kwpolska: Could use a note that this is for pretty error messages.

nikola/shortcodes.py
```diff
@@ -43,141 +259,49 @@ def apply_shortcodes(data, registry, site=None):
...
-            cname += char
-
-    # Handle last argument
-    if cvalue:
-        kwargs[cname] = cvalue
-    else:
-        args.append(cname)
-
-    return name, (args, kwargs)
+    empty_string = data[:0]  # same string type as data; to make Python 2 happy
```
Kwpolska: I think `''` will be enough in our case. By the way, this file didn’t have a `from __future__ import unicode_literals` import — I just fixed that.

Kwpolska:
* What does *non-trivial* mean?
* How fast is it for large inputs?

nikola/shortcodes.py
felixfontein: This makes some test fail, since it always returns a unicode string. My final fix makes it return the same string type as the input was. But we could of course also fix the tests :)

felixfontein:
You can also replace `non-trival` with `non-empty` if you want.

About large inputs: since it uses `str.find('{{%', pos)` to find the start of the next shortcode, and only then starts going through the string character by character until the end of this shortcode, it should be pretty efficient. If you use no shortcodes, it should not be noticeably slower than the old code.

Kwpolska:
Would you mind adding tests (real tests, not doctests!) for all the things this parser fixes?

felixfontein:
I found some more bugs, improved some error messages, and added proper tests.

Kwpolska:
LGTM, anybody else?

ralsina:
LGTM too. Merging.

ralsina:
@felixfontein how could we expand on this to make it handle both hugo shortcode cases?

1) {{% %}} contents of the shortcode are passed as-is
2) {{< >}} contents of the shortcode are passed through the markup compiler before replacing

Kwpolska:
Let’s not do `< >`. It’s tricky, because some things in the foodchain might replace it with `&lt;` and `&gt;`. We don’t need full feature parity with Hugo.

ralsina:
We could use a different marker.

Kwpolska:
Is % really not enough?

On 4 January 2016 at 20:23, Roberto Alsina <notifications@github.com> wrote:

> We could use a different marker.
>
> On Mon, Jan 4, 2016 at 12:46 PM Chris Warrick <notifications@github.com>
> wrote:
>
> > Let’s not do < >. It’s tricky, because some things in the foodchain might
> > replace it with &lt; and &gt;. We don’t need full feature parity with
> > Hugo.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/getnikola/nikola/pull/2200#issuecomment-168711749>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/getnikola/nikola/pull/2200#issuecomment-168776713>.
>



-- 
PGP: 5EAAEA16

felixfontein:
We might already have some problems, if you write something along `{{% a "b<c" %}}` it might be that the argument will be `b&lt;c` instead of `b<c` since shortcodes are applied to the transformed result.

It would be better to first replace the shortcodes with some unique IDs, then use markdown, and finally replace the markers by their transformed content. This would also make it trivial to implement support for `{{< >}}`: just apply these shortcodes before applying `markdown`.

ralsina:
The reason for  "shortcodes with markup inside them" is pretty good. Imagine you want to make a paragraph be a bootstrap warning panel, or whatever. You could do this:

````
{{< warning-panel >}}
Some markdown or rest here
{{< /warning-panel >}}
````

Requiring the content of the shortcode not be processed by the compiler makes this usecase suck.

PR #2201, opened 2015-12-29
Sorting with priority

Determinant:
As discussed with ChrisWarrick, nikola currently implements sorting in ``nikola/nikola.py`` instead of the separated ``scan_posts`` plugin. Since the "sticky posts" is a very useful and desirable feature (at least for CMS and blog sites), I guess the most direct and simple way is to extend the original sorting key: ``(p.date, p.source_path)`` with an additional sorting key which can override the order. The problem is whether to directly inject the code into ``nikola.py`` or try to move the sorting functionality to an individual plugin or to ``scan_posts`` plugin. This PR demonstrates a possible way to incoporate into ``nikola.py`` and it works without breaking the original behavior. Users can optionally use ``.. priority: 1`` to make a post sticky (which means the default is 0). They can also use other number to give a finer order control.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,4 +2,7 @@
 universal = 1
 
 [flake8]
-ignore=E501
+ignore = E501
+
+[easy_install]
+
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,7 +2,4 @@
 universal = 1
 
 [flake8]
-ignore = E501
-
-[easy_install]
-
+ignore=E501
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,7 +4,14 @@ New in master
 Features
 --------
 
+* Allow for customizing ``github_deploy`` commit messages with ``-m``
+  (Issue #2198)
+* Commit to source branch automatically in ``github_deploy``
+  if ``GITHUB_COMMIT_SOURCE`` is set to True (Issue #2186)
 * Hugo-like shortcodes (Issue #1707)
+
+Bugfixes
+--------
 
 New in v7.7.4
 =============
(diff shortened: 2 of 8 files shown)


setup.cfg
```diff
@@ -2,4 +2,7 @@
 universal = 1
 
 [flake8]
-ignore=E501
+ignore = E501
+
+[easy_install]
```
Kwpolska: delete that

Kwpolska:
:+1: from me

Correct indentation:

```py
            thing.sort(key=lambda p:
                       (int(p.meta('priority')) if p.meta('priority') else 0,
                        p.date, p.source_path))
```

felixfontein:
There are some more places where posts are sorted, for example for tag pages, archives, ... Should these also respect priority? That would require some more changes.

How about having the `Post` object return a sort key? Then all places where posts are sorted could simply use that sorting key, and future changes to the sorting order can be done in one place (that function of `Post`).

Kwpolska:
Is sorting things multiple times a good thing to do in the first place?

felixfontein:
Of course, you could also rewrite a lot of code to ensure that resorting isn't needed.

Anyway, even if that is done, putting the sort key into a function of `Post` makes it cleaner in my opinion. Especially for plugin writers who want to do something fancy with sets of posts and simply want to sort the result, instead of writing a bunch of code which compares the set with the original timeline and reconstructs the correct order from there.

Determinant:
Modified as suggested by @Kwpolska. @felixfontein, I agree with your idea. I'd definitively like to see the behavior of post sorting can be somehow customized. How do you plan to expose this functionality to users? Also via plugins? Since this commit is just a small change, you and @Kwpolska could adopt it as a temporary solution (with some TODO comments, if needed), but of course I'd like to see or implement a more elegant way of doing this. :)

ralsina:
So, I don't know :-)

@Kwpolska  if you +1 this, let's merge it. I prefer not to have PR rot (and we have some already)

PR #2207, opened 2016-01-04
WIP Copy EXIF data to resized images (fix #2204)

### LINKED ISSUE #2204: EXIF not copied when re-scaling
If you have images for a gallery with pictures greater than `MAX_IMAGE_SIZE`, when they are re-scaled they loose the EXIF information. So, if you are using `GALLERY_SORT_BY_DATE = True` you will loose the order when they are displayed.

To solve this issue, I use `convert` (from ImageMagick to resize the image). After converting all of them, I revered the sorting order.

ralsina:
This branch does a naïve copy of EXIF data from the source images to the resized targets. This means it preserves things like EXIF dates, but it **also** preserves size in pixels, which is now wrong.

This is hidden behind the PRESERVE_EXIF_DATA option, so set that if you want to test it.


Kwpolska:
We should only select date information IMO. Many people wouldn’t want to share all EXIF tags.

ralsina:
We probably need to offer to scrub EXIF and copy the right bits if the user doesn't request it.

EXIF scrubbing is a good privacy measure.

ralsina:
Looks like I *could* just remove the ImageWidth / ImageLength tags if they are present, and wrap this in a "PRESERVE_EXIF_DATA" option that defaults to False.

humitos:
I agree with @Kwpolska about only copy dates, at least for the first version.

In the future version, it would be great to have a config where you can write down all the EXIF fields you want to be shared in your gallery. If the person choose widht and size, we should put the right one in the resized image.

Regarding new dependency. I'm not sure, but I think it easier to maintain a new dependency that a bunch of code from "somewhere". In fact, this dependency could be optional as many other that nikola already has.

ralsina:
@humitos can't be done. All or nothing. "Some" requires an extra dependency.

ralsina:
This code seems to more or less do what the user would expect. My initial concern about preserving EXIF data with wrong image size seems unfounded. While theoretically possible, no images I have seen actually HAVE that data, and even if they had it, viewers happily ignore it and use the real image sizes from the image format native metadata.

ralsina:
WTH let's merge this, if it breaks anything let me know.

PR #2210, opened 2016-01-04
simplistic implementation of data caching and persistance

ralsina:
This implementation of data caching/persistence is intentionally simplistic. The parameters for the design were:

1) It should provide reasonable diffs when you add/remove data
2) Data should be typed
3) Should support basic data types

So, what do you think?

Known problems:

* ~~To make this atomic, it needs slightly more complicated saving code. Since doit can run threaded, that's important.~~

* ~~In order not to miss saved data when using multiprocess mode, it may need to reload the file on each read / some reads, which makes it much more expensive.~~


ralsina:
WTH let's merge this.

PR #2211, opened 2016-01-06
Insert space

okin:
This should improve readability because now the date and the title are separated by a space.


Kwpolska:
Thanks! (since you have commit rights, you could’ve just commited to `master` without even asking anyone).

PR #2212, opened 2016-01-07
Fix tag template (base and base-jinja)

angelgcuartero:
Somehow this got broken in the past and now it doesn’t have the right
field order. It has to be first the post title and second, the date.

### DIFF
--- a/nikola/data/themes/base-jinja/templates/tag.tmpl
+++ b/nikola/data/themes/base-jinja/templates/tag.tmpl
@@ -43,7 +43,7 @@
     {% if posts %}
     <ul class="postlist">
     {% for post in posts %}
-        <li><time class="listdate" datetime="{{ post.formatted_date('webiso') }}" title="{{ post.formatted_date(date_format)|e }}">{{ post.formatted_date(date_format)|e }}</time><a href="{{ post.permalink() }}" class="listtitle">{{ post.title()|e }}<a></li>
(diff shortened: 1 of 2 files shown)


Aeyoun:
Look in the git history. Someone changed this on purpose. You can change the presentation order using CSS position.

Also, Python’s isoformat() and Nikola’s 'webiso' doesn’t produce the same date format. 'webiso' is a subset of ISO8601 for the web and this subset is required for a valid HTML5 datetime.

angelgcuartero:
Sorry if I made a mistake pushing changes to the repo… I was trying to do some branching and commit locally and Github Desktop seemed like it was not refreshing info correctly. I might have send a PR when I was doing tests (not a git expert).

Anyway, I was testing a standard installation and it doesn't show the tag post list right. Either the template is not right or the code to inject the data into the template is not doing it as it should be, I think. I will look for more information.

Kwpolska:
@Aeyoun This change would be fine, if it was implemented properly. As of right now, the **default** themes look like this:

* — 2015-03-30 23:00[Welcome to Nikola](#)

Which is just plain awful. I’ll fix this by updating the CSS and other files that use `.listdate` and `.listtitle`… In the future, please check how things look in the built-in themes.

Kwpolska:
Superseded by #2213. Thanks for reporting the bug!

PR #2216, opened 2016-01-17
Allowing name of tidy5 executable to be specified.

felixfontein:
When compiling tidy-html 5 from source, the binary is called `tidy` on my system and not `tidy5`. Since I don't want to create symlinks/hardlinks just because Nikola assumes the command to be `tidy5` (and in fact, on some systems you might not be able to do that because you aren't root), here's a patch which allows to specify the executable name when creating a tidy5-based filter.

### DIFF
--- a/nikola/filters.py
+++ b/nikola/filters.py
@@ -142,54 +142,54 @@ def yui_compressor(infile):
             raise Exception("yui-compressor is not installed.")
             return False
 
-    return runinplace(r'{} --nomunge %1 -o %2'.format(yuicompressor), infile)
+    return runinplace('{} --nomunge %1 -o %2'.format(yuicompressor), infile)
 
 
 def closure_compiler(infile):
     """Run closure-compiler on a file."""
-    return runinplace(r'closure-compiler --warning_level QUIET --js %1 --js_output_file %2', infile)
+    return runinplace('closure-compiler --warning_level QUIET --js %1 --js_output_file %2', infile)
 
 
 def optipng(infile):
     """Run optipng on a file."""
-    return runinplace(r"optipng -preserve -o2 -quiet %1", infile)
+    return runinplace("optipng -preserve -o2 -quiet %1", infile)
 
 
 def jpegoptim(infile):
     """Run jpegoptim on a file."""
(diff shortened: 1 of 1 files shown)


nikola/filters.py
```diff
@@ -160,36 +160,36 @@ def jpegoptim(infile):
     return runinplace(r"jpegoptim -p --strip-all -q %1", infile)
 
 
-def html_tidy_withconfig(infile):
+def html_tidy_withconfig(infile, executable=r'tidy5'):
```
Kwpolska: These, as well as the commands, do not need `r'strings`'. Regular `'strings'` are enough.

Also, r'strings' can be concatenated with regular 'strings'. The r'string' is an information for the parser only, and it does not do anything if there are no backslashes in the string literal.

felixfontein: So should I remove the `r` prefix for all strings in all tidy5 related functions?

Kwpolska: yes

felixfontein:
Regarding the failing invariance tests: I assume that this is related to some other change?

nikola/filters.py
felixfontein: Fixed in f2c9e22fd87af83e8cf3424c2b6b382494bdb2ee.

Kwpolska:
That invariance test was caused by a new Pygments release (last one was a year ago, BTW). I’ll rebuild invariance and merge this.

felixfontein:
Thanks!

Aeyoun:
Note that this was for the tidy5 utility and not tidy-html. Don’t use the latter, it doesn’t spec modern web.

Aeyoun:
Note that this filter was added for the tidy5 utility and not tidy-html. Don’t use the latter, it doesn’t spec modern web and will make a mess of modern markup. I’d suggest we revert this so people don’t use tidy-html.

https://github.com/htacg/tidy-html5

Kwpolska:
In order to use that, you need to **know** this is an option, and use `functools.partial` for the filter.

felixfontein:
@Aeyoun: It is still called "HTML Tidy" (with HTML support), according to it's homepage: http://www.html-tidy.org/

Issue #2220, opened 2016-01-18
Random rebuilds of files


felixfontein:
In the last couple of days (might have been for longer), I experienced that some files are seemingly randomly rebuilt when running `nikola run`. I've tracked this down to `doit` config changes. For example, in one call I get:
```
"_link": "<bound method Nikola.link of <Nikola Site: 'A'>>"
```
and in a later call
```
"_link": "<bound method Nikola.link of <Nikola Site: 'B'>>"
```
Here, `A` and `B` are the two titles of my blog (it is bilingual). Whether the title used here is `A` or `B` seems to change from time to time per task, which causes rebuilds if it changes. Similar names (`"<bound method ... of <Nikola Site: 'xxx'>>"`) are generated also for `abs_link`, `exists` and some more.

This probably can be fixed by changing `__repr__` of the `Nikola` class to:
```
    def __repr__(self):
        """Representation of a Nikola site."""
        return '<Nikola Site: {0!r}>'.format(self.config['BLOG_TITLE'](self.config['DEFAULT_LANG']))
```
so that always the default language is used, and not `LocaleBorg`'s current language.

What do you think?

felixfontein:
Changing this locally fixed it somewhat for me, but not totally. While investigating further, I noticed that there's another entry which changes: sometimes, the config contains `"author_pages_generated": false`, while sometimes it does not. That probably depends on the order the plugins were loaded and processed in, as the `plugins/task/authors.py` plugin sets this flag in the `site.GLOBAL_CONTEXT`.

I think the `authors` plugin must be changed since this is clearly a bug. What do you think?

At least, disabling this plugin seemed to have fixed the problem for me, or at least reduced its frequency considerably.

(Also, note that the different orders in which plugins are processed probably occur mostly with Python 3, and not with Python 2.)

ralsina:
Or maybe just load the plugins in a deterministic order.

ralsina:
Your change about the repr LGTM care to propose it?

felixfontein:
That doesn't really fix the problem. I mean, what's the sense of having `author_pages_generated` in `GLOBAL_CONTEXT` for only a subset of all generated pages? Either it should be in all, or in none.

felixfontein:
A fix for the `authors` plugin could be done by adding a hook to be run after `scan_posts` and setting the variable there. That would probably be a much cleaner solution.

ralsina:
I don't see why authors needs to put that in config.... I'll take a look at that part.

[...]

PR #2222, opened 2016-01-19
Make authors plugin use a signal to avoid race conditions

ralsina:
Like @felixfontein suggested, a signal here so things are more consistent and we avoid part of #2220

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,7 @@ Features
 Bugfixes
 --------
 
+* Avoid some random file rebuilds (Issue #2220)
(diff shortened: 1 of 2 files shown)


felixfontein:
LGTM (assuming it works, I've currently disabled that plugin for my blogs ;-) )

ralsina:
Well, it passes tests, so it doesn' t CRASH ... have not tested it beyond that. I'll test it a bit later and then merge if it works.

ralsina:
Ok, merged the authors fix. Still think a more deterministic load order for plugins will make things nicer in the future, but it' s not mandatory.

PR #2225, opened 2016-01-22
fixed mistranslated korean snipplet

icq4ever:
fixed mistranslated korean snipplet


ralsina:
We are doing all the translations in transifex, changing it here doesn't work. Can you apply at https://www.transifex.com/ralsina/nikola/ ? Thanks!

icq4ever:
OK! thanks :)

PR #2241, opened 2016-02-11
Make jinjify support converting single template files.

### LINKED ISSUE #2240: jinjify: convert a single mako template file to jinja?
It looks like jinjify requires a full theme to work. For example, how do you synchronize the two templates in the [projectpages plugin](https://github.com/getnikola/plugins/tree/master/v7/projectpages/templates) with jinjify?

xuhdev:
Close #2240

### DIFF
--- a/scripts/jinjify.py
+++ b/scripts/jinjify.py
@@ -6,6 +6,7 @@ import os
 import re
 import json
 import shutil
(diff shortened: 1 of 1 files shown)


Kwpolska:
Thanks!

PR #2245, opened 2016-02-14
POC of handling filter failures

ralsina:
Don't merge this ;-)

It's a little POC of how we can handle filter failures in a more useful way, specifically by making the task that uses the filter fail if the filter fails.

I am not even sure this is a good idea.


Kwpolska:
Maybe we could use calculated-dependencies to add a `False` to `uptodate` if a filter fails?

nikola/filters.py
```diff
@@ -287,16 +290,16 @@ def php_template_injection(data):
 def cssminify(data):
     """Minify CSS using http://cssminifier.com/."""
     try:
-        url = 'http://cssminifier.com/raw'
+        url = 'http://cssminifier.comFOO/raw'
```
felixfontein: This should probably have not been in the commit?

ralsina:
Not doing it.

Issue #2170, opened 2015-11-11
"reStructuredText Extensions" in Markdown


socheon:
Hi. I have just started experimenting with Nikola.

It will be great if there is a documentation, gathering all the extensions available for Markdown (Pandoc preferably).

Instead of implementing the extensions one by one for Markdown, will it be easy to support the `.. something::` syntax in Markdown, e.g.

```
.. media::
.. vimeo::
.. youtube::
```

Perhaps by running it through some pre or post processor.

Kwpolska:
1. Markdown via Pandoc is not recommended and not officially supported, and if you do use it, please consult [Pandoc documentation](http://pandoc.org/README.html#pandocs-markdown) yourself.  That said, I don’t think you can enable the Pandoc extensions in Nikola…
2. This is sort-of doable. We could just split out the business logic and write some extension factory. We just need a regexp expert, someone who understands this mess and can generalize it:

    ```py
GIST_RST_RE = r'(?m)^\.\.\s*gist::\s*(?P<gist_id>[^\]\s]+)(?:\s*(?P<filename>.+?))?\s*$'

socheon:
Maybe this works. We can process the args and properties in code.

```
pattern = r'(?ms)^\.\.\s*(?P<command>[\w-]+)::\s*(?P<args>[^\n]+)(?P<props>.*?)\n(?!\s)'
input = """

.. media:: http://vimeo.com/72425090

.. vimeo:: 20241459
   :height: 240
   :width: 320

.. code-block:: python
   :number-lines:

   x = 1
   y = 2

.. listing:: foo.py python
"""

for m in re.finditer(pattern, input):
    if m:
        print m.groupdict()
        print
```

Output
```
{'args': 'http://vimeo.com/72425090', 'command': 'media', 'props': '\n'}

{'args': '20241459', 'command': 'vimeo', 'props': '\n   :height: 240\n   :width: 320\n'}

{'args': 'python', 'command': 'code-block', 'props': '\n   :number-lines:\n\n   x = 1\n   y = 2\n'}

{'args': 'foo.py python', 'command': 'listing', 'props': '\n'}
```

socheon:
Not sure how robust is using regexp.
Maybe is better to modify the parser. A rst directive looks a lot like a code block.

Kwpolska:
I’d rather not use a single regex to capture everything, sounds too broad and complicated…

socheon:
[...]

PR #2247, opened 2016-02-14
Implement rst directives as shortcode for markdown (Issue #2170)

ralsina:
WIP because there are a bunch of these and I will do most of them, slowly.


ralsina:
While this does not expose ALL the rst extensions as shortcodes it does expose a couple, and we can iterate later to expose more, no point in keeping it hidden.

PR #2248, opened 2016-02-14
Fix ipynb to actually render the output

bspeice:
Should just be a quick fix - the output isn't being written, so IPython/Jupyter support is totally broken at the moment.


ralsina:
Ugh. This sure needs a bugfix release :-P

ralsina:
By the way, I just noticed I did not do this before: thanks for the fix @bspeice ! It's much appreciated.

PR #2256, opened 2016-02-19
Update path_handlers.txt

samorajp:
Typo.


ralsina:
Good catch, thanks!

Kwpolska:
That file is auto-generated. I fixed the original `nikola.py` typo, too.

Issue #2273, opened 2016-02-28
Replace umlauts with correct vokals


davidak:
When you create a new post, umlauts in the slug don't get replaced correctly.

is: `ü -> u`

should be:
```
ä -> ae
ö -> oe
ü -> ue
```

Kwpolska:
We can’t do that. Only German uses this rule; other languages do not allow such substitutions (opting for `o` instead).

felixfontein:
We already had this discussion in #2172. I'd really like to see this fixed in a good way, like I suggested in that PR: pass the language to `slugify` and let `slugify` allow to extend `slugify` similar to the month name/date formatting handlers for `LocaleBorg`.

Kwpolska:
Again: we depend on the slugify output being invariant. And considering how hard it is to get locale support right sometimes (race conditions and such), we can’t do that sensibly. We would need a lot of test coverage to make sure there are no bugs caused by trusting `slugify()` to be invariant and forgetting about a minor change for one language.

That said: you can always disarm slugify. You will get something even better: spaces, capital letters, and äöüß in your addresses.

Or even better: you could have some post-processing thing (a **signal handler**, perhaps?) that would fix your post slugs by editing it in the input files. You could even make it a plugin and put it on plugins.getnikola.com. This sounds like the best resolution for everyone: the German-speaking Nikola community gets correct slugs in the most prominent places and we don’t have to introduce even more complexity to LocaleBorg.

felixfontein:
1. `slugify` is usually called in places where a link is generated, and there the language is always known (otherwise, how should the link be generated? or the link will be wrong anyway).

   Or is `slugify` used in another context than link / URL generation?

2. Of course you can disarm `slugify` completely, but that's usualy not what one wants. I want to restrict to ASCII letters in URLs, but I don't want screwed up umlaut transformation, either.

3. A signal handler sounds like a really, really hacky and even worse solution.

4. For languages which don't use the Latin alphabet, you might want completely different transformations and not just a fallback to UTF-8.

Kwpolska:
Slugify is also used for “tags are too similar” feature — it might lead to some weird bugs here or there. And if we were to change the behavior now, for everyone, some URLs would be changed, maybe even without the user noticing, and might require orphan cleanup and adding custom redirects. If this was an option, it would be unpopular and untested.

[...]

PR #2284, opened 2016-03-06
Pass language to slugify and unslugify. (WIP)

felixfontein:
First try on implementing passing of language to `slugify` and `unslugify`; see discussion in #2273.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1737,7 +1737,7 @@ class Nikola(object):
                         if _tag_slugified in slugged_tags[lang]:
                             if tag not in self.posts_per_tag:
                                 # Tags that differ only in case
-                                other_tag = [existing for existing in self.posts_per_tag.keys() if utils.slugify(existing) == _tag_slugified][0]
+                                other_tag = [existing for existing in self.posts_per_tag.keys() if utils.slugify(existing, lang) == _tag_slugified][0]
                                 utils.LOGGER.error('You have tags that are too similar: {0} and {1}'.format(tag, other_tag))
                                 utils.LOGGER.error('Tag {0} is used in: {1}'.format(tag, post.source_path))
                                 utils.LOGGER.error('Tag {0} is used in: {1}'.format(other_tag, ', '.join([p.source_path for p in self.posts_per_tag[other_tag]])))
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1722,7 +1722,7 @@ class Nikola(object):
 
         quit = False
         # Classify posts per year/tag/month/whatever
-        slugged_tags = set([])
+        slugged_tags = defaultdict(set)
         for post in self.timeline:
             if post.use_in_feeds:
                 self.posts.append(post)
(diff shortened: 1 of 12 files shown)


nikola/nikola.py
```diff
@@ -1730,19 +1730,20 @@ def scan_posts(self, really=False, ignore_quit=False, quiet=False):
...
-                    else:
-                        slugged_tags.add(utils.slugify(tag))
                     self.posts_per_tag[tag].append(post)
                 for lang in self.config['TRANSLATIONS'].keys():
+                    for tag in post.tags_for_language(lang):
+                        _tag_slugified = utils.slugify(tag, lang)
+                        if _tag_slugified in slugged_tags[lang]:
+                            if tag not in self.posts_per_tag:
+                                # Tags that differ only in case
+                                other_tag = [existing for existing in self.posts_per_tag.keys() if utils.slugify(existing) == _tag_slugified][0]
```
Kwpolska: missing `, lang` here.

ralsina:
Since slugify could be used by plugins, unless you want this to have to
wait until v8, you should give lang a default value.

Kwpolska:
I’m :-1: for this. You just wasted your time implementing a tiny obscure feature.

felixfontein:
@ralsina: I added a warning in that case. This warning will be useful while this is WIP.

@Kwpolska: I really want to have this feature for my blog, for me it's neither tiny nor obscure. And at least @davidak and @akrist seem to agree since they bothered to open an issue resp. write code for this.

davidak:
It is important for a german site since it is the grammatically correct representation.

Wordpress supports it since Version 3.6:
http://www.perun.net/2013/04/17/wordpress-3-6-umlaute-im-titel-ohne-plugin-entschaerfen/

before, most people used a plugin.

To implement this as a plugin would be ok for me.

there is a python module that does what we want:
https://pypi.python.org/pypi/translitcodec/

```
>>> import translitcodec
>>> import codecs
>>> codecs.encode(u'Schöne Grüße aus Köln', 'translit/long')
'Schoene Gruesse aus Koeln'
```

felixfontein:
@davidak: this Python module cannot be used for all languages since for most it would be simply wrong. Also, I wonder how well it works for languages such as Chinese and Japanese.

My goal is to not change Nikola's default behavior, but to just enable this to be changed by plugins without writing a really insane plugin (which does post preprocessing etc.). Such a plugin could use this Python module for the German language, though.

About WordPress: I wonder why they only support de_DE and not also de_CH and de_AT. Also, the current version of WordPress uses similar translations for Danish (da_DK).

davidak:
>Such a plugin could use this Python module for the German language, though.

that was my intention.

ralsina:
This looks ok to me, and changes no behaviour. If noone sees a bug in the next 24 hours, I'll merge.

felixfontein:
@ralsina: maybe we should leave it open a bit longer, I'd like to do some more intense tests (like with a real blog :) ). I'm not sure yet when I'll have time for this, but probably not during the next 24h.

Kwpolska:
(Note we currently use `unidecode` which is a module that does the same as `translitcodec`

Kwpolska:
(Note we currently use `unidecode` which is a module that does the same as `translitcodec`, just with different opinions, and I’m pretty sure CJK text is complete garbage with both)

tests/test_utils.py
```diff
@@ -26,6 +26,7 @@ def test_getting_metadata_from_content(self):
         post = dummy()
         post.source_path = 'file_with_metadata'
         post.metadata_path = 'file_with_metadata.meta'
+        post.default_lang = 'en'
```
Kwpolska: Would be nicer to make a special `post_dummy` object with `default_lang = 'en'` already.

ralsina:
@felixfontain ok, let's give it a few days. I want to do a release before friday because I will then be unavailable for a few weeks.

tests/test_utils.py
felixfontein: I'm using `dummy` now, since that's precisely used in these situations.

felixfontein:
Seems to work fine. I created a mini plugin for testing, for those who are interested, which can be put into the `plugins/` folder of a blog:
The description file (`t.plugin`):
```
[Core]
Name = t
Module = t
```
The code (`t.py`):
```
# -*- coding: utf-8 -*-

from nikola.plugin_categories import SignalHandler
from nikola import utils
from unidecode import unidecode
import re


_slugify_strip_re = re.compile(r'[^+\w\s-]')
_slugify_hyphenate_re = re.compile(r'[-\s]+')


def my_slugify(value, lang=None, force=False):
    if not isinstance(value, utils.unicode_str):
        raise ValueError("Not a unicode object: {0}".format(value))
    value = value.replace('ä', 'ae').replace('ö', 'oe').replace('ü', 'ue').replace('ß', 'ss')
    value = value.replace('Ä', 'ae').replace('Ö', 'oe').replace('Ü', 'ue')
    value = utils.unicode_str(unidecode(value))
    value = _slugify_strip_re.sub('', value, re.UNICODE).strip().lower()
    return 'slugified-' + _slugify_hyphenate_re.sub('-', value, re.UNICODE)


class T(SignalHandler):  # Could also be any other plugin type which is initialized early enough.
    def set_site(self, site):
        super(T, self).set_site(site)
        utils.slugify = my_slugify
```
I recommend to remove the `'slugified-'` prefix for production use ;-)

So from my side, merging is fine.

felixfontein:
@ralsina: do you want to merge, or should I just do it myself?

Kwpolska:
Merged, let's hope it works.

felixfontein:
Thanks for merging!

Issue #2262, opened 2016-02-23
optional import_wordpress to single file


davidak:
instead of separate content and meta data

ralsina:
Should not be horribly hard now that Post objects can sort of save themselves.

felixfontein:
How about adding a new function to `basic_import` which allows to store metadata and content at the same time, and then changing the WordPress importer plugin to use this function? Then we can add a flag to `basic_import` which decides whether that new function stores the result in two files (by calling the existing methods, for example) or into one.

Kwpolska:
You just need to talk to a compiler’s `create_post` function, and provide the content to it (it’s already done by `nikola new_post -i` and Coil CMS). You don’t really need a new function for that.

felixfontein:
The functions in `basic_import` also take care of rewriting HTML if asked to, so either that functionality has to be refactored somewhere else, or another function is needed (which can call compiler's `create_post`).

ralsina:
Specifically, this code in import_wordpress.py and the methods it calls in basic_import.py need refactoring:

```python
                self.write_metadata(os.path.join(self.output_folder, out_folder,
                                                 out_meta_filename),
                                    title, meta_slug, post_date, description, tags, **other_meta)
                self.write_content(
                    os.path.join(self.output_folder,
                                 out_folder, out_content_filename),
                    content,
                    rewrite_html)
```

I would do a write_post in basic_import.py and just let that one sort it out. Not sure how many other importer plugins need touching to use that, so probably just add that new method and keep the current ones too.

ralsina:
Oh what the heck, in progress.

ralsina:
Done

PR #2292, opened 2016-03-08
Implement #2262

ralsina:
Adds the --one-file option to the wordpress importer to make it use one-file posts.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -827,7 +827,7 @@ class Nikola(object):
         # Set persistent state facility
         self.state = Persistor(os.path.join('state_data.json'))
 
-        # Set cache facility
+        # Set cache 0facility
         self.cache = Persistor(os.path.join(self.config['CACHE_FOLDER'], 'cache_data.json'))
 
     def init_plugins(self, commands_only=False):
@@ -870,7 +870,7 @@ class Nikola(object):
                     # FIXME TemplateSystem should not be needed
                     if p[-1].details.get('Nikola', 'PluginCategory') not in {'Command', 'Template'}:
                         bad_candidates.add(p)
-            else:  # Not commands-only
+            elif self.configured:  # Not commands-only, and configured
                 # Remove compilers we don't use
                 if p[-1].name in self.bad_compilers:
                     bad_candidates.add(p)
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -827,7 +827,7 @@ class Nikola(object):
         # Set persistent state facility
         self.state = Persistor(os.path.join('state_data.json'))
 
-        # Set cache 0facility
+        # Set cache facility
         self.cache = Persistor(os.path.join(self.config['CACHE_FOLDER'], 'cache_data.json'))
 
     def init_plugins(self, commands_only=False):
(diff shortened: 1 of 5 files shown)


nikola/nikola.py
```diff
@@ -827,7 +827,7 @@ def __init__(self, **config):
         # Set persistent state facility
         self.state = Persistor(os.path.join('state_data.json'))
 
-        # Set cache facility
+        # Set cache 0facility
```
Kwpolska: what’s a 0facility?

nikola/plugins/basic_import.py
```diff
@@ -140,6 +140,24 @@ def write_content(cls, filename, content, rewrite_html=True):
         with open(filename, "wb+") as fd:
             fd.write(content)
 
+    @classmethod
```
Kwpolska: Do we really need a classmethod here?

nikola/plugins/command/import_wordpress.py
```diff
@@ -196,9 +196,30 @@ class CommandImportWordpress(Command, ImportMixin):
             'default': 'first',
             'help': 'lower: Convert all tag and category names to lower case\nfirst: Keep first spelling of tag or category name',
         },
+        {
+            'name': 'one_file',
+            'long': 'one-file',
+            'default': False,
```
Kwpolska: We could probably default to `True`.

ralsina: Well, it would be a problem if someone re-imports the same wordpress dump
over his site, but it's just a corner case...

nikola/plugins/basic_import.py
ralsina: They all are classmethods, but probably no :-)

nikola/nikola.py
ralsina: it's a "Roberto has really fat fingers"

nikola/plugins/basic_import.py
felixfontein: Out of curiosity: what's so bad about making something `@classmethod` in Python?

nikola/plugins/command/import_wordpress.py
```diff
@@ -931,14 +954,32 @@ def import_postpage_item(self, item, wordpress_namespace, out_folder=None, attac
...
+
+                meta = {
+                    "title": title,
+                    "slug": meta_slug,
+                    "date": post_date,
+                    "description": description,
+                    "tags": ','.join(tags),
+                }
+                meta.update(other_meta)
+                if self.onefile:
```
felixfontein: Why not pass `self.onefile` as another argument to `write_post`, so that all importer plugins can use `write_post` for both single and non-single file output? Now all importer plugins which want to support both ways have to do such a rather complex `if`-`then`-`else` statement.

ralsina: Because compilers don't support saving in 2 files and I don't want to chase
that right now.

nikola/plugins/basic_import.py
Kwpolska: Why bother? A class method should only be used for alternate constructors
that return instances of the class (eg. datetime.datetime.fromtimestamp).
Other things should be regular methods. Static methods make sense
sometimes, but a regular method is enough.

-- 

ralsina:
Duly noted the problem of classmethods and the need to make write_post support both 1 and 2-file saves but those problems are not really related to this branch. Merging.

PR #2293, opened 2016-03-09
Fix #1839

### LINKED ISSUE #1839: Get rid of sys.exit() calls
We should return error codes from functions and use `sys.exit()` preferably only in `__main__.py`.

ralsina:
Avoid calling sys.exit() from plugins if possible.


nikola/plugins/task/galleries.py
```diff
@@ -94,8 +94,8 @@ def set_site(self, site):
         for source, dest in self.kw['gallery_folders'].items():
             if source in appearing_paths or dest in appearing_paths:
                 problem = source if source in appearing_paths else dest
-                utils.LOGGER.error("The gallery input or output folder '{0}' appears in more than one entry in GALLERY_FOLDERS, exiting.".format(problem))
-                sys.exit(1)
+                utils.LOGGER.error("The gallery input or output folder '{0}' appears in more than one entry in GALLERY_FOLDERS, ignoring.".format(problem))
+                continue
```
felixfontein: Why ignoring this instead of raising an exception? This will with high probability (haven't checked to be 100% sure) lead to an error as the plugin tries to create `index.html` in the output folder twice.

nikola/plugins/task/listings.py
```diff
@@ -73,7 +73,7 @@ def set_site(self, site):
             if source in appearing_paths or dest in appearing_paths:
                 problem = source if source in appearing_paths else dest
                 utils.LOGGER.error("The listings input or output folder '{0}' appears in more than one entry in LISTINGS_FOLDERS, exiting.".format(problem))
-                sys.exit(1)
+                continue
```
felixfontein: Same potential problem here.

nikola/plugins/task/tags.py
```diff
@@ -105,8 +105,7 @@ def gen_tasks(self):
             intersect = set(tags.keys()) & set(categories.keys())
             if len(intersect) > 0:
                 for slug in intersect:
-                    utils.LOGGER.error("Category '{0}' and tag '{1}' both have the same slug '{2}'!".format('/'.join(categories[slug]), tags[slug], slug))
```
felixfontein: Why did you got rid of the `join` here?

nikola/plugins/task/galleries.py
ralsina: And that will cause an error. No need to create a specific one here.

nikola/plugins/task/listings.py
ralsina: Same answer :-)

nikola/plugins/task/tags.py
ralsina: Because if a post was in more than one category it was shown as "cat1/cat2/cat3" and I think it's clearer to show it as "cat1, cat2, cat3"

felixfontein: But a post cannot be in multiple categories. After looking at it again, I think the `join` should have never been there in the first place. I think you can just do `.format(categories[slug], ...`.

PR #2294, opened 2016-03-09
WIP Fix #2246

### LINKED ISSUE #2246: In some cases posts generate conflicting cache files and shouldn't
If you have markdown (.md) and rst (.rst) posts with the same name and different extension and slug, they conflict on the cache:

```
$ ls posts/
fo.md  fo.rst

$ nikola build
Scanning posts......done!
ERROR: Task generation 'render_site' has duplicated definition of 'render_posts:cache/posts/fo.html'
```

Solution is probably to use something like ```fo.rst.html``` and ```fo.md.html``` as cache filenames.

ralsina:
Simplistic fix, should  be enough.

DO NOT MERGE, this breaks disqus identifiers (thanks invariance test!)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Features
 Bugfixes
 --------
 
+* Avoid cache clashes when post files differ only on extension (Issue #2246)
 * Fix a JSON conversion bug in the WordPress importer (Issue #2264)
 * Don’t create download directories when not downloading WordPress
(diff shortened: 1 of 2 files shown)


ralsina:
OH HELL this is bad. We are using post.base_path (which contains CACHE_FOLDER, BTW) in a bunch of places as unique post identifiers. So, ugh.

PR #2297, opened 2016-03-11
Fix error when run nikola without options.

qytz:
When I installed the plugin navstories, run nikola command without options get error:
```
# nikola 
Scanning posts.Cannot find 'rest' compiler; it might require an extra plugin -- do you have it installed?
```


ralsina:
I am not 100% sure this is the right solution but I don't quite remember that code... give me a few minutes.

ralsina:
On further reading the code:

1) Yes, this is correct
2) We need to add PluginCategory data to all plugins in the repo, or at least those which are commands. Opened #2298

ralsina:
Thanks for the fix @qytz !!!!

qytz:
Yes, it's better to add PluginCategory data to all plugins.

PR #2310, opened 2016-03-28
Fix typo

whiteroses:
Thanks!


Kwpolska:
Thanks for your contribution! (You might want to report this upstream to docutils, where we stole that document from.)

whiteroses:
Will do!

PR #2311, opened 2016-04-04
Added compile plugin using markdown2 processor

atiro:
Added a plugin to use the https://github.com/trentm/python-markdown2 processor (mainly to take advantage of CSS tags support, especially for tables).

It seems to work, but could probably be much improved.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -381,6 +381,7 @@ class Nikola(object):
             'COMPILERS': {
                 "rest": ('.txt', '.rst'),
                 "markdown": ('.md', '.mdown', '.markdown'),
+                "markdown2": ('.md', '.mdown', '.markdown'),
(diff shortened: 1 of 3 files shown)


Kwpolska:
Thanks for your contribution! Unfortunately, we can’t accept this contribution to the main repository. However, we would love to have this in our **[plugins repo](https://getnikola.com/getnikola/plugins)**. Please contribute your plugin there. Make sure to read the `README.md` file in that repository for details and requirements (the added meta information in particular).

Also, you modified our default compiler setup. This should not be done without good reason, especially because the `markdown` plugin is not going to be removed any time soon.

Also, it looks like your commit was misattributed. Please add the e-mail address you used for this commit [to your GitHub profile](https://github.com/settings/emails) so your commits will appear as yours.

PR #2316, opened 2016-04-09
Moved `<footer>`outside of `<div role='main'>`

### LINKED ISSUE #2309: move `<footer>` outside of `<main>`
Imo the footer of a page is not part of the main content of a site (just like for example a menu). Thus, the last `<footer>` element should be moved outside and below `<main>` in the default themes for accessibility purposes.

sukiletxe:
This enhances accessibility, particularly because a footer is not part of
the main content of the page.
Fixes #2309

By the way, I didn't Jinjify the result, because I tried, and a lot of whitespace changes (evil ^m on Windows) apeared everywhere as a result. So I thought you'd prefer a cleaner diff.

### DIFF
--- a/nikola/data/themes/bootstrap3/templates/base.tmpl
+++ b/nikola/data/themes/bootstrap3/templates/base.tmpl
@@ -66,12 +66,11 @@ ${template_hooks['extra_head']()}
             <%block name="content"></%block>
         </div>
         <!--End of body content-->
-
-        <footer id="footer">
-            ${content_footer}
-            ${template_hooks['page_footer']()}
-        </footer>
     </div>
+    <footer id="footer">
+        ${content_footer}
+        ${template_hooks['page_footer']()}
+    </footer>
 </div>
 
 ${base.late_load_js()}


Kwpolska:
Bootstrap itself mixes it in their examples… would you mind complaining there first?

tritium21:
The test failure is in the invariant build.  The test should be updated with this PR.

The test is checking the html output, and comparing it to a per-modification version; you just have to make the same change to the test that you made to the templates.

Kwpolska:
The test failure cannot be averted by contributors and needs to be handled by a core dev (who has access to the invariant-builds repo, which introduces a race-condition)

tritium21:
Oh.  I didn't know that part.  At least it is documented why the one travis build is failing then.

sukiletxe:
Okay, so I have filed an issue, and they are willing to fix it. The 
problem is, which examples should I point out to them? I downloaded the 
latest release and I've looked at the examples (docs / examples), and I 
can see none with this behaviour, nor with a `<main>` and a `<footer>`.
Perhaps you meant a theme?

Kwpolska:
footer within div#container: https://getbootstrap.com/examples/jumbotron/ or https://getbootstrap.com/examples/carousel/  

footer outside of div#container: https://getbootstrap.com/examples/blog/ or https://getbootstrap.com/examples/sticky-footer/

Looking closer, it might just be *intentional* for styling and not for accessibility.

If you moved `footer` outside of `.container`, you would ruin its appearance. This patch moves it from `.row` up to `div.body-content`, which does not change anything (especially `role="main"`). That said, I’m afraid I have to close this PR…

But thanks for caring!

sukiletxe:
OK, thanks for the info.
Regarding your second comment: perhaps an extra `<div>` could be created 
(which encloses the `<div>`with `role="main"`), and the role could be 
set for the inside `<div>`. Then, the inner `<div>` could be closed and 
the `<footer>` could go after it. This would solve the issue, and I 
think it wouldn't affect styling.

sukiletxe:
Looking through old issues, I see that this was rejected in Bootstrap's side (see twbs/bootstrap#19730). So... May I ask you to reconsider this? Thank you very much.

Kwpolska:
If they won’t fix it, we also have no reason to. Fix it in your custom templates.

PR #2320, opened 2016-04-18
Pass positional arguments to template-based shortcodes

SpotlightKid:
Implements #2319. See rationale there.

Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

PR #2321, opened 2016-04-19
Update AUTHORS.txt, CHANGES.txt and docs according to #2320

SpotlightKid:
And fix minor issues in documentation on shortcodes.

Some previously existing documentation from the main manual is duplicated in other words in `extending.txt`, but I believe, it makes sense to have it both places with differing depth.

Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -854,12 +854,16 @@ chart
 Template-based shortcodes
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you put a template in ``shortcodes/`` called ``mycode.tmpl`` then Nikola will create a shortcode
-called ``mycode`` you can use. Any options you pass to the shortcode will be available as variables
-for that template. If you use the shortcode as paired, then the contents between the paired tags will
-be available in the ``data`` variable.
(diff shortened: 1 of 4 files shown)


docs/manual.txt
```diff
@@ -854,12 +854,16 @@ chart
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you put a template in ``shortcodes/`` called ``mycode.tmpl`` then Nikola will create a shortcode
-called ``mycode`` you can use. Any options you pass to the shortcode will be available as variables
-for that template. If you use the shortcode as paired, then the contents between the paired tags will
-be available in the ``data`` variable.
+If you put a template in ``shortcodes/`` called ``mycode.tmpl`` then Nikola
+will create a shortcode called ``mycode`` you can use. Any options you pass to
+the shortcode will be available as variables for that template. Non-keyword
+options will be passed in a tuple varaible named ``_args``.
```
ralsina: The part about the contents when using shortcodes in the paired form seems to have been missed in the transition here.

ralsina: Ah, ignore me, it's below :-)

PR #2327, opened 2016-04-27
fix string format specifiers

tikki:
Format specifiers written right-to-left don't work for python string formatting; leads to "ValueError: incomplete format" when running a build.


ralsina:
Interesting! I'll have to trust you on this :-)

Kwpolska:
(PS. You may want to configure a valid e-mail on GitHub and in git on your machine.)

PR #2329, opened 2016-05-01
'saniziting' is not how you spell it!

dmwyatt:
I couldn't get this command working forever until I realized I was spelling sanitizing correctly!


Kwpolska:
Good catch!

PR #2330, opened 2016-05-01
If a title has line breaks it messes up meta generation.

dmwyatt:


nikola/plugins/command/import_wordpress.py
```diff
@@ -808,6 +808,12 @@ def import_postpage_item(self, item, wordpress_namespace, out_folder=None, attac
             out_folder = 'posts'
 
         title = get_text_tag(item, 'title', 'NO TITLE')
+
+        # titles can have line breaks in them, particularly when they are
+        # created by third-party tools that post to Wordpress.
+        # Handle windows-style and unix-style line endings.
+        title = title.replace('\r\n', ' ').replace('\n')
```
Kwpolska: missing the second argument to str.replace in the second call

Kwpolska:
Thanks for your contribution!

PR #2338, opened 2016-05-06
Added enclosure lengths so podcast RSS is valid for more/all readers

x1101:
Prior to this change, RSS XML generate included the enclosure URL, but a file length of "0". This causes many podcatchers to reject the feed. This resolves that issue.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -293,9 +293,9 @@ LEGAL_VALUES = {
 
 def _enclosure(post, lang):
     """Add an enclosure to RSS."""
-    enclosure = post.meta('enclosure_url', lang)
+    enclosure = post.meta('enclosure', lang)
     if enclosure:
-        length = post.meta('enclosure_length', lang)
+        length = int(post.meta('enclosure_length', lang) or 0)
         url = enclosure
         mime = mimetypes.guess_type(url)[0]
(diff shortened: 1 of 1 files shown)


nikola/nikola.py
```diff
@@ -293,9 +293,9 @@
 
 def _enclosure(post, lang):
     """Add an enclosure to RSS."""
-    enclosure = post.meta('enclosure', lang)
+    enclosure = post.meta('enclosure_url', lang)
```
Kwpolska: Don’t change the name.

nikola/nikola.py
```diff
@@ -293,9 +293,9 @@
...
     if enclosure:
-        length = 0
+        length = post.meta('enclosure_length', lang)
```
Kwpolska: Default to zero and use integers. 
```py
length = int(post.meta('enclosure_length', lang) or 0)

x1101:
Applied cleanup as suggested.

Kwpolska:
Right, that’s still not enough. I’ll fix it myself and merge.

Thanks for your contribution!

PS. you may want to set your Git client to something else than `Your full name` for `user.name`.

Kwpolska:
Merged and fixed in commit cfc8554.

PR #2339, opened 2016-05-08
Show post's location even when it exists.

punchagan:
It is useful to print the post's location even when it already exists, similar to how it is printed when we create a new post.  This is especially useful when the command is being called from a script, and the post's location needs to be parsed.


Kwpolska:
Not sure about this, especially about the phrasing — perhaps “existing post” instead of “your post” would be better?

punchagan:
I wasn't sure about it, and have been using a plugin to [do this](https://github.com/punchagan/punchagan.muse-amuse.in/blob/master/plugins/print_post_paths/print_post_paths.py) until now.

ralsina:
I like it with the "Existing post" text. Merging.

PR #2341, opened 2016-05-13
Add <link> even if post is untranslated (to stop Google complaints)

Kwpolska:
Solves this e-mail I received from Google Webmaster Tools today:

> Google has detected that some pages on your site have implemented the rel-alternate-hreflang tag incorrectly. In particular, there seems to be a problem with incorrect language and region codes, or incorrect bi-directional linking (if page A links with hreflang to page B, there must be a link back from B to A as well). Google uses the hreflang attributes to serve the correct language or regional URL to the right users in search results.
> 
> Until you correct the errors on the hreflang links, your website will not benefit from additional language or region targeting.
> 
> Fix this problem now:
> 
> Find incorrect hreflang implementation
> Use the examples provided in the International Targeting feature in Search Console to get a sample of pages with incorrect hreflang implementation.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,8 @@ Features
 Bugfixes
 --------
 
+* Add ``<link>`` tags to other languages even if the post is
+  untranslated (Google complained otherwise)
 * Create Persistor directories only if site is configured (Issue #2334)
 * Remove newlines in imported WordPress titles (Issue #2330)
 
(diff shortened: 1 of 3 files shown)


Aeyoun:
I’ll check this format with Baidu, Bing, and Yandex too. Takes ~3 days for them to process.

Aeyoun:
Looks fine. :+1:

Kwpolska:
Merging, thanks for testing this.

PR #2344, opened 2016-05-19
Cache messages that we have already loaded.

punchagan:
Messages were being loaded each time the `MESSAGES` attribute of `site`
was accessed. This commit switches to using cached values.


ralsina:
Neat!

punchagan:
Done.

Kwpolska:
Merged, thanks!

PR #2353, opened 2016-05-30
Remove incorrect attempt to change ipynb template

punchagan:
The previous code intended to set the template to basic, but it doesn't
really work and the full template was being used.

### DIFF
--- a/nikola/plugins/compile/ipynb.py
+++ b/nikola/plugins/compile/ipynb.py
@@ -82,7 +82,6 @@ class CompileIPynb(PageCompiler):
         """Export notebooks as HTML strings."""
         if flag is None:
             req_missing(['ipython[notebook]>=2.0.0'], 'build this site (compile ipynb)')
-        HTMLExporter.default_template = 'basic'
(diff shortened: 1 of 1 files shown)


punchagan:
@damianavila Does this change look OK to you? It works fine for me.

damianavila:
As long as we are not breaking the `status quo`, I am OK.
Originally the idea was to use the basic template so anyone can do their own stuff... BUT...
Which is the default experience if we turn this to actually use the basic template?
I don't want to break the current user experience in the default setup...

punchagan:
@damianavila Fair enough.  I just removed the confusing line which attempts to change the template, but doesn't work. 

Users can use `IPYNB_CONFIG` if they really want the basic template.

Kwpolska:
Let’s merge this if the original was broken.

punchagan:
Just to clarify, the PR now just removes the code which tries to change the template to basic, but doesn't work correctly.  The full template will be used, and users can switch to basic template using the `IPYNB_CONFIG` setting.

damianavila:
Yep, understood... I agree with the merge. Thanks @punchagan

PR #2363, opened 2016-06-04
fix #2360

### LINKED ISSUE #2360: output of 'nikola auto' not visible with light themes
I use solarized-light theme in my(gnome) terminal and when I e.g. run 'nikola build', most of the output is not clearly visible - see the shot: http://pasteboard.co/1pIJ9kw9.png, so wonder if something can be done to make it more ui-friendly for light color themes?

ralsina:
If NIKOLA_MONO is set, don't use colors.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Option to disable color output using NIKOLA_MONO envvar (Issue #2360)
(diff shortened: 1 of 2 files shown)


Kwpolska:
Why not make this a config option instead (say, `__colorful__`)? Environment variables are a bit ugly in general. We could move things around a bit to facilitate that.

ralsina:
because it also should affect things done without config, like init

PR #2364, opened 2016-06-04
When there is more than one copy of a plugin, use the most local one (fix #2362)

### LINKED ISSUE #2362: Multiple copies of a plugin are activated
When you have the same plugin installed in more than one place, instead of using the "most local" one, both are activated, which can cause conflicts.

Proposed solution: if there are more than 1 copy of the plugin, deactivate those that are less local.

ralsina:


nikola/nikola.py
```diff
@@ -930,6 +929,23 @@ def init_plugins(self, commands_only=False, load_all=False):
                         bad_candidates.add(p)
                         utils.LOGGER.debug('Not loading compiler extension {}', p[-1].name)
             self.plugin_manager._candidates = list(set(self.plugin_manager._candidates) - bad_candidates)
+
+        # Find repeated plugins and discard the less local copy
+        def plugin_position_in_places(plugin):
+            for i, place in enumerate(self._plugin_places):
+                if plugin[0].startswith(place):
```
Kwpolska: could use some explainer comments that tell future readers what `plugin` is here, it took me a while to parse all this

nikola/nikola.py
```diff
@@ -930,6 +929,23 @@ def init_plugins(self, commands_only=False, load_all=False):
...
+                    return i
+
+        plugin_dict = defaultdict(list)
+        for data in self.plugin_manager._candidates:
+            plugin_dict[data[2].name].append(data)
```
Kwpolska: why `[2]`, what are the previous positions?

ralsina: they are "path to the .plugin file, path to the module, metadata about the
plugin"

nikola/nikola.py
```diff
@@ -28,7 +28,6 @@
 
 from __future__ import print_function, unicode_literals
 import io
-from collections import defaultdict
```
Kwpolska: Breaks everything!

ralsina: Looks like it. I'll take a look when I come back

Kwpolska: I fixed it and I’m currently testing this. Will merge if all works okay (which it currently does)

Kwpolska:
Works okay, merging if there are no errors.

Issue #2369, opened 2016-06-13
Add a signal after a post is compiled.


tritium21:
I want to write a plugin to post-process the compiled HTML fragment output by Pandoc, though what I want to do could be applied to any of the compilers.  I need to hook into Nikola before post teasers are extracted from the compiled HTML fragments.  There is no way to do what I want without re-implementing the compiler plugin.

For Pandoc, this is no great burden - the method in question is 11 lines of code, and the entire module, including boilerplate and license is only 85 lines long.  It is almost trivial to just replace the plugin.  However, if I wanted to do something more general, and support RestructuredText, for example, that is a much bigger burden.

I propose that a signal be added (probably to ``nikola.post.Post.text(...)`` or ``nikola.post.Post.compile(...)``) to allow SignalHandler plugins to intercept when a page is compiled to cache, before nikola does anything else with it.  The data on the signal should be a dictionary of the post source path, the cache file path, and the post object; perhaps more, but that is what comes to mind at least to me.

I considered suggesting that the signal be emitted from the compiler's ``compile_html`` method, however I decided against suggesting that - that would change the plugin api on everyone implementing compilers.

Kwpolska:
:+1: for this.

Kwpolska:
Closing via #2370.

PR #2370, opened 2016-06-13
Added 'compiled' signal

tritium21:
Adds a signal to ``nikola.post.Post().compile()`` after the page is compiled, before anything is done with the compiled file.

addresses #2369

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -43,6 +43,7 @@ except ImportError:
 
 from . import utils
 
+from blinker import signal
 import dateutil.tz
 import lxml.html
 import natsort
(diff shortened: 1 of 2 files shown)


nikola/post.py
```diff
@@ -491,6 +492,12 @@ def wrap_encrypt(path, password):
             dest,
             self.is_two_file),
 
+        signal('compiled').send(dict(
```
Kwpolska: We like the `{}` syntax more, I’ll fix that.

Kwpolska:
Looks good, thanks!

(I’ll make some minor style fixes and add this to the changelog.)

PR #2375, opened 2016-06-16
WIP: Initial snapcraft packaging

ralsina:
Still experimental.


snapcraft/README
```diff
@@ -0,0 +1,26 @@
...
+
+To build, in Ubuntu 16.04:
+
+apt install snapd snapcraft
+snap login some.account@whatever.com
+snapcraft
+cp nikola.py stage/usr/bin/nikola
+snapcraft
+
+To just use it:
```
Kwpolska: This should live in our getting started document when snaps actually work (i.e. you can build the demo site correctly). We should also mention other distros, and provide a link to snapcraft.io.

ralsina: Yes, currently it's just a draft so I can keep track of where I am with
this.

ralsina:
Merging it for now. In order for this to be fully operational, we need a couple more features in snapscraft (pushing package to a specific channel, maybe some new adaptors)

PR #2379, opened 2016-06-18
Allow setting custom GUID in feeds (fixes #2378)

### LINKED ISSUE #2378: Allow setting custom GUID in feeds
It should be possible to set the GUID in RSS feeds or ID in Atom feeds to a value set in the post metadata. This helps renaming pages or moving the whole site to a new domain without the posts showing up as new in feedreaders.

I've already implemented this and I will submit a pull request soon.

I'm not yet sure about a detail: What should be the default GUID of new posts? It could either be their URL as it always is at the moment or a random value could be written to the file.

julianbrost:


Aeyoun:
RFC 4287 section-4.2.6: “[the ID value] MUST be an IRI”.

Kwpolska:
:-1:

Kwpolska:
The consensus seems to be to reject this idea. I’ll close this.

PR #2383, opened 2016-06-27
Add horizontal scrollbar to listings

Kwpolska:
This replaces code tables that would go beyond screen width with code tables that always have horizontal scrollbars. This is not a perfect solution, though — it looks uglier when there is no overflow:

image

There’s also `overflow-x: overlay;` which looks like this and partially covers text:

image

Which one is better for most users? Opinions welcome.

cc @iAmMrinal0 — via getnikola/nikola-themes#86 (pending)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,7 @@ Features
 Bugfixes
 --------
 
+* Add horizontal scrollbar to listings (via getnikola/nikola-themes#86)
 * Copy files when importing two-file posts instead of reading and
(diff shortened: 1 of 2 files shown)


iAmMrinal0:
I prefer the first one. The main point of having a design is readability, which is hard in the second design even if it's a single line.

felixfontein:
Aren't these *horizontal* scrolling bars?

Kwpolska:
Oops, one typo that went everywhere. Should be fixed everywhere.

Kwpolska:
Merging.

PR #2394, opened 2016-07-13
Fix #1889 -- support dates in post_list

### LINKED ISSUE #1889: Augment post-list directive with more filters
Currently, filtering posts that go into post-lists only works with tags or slugs. In the legacy site I currently want to migrate, I have views in the form of /press/releases/<year>/ that would filter by date but could also filter by globbing the path name. I wonder if there is a generic way to filter by multiple meta data entries.

While we're at it: once you add powerful filters to post-lists and get something what is called "the loop" in php sites, i.e. filter/sort/slice/templatize, the currently implemented indexes for posts-by-year, posts-by-author and posts-by-category just become special cases of these posts-lists. You could basically create paginated indexes for all posts-list queries and offer rss feeds for each of them, too.

Kwpolska:
This adds support for date filtering in post_list using a set of filters. I vendored in my [datecond](https://github.com/Kwpolska/datecond) library, under the 3-clause BSD license (but if anyone objects, we could make it an optional external dependency, or I could even agree to switch to dual-licensing)

Example:

```rst
.. post-list::
   :date: year < 2016, month != 5
```

Shows all posts written before 2016 and not written in May.

cc @erdgeist (reporter of #1889)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,8 @@ New in master
 Features
 --------
 
+* Support ``date`` filtering in the post list directive
+  (Issue #1889)
 * Added Albanian translation by Vango Stavro
 * Added ``post-(type)`` class to ``story.tmpl`` (uses the ``type``
   meta field, defaults to ``post-text`` — same behavior as posts)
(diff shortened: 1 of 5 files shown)


ralsina:
This looks pretty awesome!

ralsina:
Sorry, I merged a couple of branches, if you fix conflicts, feel free to merge, code LGTM

PR #2395, opened 2016-07-18
Improved EXIF orientation handling

ralsina:
Improve EXIF orientation handling according to http://www.daveperrett.com/articles/2012/07/28/exif-orientation-handling-is-a-ghetto/

To test IRL, get the test images, put them in a gallery, set PRESERVE_EXIF to True and it should work well.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Features
 Bugfixes
 --------
 
+* Improved EXIF orientation handling
 * Register post list template as a dependency (Issue #2391)
 * Fix section color hashing when using Python 2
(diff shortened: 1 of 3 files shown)


Kwpolska:
Do we need an extra dependency for this? Or maybe we could vendor it in?

ralsina:
Sadly, yes, because Pillow's support for EXIF is lame. It's a pure-python
dependency, so it should not be much of a problem. We could vendor it in,
yes.

ralsina:
This fix would not be worth it, but the upcoming exif filtering feature
makes it worth it, if nothing else for privacy reasons.

PR #2399, opened 2016-07-24
Remove leading slash in redirections

ChillarAnand:


nikola/plugins/task/redirect.py
```diff
@@ -50,7 +50,7 @@ def gen_tasks(self):
         yield self.group_task()
         if kw['redirections']:
             for src, dst in kw["redirections"]:
-                src_path = os.path.join(kw["output_folder"], src)
+                src_path = os.path.join(kw["output_folder"], src.rstrip('/'))
```
Kwpolska: leading slash is `lstrip` :unamused:

ChillarAnand: Duh! Updated

Kwpolska:
This is actually a good way to solve this. Thanks!

PR #2405, opened 2016-07-26
Update jQuery to v3.1.0

Kwpolska:
This updates jQuery to v3.1.0, which might break backwards compatibility. As such, testing is required. Please test with your own sites and JS stuff.


Kwpolska:
Breaks flowr. We can’t go for a newer jQuery unless we change *at least* that.

PR #2417, opened 2016-07-30
Fix #2412 -- Register dependencies in post-list shortcode

### LINKED ISSUE #2412: post-list shortcode does not care about dependencies
We need to make it possible for shortcodes to add dependencies to posts, like is done with reST. Perhaps we could just output `.dep` files from post-list if reST is not used?

Kwpolska:
This is #2412.

cc @ralsina — requires review due to some refactoring this caused.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Features
 Bugfixes
 --------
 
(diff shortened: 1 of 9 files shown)


ralsina:
Looks sane to me.

Issue #2418, opened 2016-07-30
Image rotation in gallery does not work well


tikank:
I have added an image to gallery. The image has orientation which is not 1 (normal), so gallery rotates it, but the result is not good. Size of the image is not changed and resulting image have black bands on sides and parts of the image is cropped.

If I follow code correctly, the rotation is done in nikola/image_processing.py and uses Image.rotate method. It should have expand=True flag to set the size correctly. Or maybe it would be better use Image.transpose method to rotate or flip the image.

ralsina:
On what version of Nikola?

tikank:
Nikola v7.7.11

tikank:
I made a fix for this. I'll try to make pull request.

I also generated some test images using this script:

```
from PIL import Image, ImageDraw
import piexif

def save(orientation, im):
    im.save('{}_up.jpg'.format(orientation), format='jpeg')
    exif_data = { '0th' : { piexif.ImageIFD.Orientation : orientation } }
    im.save('{}_up_o.jpg'.format(orientation), format='jpeg', exif=piexif.dump(exif_data))

im = Image.new('RGBA', (100, 200), color=(255,255,255,255))
draw = ImageDraw.Draw(im)
draw.rectangle([(5,5), (95,195)], outline=(0,0,0,255))
draw.line([(50,80), (50,180)], fill=(0,0,0,255))
draw.line([(50,80), (20,110)], fill=(0,0,0,255))
draw.line([(50,80), (80,110)], fill=(0,0,0,255))
text = "This\nside\nup"
(w, h) = draw.textsize(text)
draw.text((50 - w/2, 20), text, fill=(0,0,0,255), align='center')

save(1, im.copy())
save(2, im.copy().transpose(Image.FLIP_LEFT_RIGHT))
save(3, im.copy().transpose(Image.ROTATE_180))
save(4, im.copy().transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_180))
save(5, im.copy().transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_90))
save(6, im.copy().transpose(Image.ROTATE_90))
save(7, im.copy().transpose(Image.FLIP_LEFT_RIGHT).transpose(Image.ROTATE_270))
save(8, im.copy().transpose(Image.ROTATE_270))

```

ralsina:
Awesome!

ralsina:
Fixed by #2421

PR #2421, opened 2016-08-01
Fix for gallery image rotation

tikank:
Fix for issue #2418. Rotate is replaced with transpose which sets image size correctly.

### DIFF
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -39,12 +39,11 @@ from nikola import utils
 
 Image = None
 try:
-    from PIL import ExifTags, Image, ImageOps  # NOQA
+    from PIL import ExifTags, Image  # NOQA
 except ImportError:
     try:
         import ExifTags
         import Image as _Image
-        import ImageOps
         Image = _Image
     except ImportError:
         pass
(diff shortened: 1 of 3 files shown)


nikola/image_processing.py
```diff
@@ -117,13 +117,13 @@ def resize_image(self, src, dst, max_size, bigger_panoramas=True, preserve_exif_
...
+                im = im.transpose(Image.ROTATE_180)
             elif value in (5, 6):
-                im = im.rotate(270)
+                im = im.transpose(Image.ROTATE_270)
             elif value in (7, 8):
-                im = im.rotate(90)
+                im = im.transpose(Image.ROTATE_90)
             if value in (2, 4, 5, 7):
-                im = ImageOps.mirror(im)
+                im = im.transpose(Image.FLIP_LEFT_RIGHT)
```
ralsina: This means you can delete the ImageOps import, I think (Travis tests will fail if you don't)

ralsina:
LGTM, could you add a 1-liner in CHANGES and add yourself to AUTHORS?

tikank:
Added.

ralsina:
Cool! Thanks for the contribution!

PR #2423, opened 2016-08-02
Fix #2422

### LINKED ISSUE #2422: admonitions (warning, attention, hint, ...) not localized
(reST?) Admonitions are not localized in Nikola. The Docutils localization strings are already present in **docutils/languages/de.py**, but changing DEFAULT_LANG to "de" in conf.py does not use these localized strings when using the following (reST) code:

    .. attention:: Headline
    
            Contents

The heading should be "Achtung!" but instead it is "Attention!"

ralsina:
Pass current language to docutils so things lik ```.. contents::``` and admonitions are localized correctly.

### DIFF
--- a/nikola/plugins/compile/rest/__init__.py
+++ b/nikola/plugins/compile/rest/__init__.py
@@ -40,7 +40,14 @@ import docutils.parsers.rst.directives
 from docutils.parsers.rst import roles
 
 from nikola.plugin_categories import PageCompiler
-from nikola.utils import unicode_str, get_logger, makedirs, write_metadata, STDERR_HANDLER
+from nikola.utils import (
+    unicode_str,
+    get_logger,
+    makedirs,
+    write_metadata,
+    STDERR_HANDLER,
+    LocaleBorg
+)
 from nikola.shortcodes import apply_shortcodes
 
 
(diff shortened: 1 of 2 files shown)


nikola/plugins/compile/rest/__init__.py
```diff
@@ -63,16 +70,19 @@ def compile_html_string(self, data, source_path=None, is_two_file=True):
...
         default_template_path = os.path.join(os.path.dirname(__file__), 'template.txt')
+        settings_overrides = {
+            'initial_header_level': 1,
+            'record_dependencies': True,
+            'stylesheet_path': None,
+            'link_stylesheet': True,
+            'syntax_highlight': 'short',
+            'math_output': 'mathjax',
+            'template': default_template_path,
+            'language_code': LocaleBorg().current_lang,
```
Kwpolska: How does this react to `pt_BR` which docutils wants as `pt_br`? Should we perhaps make another dict for docutils languages in LEGAL_VALUES in nikola.py?

ralsina: ugh?

ralsina: Just checked, we also have pt_br as our translation name, so it all works out.

Kwpolska: There will be warnings if a language is unsupported by docutils, but I guess we can live with that. LGTM, merge at will.

Issue #2229, opened 2016-01-25
Make Nikola show the user's Internet footprint.


ralsina:
The title is mighty pretentious but the concept is rather simpler ;-)

Nikola should be able to be a hub where all the "important" internet content creation of a user is published.
- You post book reviews on goodreads? Then they should appear in Nikola
- You publish videos in Youtube? Then they should appear in Nikola
- You tweet a lot? Then you should be able to make some specially marked tweets appear in Nikola
- You have a tumblr? Same thing

And so on. I have some ideas about _how_ to do this, too:
- Base it in the idea of planetoid: RSS/Atom feeds are provided by most things
- Use the new proposed state feature to keep last-checked dates and so on
- Use the new templated shortcodes feature to process the data

So, for example: goodreads -> RSS -> JSON -> template -> new post

It should not be horribly hard to do.

gour:
Several times you write 'should', so I wonder whether it should be 'could' ?

ralsina:
I am working on a proof-of-concept in https://github.com/getnikola/nikola/pull/2428

punchagan:
- Would it be useful to support API calls, where a feed is not available?  For example: facebook/twitter posts. 
- Is there a reason this is being implemented as a core plugin (nikola repo) as opposed to an additional plugin (plugins repo)?

Aeyoun:
I wrote something like this for myself a few months ago. What I found very useful was the ability to search-and-replace in the feed titles. With any given feed, I have one or two associated regular expressions to clean up and modify the feed titles to make them more suitable for publishing.

For example, I would turn “Aeyoun published a new video on Vimeo: ‘Video Title’” to “New video: ‘Video Title’”, or “Aeyoun opened a new issue in getnikola/getnikola” to “I opened a new bug in getnikola on GitHub”, or “Aeyoun’s new photo on Flickr: Blue Pants” to “Photo: Blue Pants”.

reubano:
@aeyoun mind sharing your code?

ralsina:
Closing this since I am making it a plugin and releasing it.

PR #2428, opened 2016-08-04
WIP Initial implementation of #2229

ralsina:
Still needs a lot of work but it does work for goodreads


Kwpolska:
How is this different from the planet plugin? Could those two be merged?

ralsina:
@Kwpolska it's a more mature reimplementation of the basic concept, plus it uses feed-specific templates, so we can show each feed nicely.

ralsina:
I am making this into a non-core plugin so I can release it and work on it iteratively.

PR #2429, opened 2016-08-04
Added EXTRA_THEMES_DIRS.

### LINKED ISSUE #2427: EXTRA_THEMES_DIRS?
I started reorganizing my Nikola projects (by moving them around), when I noticed that Nikola will load themes only from `./themes` or from the `themes` directory from Nikola's install, and that it is not possible to specify additional theme dirs (similar to `EXTRA_PLUGINS_DIRS`).

Is there a reason for this? Or does anyone mind if I add a setting `EXTRA_THEMES_DIRS`?

felixfontein:
Implements a first version of `EXTRA_THEMES_DIRS`. Requires that `site.plugins_dirs` is passed to some theme-related functions in `utils.py`. This might break some other plugins outside this repository. (The ones in this repository are all adjusted accordingly.)

fixes #2427


felixfontein:
I've tested this a bit, works well for me so far.

What might be missing: example entry in `conf.py`. I'm kind of reluctant to add this yet as this feature might not work with some third-party plugins (yet). On the other hand, if this feature isn't documented properly enough, nobody will bother to add support to other plugins.

Any thoughts on this?

ralsina:
I am not sure about this. You seem to be passing THEMES_DIRS everywhere... once you have built the theme chain, why do you need to pass it? The theme chain should have all the needed data.

felixfontein:
That would definitely simplify it. I'll look into it later.

felixfontein:
Ok, I tried that. Seems to work. I hope no other plugin is assuming that `site.THEMES` contains the theme names.

felixfontein:
@ralsina: any more thoughts on this?

ralsina:
LGTM merging!

felixfontein:
Thanks :)

PR #2430, opened 2016-08-06
Update Pygal chart options

andredias:
There are a few new Pygal 2.2.3 options that could be used in chart directive.


nikola/plugins/compile/rest/chart.py
```diff
@@ -73,52 +73,68 @@ class Chart(Directive):
     has_content = True
     required_arguments = 1
     option_spec = {
-        "copy": directives.unchanged,
+        "box_mode": directives.unchanged,
+        "classes": directives.unchanged,
         "css": directives.unchanged,
+        # "defs": directives.unchanged,
```
Kwpolska: Why is that commented out?

andredias: I thought not to include this option because it should receive a [multiline value](http://www.pygal.org/en/stable/documentation/configuration/rendering.html#defs) and it is kind of hard to write in a rst directive. But it could be accomplished using '\n' and spaces. So it can be included after all. Should I provide another pull request?

Kwpolska: If you can get it to work, you can just push one more commit to your fork with this uncommented.

ralsina:
Also, please add a one-liner in CHANGES and yourself to AUTHORS if you are not already there :-)

Kwpolska:
One more housekeeping request, it looks like your commit was misattributed. Please add the e-mail address you used  (pronus dot io) for this commit [to your GitHub profile](https://github.com/settings/emails) so your commits will appear as yours.

Thanks!

PR #2435, opened 2016-08-08
sorted imports

ralsina:
I have been using this in other projects and it does make the code slightly cleaner if run once in a while.


ralsina:
Forget about it, breaks the world, too boring to fix.

PR #2436, opened 2016-08-09
Update Persian translation

shahinism:
Provide translation for new words.


ralsina:
The translations are handled at transifex:
https://www.transifex.com/ralsina/nikola/dashboard/ please create an
account there and translate all you want :-)

Issue #1286, opened 2014-05-16
Image metadata support


devurandom:
It would be great if metadata could be added to images. This could then be displayed in the galleries, e.g. as a caption below the image.

Process could be as follows:
1. Extract metadata from `image.jpg`, i.e. EXIF
2. Add/replace with data from `image.jpg.meta`, e.g. `.. caption: …` or `.. date: …`
3. Generate captions, etc.

Step 1 is just mentioned for completeness and describes only a minor and distant wish. This request is primarily about steps 2 and 3.

devurandom:
Related to #786?

guyou:
I started a feature branch on this topic:
https://github.com/guyou/nikola/tree/feature-images-metadata

It is quite functionnal and expect your review.

BryanQuigley:
Would this let me do something like https://irongiantdesign.com ?

IN this case it's implemented in Wordpress posts and just displays the first picture in the post.  Thanks!

Kwpolska:
@BryanQuigley No, but you can just use the existing `previewimage` meta field on posts and create custom templates that make use of it.

PR #2438, opened 2016-08-09
[WIP] Feature images metadata

guyou:
In order to offer the ability to create theme with advanced image galleries, nikola needs to offer a way to adds extended/manual metadata to each image files.

This feature branch add a support for additional .meta file associated to image files.

Cf. #1286

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -455,7 +455,6 @@ class Nikola(object):
             'FRONT_INDEX_HEADER': '',
             'GALLERY_FOLDERS': {'galleries': 'galleries'},
             'GALLERY_SORT_BY_DATE': True,
-            'GALLERY_USE_METADATA': True,
             'GLOBAL_CONTEXT_FILLER': [],
             'GZIP_COMMAND': None,
             'GZIP_FILES': False,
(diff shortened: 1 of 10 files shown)


nikola/nikola.py
```diff
@@ -455,6 +455,7 @@ def __init__(self, **config):
             'FRONT_INDEX_HEADER': '',
             'GALLERY_FOLDERS': {'galleries': 'galleries'},
             'GALLERY_SORT_BY_DATE': True,
+            'GALLERY_USE_METADATA': True,
```
ralsina: I don't think we need an option, since it's unlikely that metadata files will randomly appear next to images. If people don't want to show image metadata, then don't put meta files there :-)

guyou: You are right. I initially added this option, thinking on conflicting processing of title. But finally, I change my mind in the same way: if there is a meta file, it is the most important source of information.

I will remove that.

ralsina:
@guyou need a hand with anything? Want me to take over and push your work? What you have done so far is awesome, I'd love to have in Nikola!

guyou:
I would appreciate to test it on a more complex "gallerie" theme, but spare time is missing. If you consider the current state meets all you requirements, so please, pull it on main branch.

ralsina:
@guyou awesome, I will try to give this a push this weekend and see how it goes!

ralsina:
I am taking over this branch and superseding the PR with a new one. Thanks for the work @guyou and sorry it took me so long to get to it.

felixfontein:
The story continues in #2837.

PR #2442, opened 2016-08-15
New option to skip DocTitle transform in docutils (Fix #2382)

### LINKED ISSUE #2382: ReST plugin strips single title
I've noticed that if I have a .rst post which has one header, that one will be removed from the output. So if I have two pages:

Page 1:

```
Test
====

This is a test.
```

Page 2:

```
Test
====

This is a test.

Test2
=====

This is another test.
```

Then I would expect that Page 1 looks like Page 2 with the second section removed, but it ends up being just "This is a test." for Page 1 with the section header removed. I don't know why this is the default behavior, but I noticed it can be changed by modifying the ReST plugin by replacing

``` .py
return pub.writer.parts['docinfo'] + pub.writer.parts['fragment'], pub.document.reporter.max_level, pub.settings.record_dependencies
```

with

``` .py
return pub.writer.parts['html_title'] + pub.writer.parts['docinfo'] + pub.writer.parts['fragment'], pub.document.reporter.max_level, pub.settings.record_dependencies
```

at the end of `rst2html`.

I guess changing the default is not a good idea, but would it be acceptable to add an option to force inclusion of the HTML title?

ralsina:
Add a new option, defaulting to False, to skip the DocTitle transform in docutils, which has a very unexpected behaviour for most people.

### DIFF
--- a/nikola/plugins/compile/rest/__init__.py
+++ b/nikola/plugins/compile/rest/__init__.py
@@ -81,7 +81,8 @@ class CompileRest(PageCompiler):
         }
 
         output, error_level, deps = rst2html(
-            data, settings_overrides=settings_overrides, logger=self.logger, source_path=source_path, l_add_ln=add_ln, transforms=self.site.rst_transforms)
+            data, settings_overrides=settings_overrides, logger=self.logger, source_path=source_path, l_add_ln=add_ln, transforms=self.site.rst_transforms,
+            no_title_transform=self.site.config.get('NO_DOCUTILS_TITLE_TRANSFORM', False))
         if not isinstance(output, unicode_str):
             # To prevent some weird bugs here or there.
             # Original issue: empty files.  `output` became a bytestring.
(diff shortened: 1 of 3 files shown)


ralsina:
Looks like I somehow broke python 2.7 in some random place and I have no idea why. Weeeee

nikola/plugins/compile/rest/__init__.py
```diff
@@ -177,11 +178,15 @@ class NikolaReader(docutils.readers.standalone.Reader):
     def __init__(self, *args, **kwargs):
         """Initialize the reader."""
         self.transforms = kwargs.pop('transforms', [])
-        docutils.readers.standalone.Reader.__init__(self, *args, **kwargs)
+        self.no_title_transform = kwargs.pop('no_title_transform', False)
+        super(NikolaReader, self).__init__(*args, **kwargs)
```
felixfontein: I think you forgot `self` as the first argument to the call to `super`'s `__init__`.

PR #2443, opened 2016-08-15
Initial implementantion of deps for shortcodes

ralsina:
Initial attempt at supporting file dependencies from shortcodes.

- [x] Added optional with_dependencies arg for apply_shortcodes, which makes it return (output, dependencies)
- [x] Added dependency tracking for templated shortcodes
- [x] Track template dependencies usint the template plugin
- [x] Changed the reSt compiler to support shortcodes with dependencies
- [x] Changed all included compilers to support shortcodes with dependencies
- [ ] ~~Change all compiler plugins to support shortcodes + dependencies~~ will be handled in a separate branch
- [x] Change docs to describe shortcode dep tracking
- [x] Change shortcode plugin class to help with dependencies

### DIFF
--- a/docs/extending.txt
+++ b/docs/extending.txt
@@ -577,13 +577,25 @@ So, if you are creating a plugin that generates markup, it may be a good idea
 to register it as a shortcode in addition of to restructured text directive or
 markdown extension, thus making it available to all markup formats.
 
-To implement your own shortcodes from a plugin, you can create a plugin inheriting ``ShortcodePlugin`` and call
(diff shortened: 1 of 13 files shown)


ralsina:
@Kwpolska could you take a look? You can play with template shortcodes to see how the dependency tracking works. 

If a shortcode includes or extends another template, then changes on any of the relevant templates should trigger a rebuild. Secondary templates are loaded from templates/ or the theme, not shortcodes/ for boring and complicated reasons :-P

Kwpolska:
I won't be able to check it until next week.

-- 

ralsina:
@Kwpolska no worries, I think it's good :-)

PR #2448, opened 2016-08-18
Wordpress MarkDown conversion

### LINKED ISSUE #2261: import_wordpress don't convert to markdown
import without errors

text from XML `<content:encoded>` is copied with a newline after each line.

export: https://davidak.basecom.de/paste/?dc58ba722a84925a#fTZSvcajZOpwJv5uhFRSe7vLARmWCqhtL7497BObAto=

bildschirmfoto 2016-02-23 um 17 09 25

felixfontein:
Fixes #2261.

This added two options:
1. `--html2text` simply runs `html2text` on the (minimally preprocessed) WordPress post;
2. `--transform-to-markdown` first runs the WordPress compiler and then `html2text` on the WordPress post.

The second option is usually better as a lot of WordPress-specific things (like conversion of "..." to the corresponding Unicode character, smart quotes etc.) are then handled.

It won't work nice with code blocks, though, since the WordPress compiler compiles them to HTML which `html2text` cannot really decode back to what they should be in MarkDown. The first option doesn't work with code blocks, either, as `html2text` doesn't know about WordPress shortcodes. To fix this, it's probably best to first remove all `[code]` blocks and replace them with simple to identify and unique strings, then do the conversion, and finally put the code block contents back in.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,6 +7,8 @@ Features
 * New NO_DOCUTILS_TITLE_TRANSFORM (Issue #2382)
 * Update options of chart directive to Pygal 2.2.3
 * Pass global context to template shortcodes (Issue #2424)
+* Added new options --html2text and --transform-to-markdown
+  to WordPress importer (Issue #2261)
 
 Bugfixes
 --------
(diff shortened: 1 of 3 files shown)


ralsina:
LGTM

PR #2454, opened 2016-08-19
listing: Guess the lexer from its contents if the file name is unknown to pygments.

xuhdev:


ralsina:
I am slightly worried about this one because, well, what happens if pygments guesses wrong? TextLexer at least is never actively broken :-)

Anyway, merging, we can always iterate!

ralsina:
Could you add a one-liner to the features section in CHANGES.txt and add yourself to authors if you are not there? Thanks!

xuhdev:
Thanks, done! If we eventually want to remove the guess part, maybe the best thing is to allow users to manually select a lexer.

ralsina:
Yeah, we can always improve stuff. Thanks for the code!

PR #2462, opened 2016-08-20
Add data metadata (Fix #2450)

### LINKED ISSUE #2450: Data support
While trying to do a sample ite for [Stay Static](http://staystatic.github.io/) I noticed Jekyll has a rather nice feature. If you put YAML or JSON files in ``data/`` they are added to the equivalent of our ``GLOBAL_CONTEXT``.

I worked around not having that by just loading the file from conf.py but I think it would be a neat and low-cost addition.

Alternatively, or in addition to this, a ``data`` field in a post metadata could load a data set from a file. Using the new ``template`` shortcut, then you can use the template engine to iterate over that data! That could be interesting, I think :-)

ralsina:
This adds support for a ``data`` field in metadata, which loads external data files into a post.

Also, it exposes the post object to shortcodes, so you can do something like this:

A ``data.json`` file:

```
{'links': [
   ['http://foo', 'Foo site'],
   ['http://bar', 'Bar site']
]}

```

And a post like this:

```
.. data:: data.json

{{% template %}}
<ul>
% for url, description in post.data('links'):
    <li><a href=${url}>${description}
%endfor
</ul>
{{% /template %}}
```

And voilá :-)

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -363,6 +363,14 @@ to your configuration:
    enclosure
        Add an enclosure to this post when it's used in RSS. See `more information about enclosures <http://en.wikipedia.org/wiki/RSS_enclosure>`__
 
+   data
+       Path to an external data file, relative to ``conf.py``. Based on it's extension it will
+       be interpreted as JSON or YAML and be available as ``post.data``. The data **must**
+       decode into a python dictionary, and it's keys are available for templates as ``post.data('key')``.
+
+       Translated posts can have different values for this field, and the correct one will be
+       used.
+
 .. note:: The Two-File Format
 
    Nikola originally used a separate ``.meta`` file. That will still work!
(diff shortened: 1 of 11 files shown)


docs/manual.txt
```diff
@@ -363,6 +363,14 @@ to your configuration:
    enclosure
        Add an enclosure to this post when it's used in RSS. See `more information about enclosures <http://en.wikipedia.org/wiki/RSS_enclosure>`__
 
+   data
+       Path to an external data file, relative to ``conf.py``. Based on it's extension it will
+       be interpreted as JSON or YAML and be available as ``post.data``. The data **must**
+       decode into a python dictionary, and it's keys are available for templates as ``post.data('key')``.
+
+       Translated posts can have different values for this field, and the correct one will be
+       used.
```
Kwpolska: Perhaps add a word about shortcodes here?

Kwpolska:
LGTM

docs/manual.txt
ralsina: Good idea, I mentioned this the other way around, but these 2 features go well together.

Issue #2068, opened 2015-09-11
RSS for sections


Kwpolska:

ralsina:
So, it looks like this should be pretty straightforward as part of the ``indexes`` plugin, giving it a shot.

ralsina:
Looking a bit harder, we are already generating atom feeds if generate_atom is true... so, is this actually useful?

ralsina:
Anyway: https://github.com/getnikola/nikola/pull/2463

ralsina:
Done.

PR #2463, opened 2016-08-20
Rss for sections (#2068)

ralsina:
Ouch, branched out of the wrong place. Let's not merge this without #2462 first, I don't want to have to split diffs.

~~Also, this has path conflicts on multilingual sites for some reason I still have not debugged.~~

### DIFF
--- a/nikola/plugins/task/indexes.py
+++ b/nikola/plugins/task/indexes.py
@@ -341,6 +341,6 @@ class Indexes(Task):
 
         Example:
 
-        link://section_index_atom/cars => /cars/rss.xml
+        link://section_index_rss/cars => /cars/rss.xml
         """
         return self.index_section_path(name, lang, is_rss=True)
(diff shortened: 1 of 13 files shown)


nikola/plugins/task/indexes.py
```diff
@@ -296,3 +335,12 @@ def index_section_atom_path(self, name, lang):
         link://section_index_atom/cars => /cars/index.atom
         """
         return self.index_section_path(name, lang, is_feed=True)
+
+    def index_section_rss_path(self, name, lang):
+        """Link to the RSS feed for a section.
+
+        Example:
+
+        link://section_index_atom/cars => /cars/rss.xml
```
Kwpolska: That’s supposed to be `section_index_rss`, right?

ralsina: Yep.

PR #2470, opened 2016-08-24
Don’t store timestamp in github_deploy

Kwpolska:
Rationale: the Git repo will *never* be clean, because the state file
will be created after we push everything.

Moving it before we commit to the source branch means we’d store
timestamp of last deployment attempt.


ralsina:
Ok, I agree. Too bad there is no solution for this.

PR #2471, opened 2016-08-24
Disable <base> tag by default

Kwpolska:
This disables the `<base>` tag, as discussed on the mailing list recently. Rationale in `conf.py.in`.

I also disabled it for existing sites in `nikola.py` but we could undo that (but why? The tag is completely useless for 99% of people; show me a major website using it!)


ralsina:
LGTM

PR #2476, opened 2016-08-26
Fix #2332 -- fix Windows crash with calendar locales

### LINKED ISSUE #2332: When building a site with multiple date_formats, Nikola sometimes crashes and others it messes up the formats
My site source: <https://github.com/sukiletxe/sukiletxe.github.io/tree/src>
Some more data to track down the issue:
* IRC log: <https://irclogs.getnikola.com/logs/2016-05-03/>
* Traceback: <http://pastebin.com/WJmyCjT7>
* Weird things happen when setting locale: <http://pastebin.com/DsJ4eED8> (generated using <https://gist.github.com/Kwpolska/3cf154f5fc7444660a9c26a2e8adee7c>

Note: Is it necessary to use calendar.different_locale? This may be causing the issue...

Kwpolska:
Turns out the `calendar` module still breaks with set/get/set.

This is #2332. cc @ralsina @sukiletxe

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -18,6 +18,7 @@ Features
 Bugfixes
 --------
 
+* Fix Windows crash with calendar locales (Issue #2332)
 * Remove the (useless) ``<base>`` tag by default; change conf.py to
   opt in (Issue #2471)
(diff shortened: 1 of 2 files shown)


ralsina:
FTS, LGTM

Issue #2477, opened 2016-08-27
Global data support


ralsina:
Jekyll-style: load data files from data/ and put them in global context, it's probably a 10-line patch

tritium21:
Data should be namespaced in the global context.  I wouldn't want a misspelled key in a datafile to b0rk my entire site (very likely without an error message for feedback)

ralsina:
Yes, it will be in GLOBAL_CONTEXT[data][nameofthefilewithoutextension]

PR #2478, opened 2016-08-27
global data (Issue #2477)

ralsina:
Sure, it could use some error checking, some consistency checking, some non-overlapping checking, but it's the important bit.

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -956,6 +956,20 @@ In that case, the template engine used will be your theme's and the arguments yo
 as well as the global context from your ``conf.py``, are available to the template you
 are creating.
 
+.. sidebar:: The Global Context
+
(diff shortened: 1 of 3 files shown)


docs/manual.txt
```diff
@@ -956,6 +956,20 @@ In that case, the template engine used will be your theme's and the arguments yo
 as well as the global context from your ``conf.py``, are available to the template you
 are creating.
 
+.. sidebar:: The Global Context
```
Kwpolska: Should be a full section. We should also have some information about `.. data:` in a more discoverable place than the large „other metadata fields” sidebar.

Kwpolska:
:+1: from me

docs/manual.txt
ralsina: I'll do a separate branch with those. The "other metadata" should not be a sidebar anymore, either.

PR #2480, opened 2016-08-28
Fix #2472

### LINKED ISSUE #2472: reST listings should also have source links
This will make it easy to download listings displayed in posts.

ralsina:
This should not land before #2474 right?

### DIFF
--- a/nikola/plugins/task/listings.py
+++ b/nikola/plugins/task/listings.py
@@ -54,7 +54,7 @@ class Listings(Task):
     def set_site(self, site):
         """Set Nikola site."""
         site.register_path_handler('listing', self.listing_path)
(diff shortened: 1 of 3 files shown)


nikola/plugins/task/listings.py
```diff
@@ -54,6 +54,7 @@ def register_output_name(self, input_folder, rel_name, rel_output_name):
     def set_site(self, site):
         """Set Nikola site."""
         site.register_path_handler('listing', self.listing_path)
+        site.register_path_handler('listing-source', self.listing_source_path)
```
Kwpolska: Should be `listing_source` with an underscore for consistency.

ralsina: makes sense, fixing.

Kwpolska:
👍

PR #2497, opened 2016-09-11
update KaTeX version and introduce KATEX_AUTO_RENDER label

fehiepsi:
The new KaTeX version 0.6.0 supports aligned math display, which is useful. In addition, this pull request introduces the new KATEX_AUTO_RENDER label, which is useful for people who wants to create posts by using Jupyter Notebook. In particular, it will support the inline math $.$, which is the default math delimiter in Jupyter Notebook. More on KaTeX auto render can be seen from this [link](https://github.com/Khan/KaTeX/tree/master/contrib/auto-render).

This pull request will add the following code to the html source file of the post.
```javascript
<script src="https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.6.0/contrib/auto-render.min.js"></script><script>
                renderMathInElement(document.body,
                {
                    delimiters: [
                        {left: "$$", right: "$$", display: true},
                        {left: "\\[", right: "\\]", display: true},
                        {left: "$", right: "$", display: false},
                        {left: "\\(", right: "\\)", display: false}
                    ]
                }
                );
</script>

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -31,6 +31,7 @@
 * `Dhruv Baldawa <https://github.com/dhruvbaldawa>`_
 * `Dirk Engling <https://github.com/erdgeist>`_
 * `Dmitry Verkhoturov <https://github.com/paskal>`_
+* `Du Phan <https://github.com/fehiepsi>`_
 * `Duncan Lock <https://github.com/dflock>`_
 * `Edinei Cavalcanti <https://github.com/neiesc>`_
 * `Eduardo Schettino <https://github.com/schettino72>`_
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Bugfixes
 Features
 --------
 
+* Introduce the new KATEX_AUTO_RENDER label
+* Update KaTeX version to 0.6.0: support aligned math display
 * Better support for a tree of files in ``data/``
 
 New in v7.8.0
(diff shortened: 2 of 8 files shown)


Kwpolska:
Add an entry to CHANGES.txt and AUTHORS.txt.

Also, please reword the conf.py part about this setting. It’s not necessary for Jupyter Notebooks, they can deal with `\(…\)` too.

fehiepsi:
As far as I see, Jupyter Notebook just supports auto render markdown shell for delimiters `$.$` and `$$.$$`. My test agrees with the [notebook document](http://jupyter-notebook.readthedocs.io/en/latest/notebook.html). If by some way, we can use other delimiters `\(.\)` in Jupyter Notebook, then this pull request is not necessary (because it just introduced the buggy delimiter `$.$`, which is not supported by Nikola team as I understand).

Kwpolska:
In Markdown, you need to use `\\(…\\)` and it works fine.

fehiepsi:
I see. Thank you very much! I will change the `conf` files in case you still see that this new label is useful.

Kwpolska:
Please do, and please add this to CHANGES.txt and yourself to AUTHORS.txt.

fehiepsi:
I have made some changes according to your suggestion. Thanks!

Kwpolska:
Merging, thanks for your contribution!

PS. it’s called a setting. I’ll fix this in the changelog.

PR #2498, opened 2016-09-15
Try to fix #2496

### LINKED ISSUE #2496: PostScanner plugins have problems due to compilers being disabled if not used by default post scanning plugin
Nikola only enables compilers which are used by posts found by the standard post scanning plugin (i.e. posts mentioned in `POSTS` and `STORIES`) despite them being mentioned in `COMPILERS`. So if another post scanning plugin tries to add posts which use compilers which appear in `COMPILERS`, but do not appear in `POSTS` and `STORIES`, Nikola won't compile these posts and stop with an error.

The main problem are the following lines in `nikola/nikola.py`:

``` .py
        # Avoid redundant compilers
        # Remove compilers that match nothing in POSTS/PAGES
        # And put them in "bad compilers"
        pp_exts = set([os.path.splitext(x[0])[1] for x in self.config['post_pages']])
        self.config['COMPILERS'] = {}
        self.disabled_compilers = {}
        self.bad_compilers = set([])
        for k, v in compilers.items():
            if pp_exts.intersection(v):
                self.config['COMPILERS'][k] = sorted(list(v))
            else:
                self.bad_compilers.add(k)
```

If `pp_exts.intersection(v)` is replaced by `True`, this problem goes away.

felixfontein:
First try to fix #2496. Doesn't work, since it needs `PostScanner` plugins loaded to fill `bad_compilers`, which is needed during plugin loading.

I have no idea how to fix this (without adding a lot of extra logic). Simplest resolution might be to get rid of `bad_compiler` optimization?

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -867,6 +867,25 @@ class Nikola(object):
             self.tzinfo = dateutil.tz.gettz()
         self.config['__tzinfo__'] = self.tzinfo
 
+        # Get search path for themes
+        self.themes_dirs = ['themes'] + self.config['EXTRA_THEMES_DIRS']
+
+        # Determine which compilers are needed and sets global context
+        self._init_compilers_and_global_context()
+
+        # Set persistent state facility
+        self.state = Persistor('state_data.json')
+
+        # Set cache facility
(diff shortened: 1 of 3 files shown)


felixfontein:
I tried to look into this again, but I still don't know how to fix this without playing a lot more with `yapsy` internals.

PR #2501, opened 2016-09-19
Fixing dependency generation for jinja templating engine.

felixfontein:
Jinja template dependencies sometimes miss the main template itself, depending on whether the dependencies where stored in the internal cache in `JinjaTemplates.get_string_deps` or in `JinjaTemplates.template_deps`.

### DIFF
--- a/nikola/plugins/template/jinja.py
+++ b/nikola/plugins/template/jinja.py
@@ -112,7 +112,7 @@ class JinjaTemplates(TemplateSystem):
             filename = self.lookup.loader.get_source(self.lookup, dep_name)[1]
             deps.add(filename)
             sub_deps = self.get_deps(filename)
-            self.dependency_cache[dep_name] = sub_deps
+            self.dependency_cache[dep_name] = [filename] + sub_deps
             deps |= set(sub_deps)
         return list(deps)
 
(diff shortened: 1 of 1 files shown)


nikola/plugins/template/jinja.py
```diff
@@ -112,7 +112,7 @@ def get_string_deps(self, text):
             filename = self.lookup.loader.get_source(self.lookup, dep_name)[1]
             deps.add(filename)
             sub_deps = self.get_deps(filename)
-            self.dependency_cache[dep_name] = sub_deps
+            self.dependency_cache[dep_name] = [filename] + sub_deps
```
Kwpolska: Are you sure this is the place? We return `deps` from this function, and it’s not taken from the cache. Perhaps add `[filename]` a line earlier?

felixfontein: In that case, the line `deps.add(filename)` can go as well.

Kwpolska: Would be simpler to just modify `sub_deps` right when we define it, wouldn’t it?

felixfontein: Check the current commit, is it what you want?

Kwpolska: Perfect, thanks. Approving and merging.

felixfontein:
Thanks for merging!

PR #2512, opened 2016-09-25
Fix #2504 -- add nikola theme --list-installed

### LINKED ISSUE #2504: Missing --list--installed flag from nikola theme command
For `nikola plugin`, we have `--list-installed` flag, but not so for `nikola theme`.

Is it possible to implement this?

Cheers.

Kwpolska:
This is #2504.

cc @ralsina, @stefanos82.

### DIFF
--- a/nikola/plugins/command/theme.py
+++ b/nikola/plugins/command/theme.py
@@ -283,8 +283,9 @@ class CommandTheme(Command):
         themes_dirs = self.site.themes_dirs + [resource_filename('nikola', os.path.join('data', 'themes'))]
         for tdir in themes_dirs:
             themes += [(i, os.path.join(tdir, i)) for i in os.listdir(tdir)]
-        for theme in sorted(set(themes)):
-            print("{0} at {1}".format(*theme))
+        for tname, tpath in sorted(set(themes)):
+            if os.path.isdir(tpath):
+                print("{0} at {1}".format(tname, tpath))
 
     def copy_template(self, template):
         """Copy the named template file from the parent to a local theme or to templates/."""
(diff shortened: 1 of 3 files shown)


nikola/plugins/command/theme.py
```diff
@@ -263,6 +275,17 @@ def list_available(self, url):
...
         return 0
 
+    def list_installed(self):
+        """List all installed themes."""
+        print("Installed Themes:")
+        print("-----------------")
+        themes = []
+        themes_dirs = self.site.themes_dirs + [resource_filename('nikola', os.path.join('data', 'themes'))]
+        for tdir in themes_dirs:
+            themes += [(i, os.path.join(tdir, i)) for i in os.listdir(tdir)]
```
ralsina: Shouldn't you filter themes by isdir() just in case there is some random file in themes/ or similar?

Kwpolska: Fixed.

stefanos82:
That's very helpful and so nice of you @Kwpolska and @ralsina for implementing this in such a short period of time.

Another kind suggestion: can we do something about preventing a user from even attempting to install `bootstrap3` or other default templates that come pre-configured with Nikola installation?

Even though it throws an error, it should have prevented me from attempting to do so by simply invalidating the theme name.

``` bash
nikola theme --install=bootstrap3
[2016-09-25T18:35:12Z] INFO: requests.packages.urllib3.connectionpool: Starting new HTTPS connection (1): themes.getnikola.com
[2016-09-25T18:35:18Z] ERROR: theme: Theme 'bootstrap3' is already installed in /home/stefanos/.virtualenvs/mynewenv/lib/python3.5/site-packages/nikola/data/themes/bootstrap3
```

Kwpolska:
The theme doesn’t get overwritten (in fact, you can’t download base and bootstrap3 from the site). We could check the built-in theme list if we really cared, but this is good enough IMO.

stefanos82:
That should do too, no worries.

PR #2517, opened 2016-10-02
Allowing to override url_type in render_template and rewrite_links methods.

felixfontein:
This allows a plugin to render a page which can be put at various URLs even if `URL_TYPE` in the site's config is set to `rel_path`. This can for example be used for creating 404 pages.


felixfontein:
As soon as I've implemented it, yes ;) I'm just in the process of checking what I need for it.

PR #2519, opened 2016-10-02
Refactored generic_page_renderer and generic_post_list_renderer.

felixfontein:
Preliminary work for @Kwpolska's suggestion in getnikola/plugins#175.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1986,7 +1986,7 @@ class Nikola(object):
             sys.exit(1)
         signal('scanned').send(self)
 
-    def generic_renderer(self, lang, output_name, template_name, filters, file_deps=None, uptodate_deps=None, pre_context=None, post_context=None, context_deps_remove=None, post_deps_dict=None, url_type=None):
+    def generic_renderer(self, lang, output_name, template_name, filters, file_deps=None, uptodate_deps=None, context=None, context_deps_remove=None, post_deps_dict=None, url_type=None):
         """Helper function for rendering pages and post lists and other related pages.
 
         lang is the current language.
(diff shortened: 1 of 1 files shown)


nikola/nikola.py
```diff
@@ -1986,38 +1986,27 @@ def scan_posts(self, really=False, ignore_quit=False, quiet=False):
             sys.exit(1)
         signal('scanned').send(self)
 
-    def generic_page_renderer(self, lang, post, filters, context=None):
-        """Render post fragments to final HTML pages."""
+    def generic_renderer(self, lang, output_name, template_name, filters, deps=[], uptodate_deps=[], pre_context=None, post_context=None, context_deps_remove=None, post_deps_dict=None, url_type=None):
```
Kwpolska: You can’t use `deps=[], uptodate_deps=[]`. You should also document what arguments mean.

nikola/nikola.py
```diff
@@ -1986,38 +1986,27 @@ def scan_posts(self, really=False, ignore_quit=False, quiet=False):
...
+        if post_context:
+            context.update(post_context)
+
         deps_dict = copy(context)
-        deps_dict.pop('post')
-        if post.prev_post:
-            deps_dict['PREV_LINK'] = [post.prev_post.permalink(lang)]
-        if post.next_post:
-            deps_dict['NEXT_LINK'] = [post.next_post.permalink(lang)]
+        if context_deps_remove:
```
Kwpolska: Could use a comment to explain this.

felixfontein: Do you mean what `context_deps_remove` does? That should now be covered in the argument documentation. Please tell me if that's enough.

nikola/nikola.py
```diff
@@ -1986,38 +1986,40 @@ def scan_posts(self, really=False, ignore_quit=False, quiet=False):
             sys.exit(1)
         signal('scanned').send(self)
 
-    def generic_page_renderer(self, lang, post, filters, context=None):
-        """Render post fragments to final HTML pages."""
+    def generic_renderer(self, lang, output_name, template_name, filters, file_deps=[], uptodate_deps=[], pre_context=None, post_context=None, context_deps_remove=None, post_deps_dict=None, url_type=None):
```
Kwpolska: I meant, you can’t provide an empty list as a default argument to file_deps and uptodate_deps.

nikola/nikola.py
Kwpolska: It’s enough.

nikola/nikola.py
felixfontein: Ah, I forgot about that trap... I'll fix it.

felixfontein:
One note: this will cause all pages rendered with `generic_post_list_renderer` to be rendered again as their `deps_dict` now also contains `OUTPUT_FOLDER` and `TRANSLATIONS`. (It didn't have that before.) Should I move them so they're only added for `generic_page_renderer`, or is it ok if `generic_post_list_renderer` also adds them? (For me, the latter seems more natural and that's why I implemented it this way.)

felixfontein:
Also, another thing: the difference between `pre_context` and `post_context` is now only that only `post_context` allows to override `lang`, while `lang` in `pre_context` will be overridden. Assuming that you never want to override `lang`, how about merging them to one variable?

Kwpolska:
1. Wouldn’t that dependency come from the GLOBAL_CONTEXT anyway? Or do you mean something else?
2. :+1: for that merge, overriding `lang` sounds like a very stupid thing to do — let’s not support it.

felixfontein:
1. So far `GLOBAL_CONTEXT` was merged in by the callers of `generic_post_list_renderer` and `generic_page_renderer`, but of course not everyone has to do that.

2. Ok, I'll fix that then.

nikola/nikola.py
```diff
@@ -1995,22 +1995,19 @@ def generic_renderer(self, lang, output_name, template_name, filters, file_deps=
...
         """
         utils.LocaleBorg().set_locale(lang)
 
-        file_deps = [] if file_deps is None else file_deps
+        file_deps = copy(file_deps) if file_deps else []
         file_deps += self.template_system.template_deps(template_name)
         file_deps = sorted(list(filter(None, file_deps)))
 
-        context = copy(pre_context) if pre_context else {}
+        context = copy(context) if context else {}
```
Kwpolska: Do we need a second copy here, just to override `lang`?

felixfontein: Yes, in case someone calls this function with the same context for different languages in a row. That person will be quite surprised if they all end up with the same `lang` value in their context.

felixfontein:
Thanks for merging. I'll now update the plugin accordingly.

PR #2523, opened 2016-10-05
Set markdown output_format to html5 instead of the default xhtml1

### LINKED ISSUE #2522: Nikola generates invalid html5 when markdown footnote extension is used
The [default output type](http://pythonhosted.org/Markdown/reference.html#markdown) of the python markdown library is xhtml1. The 4 templates that ship with Nikola have <!DOCTYPE html> which defines them as html5, so I'm assuming that we're intending to generate html5.

When the footnote markdown extension is used, it generates invalid html5 according to the w3c validator.

`<a class="footnote-ref" href="..." rev="footnote">...</a>`

(rev="footnote" is valid html4, but not html5)

The markdown library indicates that this is invalid html5 (https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/footnotes.py#L149) so we can trigger the correct behaviour by setting the output_format.

Given the markdown library does not make much use of the output_format variable, I don't think this is likely to materially change the output for many people at all - https://github.com/waylan/Python-Markdown/search?utf8=%E2%9C%93&q=output_format)

edwinsteele:
Fixes #2522 

Given the markdown library does not make much use of the output_format variable, I don't think this is likely to materially change the output for many people at all - https://github.com/waylan/Python-Markdown/search?utf8=%E2%9C%93&q=output_format)

### DIFF
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -78,7 +78,7 @@ class CompileMarkdown(PageCompiler):
                 data = in_file.read()
             if not is_two_file:
                 _, data = self.split_metadata(data)
-            output = markdown(data, self.extensions)
+            output = markdown(data, self.extensions, output_format="html5")
             output, shortcode_deps = self.site.apply_shortcodes(output, filename=source, with_dependencies=True, extra_context=dict(post=post))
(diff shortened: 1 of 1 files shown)


Kwpolska:
We’ll go with this, thanks!

PR #2532, opened 2016-10-14
Fix #2531 — pass post and lang to compilers

### LINKED ISSUE #2531: Pass post object to compile_html
Instead of using a hack with paths to save dependencies (and potentially do other things), we could just pass a `post` argument to `compile_html`.

Kwpolska:
This is #2531. There isn’t anything else we could post to this function, so hopefully we won’t have to do `real_compile` next time.

cc @felixfontein, reviews welcome.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,14 @@
+New in master
+=============
+
+Bugfixes
+--------
+
+Features
+--------
+
+* Pass ``post`` object and ``lang`` to post compilers (Issue #2531)
+
 New in v7.8.1
 =============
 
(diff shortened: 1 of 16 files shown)


Kwpolska:
This is so stupid. Apparently, `is_one_file` can make things break if set incorrectly (perhaps due to metadata splitting). I’ll fix it and merge.

felixfontein:
Why not fix the test instead of (again) having the compilers `compile` signature disagree with the signature in `PageCompiler`?

Kwpolska:
Or we can make the PageCompiler signature match.

And the problem is not really in the test. `is_two_file` set to False means some extra manipulation which should not be the default.

PR #2544, opened 2016-10-22
Allow overriding URL_TYPE via meta

Kwpolska:
Useful for HTML fragments inserted via JavaScript that contain links.

Use case: nikola-plugins-v2 and its side menu that can be changed
without rebuilding all pages.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Bugfixes
 Features
 --------
 
+* Allow posts to set custom ``URL_TYPE`` by using the ``url_type``
+  meta tag (useful for HTML fragments inserted using JavaScript)
 * Plugins can depend on other plugins being installed (Issue #2533)
 * The destination folder in ``POSTS`` and ``PAGES`` can now be
   translated (Issue #2116)
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2085,7 +2085,8 @@ class Nikola(object):
                                     uptodate_deps=uptodate_deps,
                                     context=context,
                                     context_deps_remove=['post'],
-                                    post_deps_dict=deps_dict)
+                                    post_deps_dict=deps_dict,
+                                    url_type=post.url_type)
 
     def generic_post_list_renderer(self, lang, posts, output_name, template_name, filters, extra_context):
         """Render pages with lists of posts."""
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -249,6 +249,10 @@ class Post(object):
         self.use_in_feeds = use_in_feeds and not is_draft and not is_private \
             and not self.publish_later
 
+        # Allow overriding URL_TYPE via meta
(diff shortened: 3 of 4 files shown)


felixfontein:
I'm wondering how you're using this to create HTML fragments. To me it seems hard to get Nikola to not create whole HTML pages but only fragments.

Wouldn't it make more sense to write a plugin to create such HTML fragments?

Kwpolska:
I’m creating fragments for some value of fragments. Our current rendering pipeline (lxml) will add a DOCTYPE, `<html>` and `<body>`. This is good enough, because jQuery’s `$.load('/sidemenu.html div#sidemenu')` will take care of this.

And if I really wanted fragments, I could make a micro-plugin to copy /cache/pages/sidemenu.html to /output/real_sidemenu.html as-is.

felixfontein:
Wouldn't it still be nice if it would be possible to instruct `generic_renderer` and `render_template` to allow outputting a real fragment? This would also allow to do a simple text substitution to do the work on the server, instead of using JavaScript on the client.

Kwpolska:
That’s going to be slightly tricky, but we could skip the lxml part. This would have side effects (like the possibility of broken HTML in output or URL_TYPE effectively being completely ineffective, and I don’t want any new *not server root* bugs)

Unless we could tell lxml to output fragments, if that’s possible.

felixfontein:
From here: http://stackoverflow.com/questions/3443831/python-and-elementtree-return-inner-xml-excluding-parent-element it looks it can be done like this:
``` .py
parser = lxml.html.HTMLParser(remove_blank_text=True)
doc = lxml.html.fragment_fromstring(data, parser)

doc.rewrite_links(lambda url: rewrite(url))
for obj in list(doc.xpath('(//img|//source)')):
    if 'srcset' in obj.attrib:
        ...

data = (body.text or '').encode('utf-8') + ''.encode('utf-8').join([lxml.html.tostring(child, encoding='utf-8', method='html') for child in body.iterchildren()])
```
So essentially parsing the document is changed a little (`fragment_fromstring` instead of `document_fromstring`), the `srcset` fix is changed a little (removal of `*` in `'(*//img|*//source)'`), and recombining the doc into a HTML fragment changes (call `tostring` on children, joining the result and prepending `doc.text`). (If we'd just call `tostring` on the whole of `doc`, a `<div></div>` is added to stand-alone text.)

Kwpolska:
That `<div></div>` sounds like a good compromise to me if that’s the only thing that happens.

felixfontein:
That was the only thing I noticed in a very limited minimal test. So it could be that worse things happen :)

felixfontein:
It actually seems to always add a `<div></div>` around the *whole* output.

felixfontein:
(So `<div>1</div>` gets converted to `<div><div>1</div></div>` with a newline before the final `</div>`.)

PR #2545, opened 2016-10-23
Passing url_type into template context.

felixfontein:
Also modifies base theme to use that when calling `url_replacer`. This allows to fix the Nikola core part of getnikola/plugins#176.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1306,7 +1306,7 @@ class Nikola(object):
         for k in self._GLOBAL_CONTEXT_TRANSLATABLE:
             local_context[k] = local_context[k](local_context['lang'])
         local_context['is_rtl'] = local_context['lang'] in LEGAL_VALUES['RTL_LANGUAGES']
-        local_context['url_type'] = self.site.config['URL_TYPE'] if url_type is None else url_type
+        local_context['url_type'] = self.config['URL_TYPE'] if url_type is None else url_type
         # string, arguments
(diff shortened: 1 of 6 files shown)


nikola/nikola.py
```diff
@@ -1306,6 +1306,7 @@ def render_template(self, template_name, output_name, context, url_type=None):
         for k in self._GLOBAL_CONTEXT_TRANSLATABLE:
             local_context[k] = local_context[k](local_context['lang'])
         local_context['is_rtl'] = local_context['lang'] in LEGAL_VALUES['RTL_LANGUAGES']
+        local_context['url_type'] = self.config['URL_TYPE'] if url_type is None else url_type
```
Kwpolska: Now that we’re doing all this, perhaps we could pass `url_type` into this function from callers?

felixfontein: Why should callers pass the argument twice (both into context and to `render_template`)? After all this value is used in the default base theme in essentially every template.

felixfontein: I couldn't find a reason why anyone calling this function does NOT want to have `url_type` in context (except that it causes all pages to recompile, that is).

Kwpolska: Ah, `generic_` don’t use `url_type` themselves. Feel free to merge.

felixfontein: Ok, thanks, I'll merge then.

PR #2546, opened 2016-10-23
Allowing to write HTML fragments instead of only whole documents.

felixfontein:
Test implementation for the dicussion at the bottom of #2544.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -16,6 +16,8 @@ Features
   translated (Issue #2116)
 * Pass ``post`` object and ``lang`` to post compilers (Issue #2531)
 * Pass ``url_type`` into template's context.
+* ``render_template`` and ``generic_renderer`` can now create HTML
+  fragments.
 
 New in v7.8.1
 =============
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Bugfixes
 Features
 --------
 
+* ``render_template`` and ``generic_renderer`` can now create HTML
+  fragments.
 * Allow posts to set custom ``URL_TYPE`` by using the ``url_type``
   meta tag (useful for HTML fragments inserted using JavaScript)
 * Plugins can depend on other plugins being installed (Issue #2533)
@@ -16,8 +18,6 @@ Features
   translated (Issue #2116)
 * Pass ``post`` object and ``lang`` to post compilers (Issue #2531)
 * Pass ``url_type`` into template's context.
-* ``render_template`` and ``generic_renderer`` can now create HTML
-  fragments.
 
 New in v7.8.1
 =============
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1348,7 +1348,7 @@ class Nikola(object):
             # argument with_tail has default value True).
             data = (doc.text or '').encode('utf-8') + b''.join([lxml.html.tostring(child, encoding='utf-8', method='html') for child in doc.iterchildren()])
         else:
-            data = b'<!DOCTYPE html>\n' + lxml.html.tostring(doc, encoding='utf8', method='html', pretty_print=True)
+            data = lxml.html.tostring(doc, encoding='utf8', method='html', pretty_print=True, doctype='<!DOCTYPE html>')
         with open(output_name, "wb+") as post_file:
             post_file.write(data)
 
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1341,7 +1341,7 @@ class Nikola(object):
             doc = lxml.html.fragment_fromstring(data, parser)
         else:
             doc = lxml.html.document_fromstring(data, parser)
-        self.rewrite_links(doc, src, context['lang'], url_type, is_fragment=is_fragment)
+        self.rewrite_links(doc, src, context['lang'], url_type)
         if is_fragment:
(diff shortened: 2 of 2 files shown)


CHANGES.txt
```diff
@@ -16,6 +16,8 @@ Features
   translated (Issue #2116)
 * Pass ``post`` object and ``lang`` to post compilers (Issue #2531)
 * Pass ``url_type`` into template's context.
+* ``render_template`` and ``generic_renderer`` can now create HTML
```
Kwpolska: PS. new changes go at the top.

nikola/nikola.py
```diff
@@ -1334,19 +1337,25 @@ def render_template(self, template_name, output_name, context, url_type=None):
...
-        doc = lxml.html.document_fromstring(data, parser)
-        self.rewrite_links(doc, src, context['lang'], url_type)
-        data = b'<!DOCTYPE html>\n' + lxml.html.tostring(doc, encoding='utf8', method='html', pretty_print=True)
+        if is_fragment:
+            doc = lxml.html.fragment_fromstring(data, parser)
+        else:
+            doc = lxml.html.document_fromstring(data, parser)
+        self.rewrite_links(doc, src, context['lang'], url_type, is_fragment=is_fragment)
+        if is_fragment:
+            data = (doc.text or '').encode('utf-8') + ''.encode('utf-8').join([lxml.html.tostring(child, encoding='utf-8', method='html') for child in doc.iterchildren()])
```
Kwpolska: TIL lxml has awful trees.

How would the solution that adds the extra div tag look? I don’t think we would really mind one little <div> there. Or worst case scenario, if we know the `<div>` is *guaranteed* to always be there, we could slice it out.

nikola/nikola.py
```diff
@@ -1334,19 +1337,25 @@ def render_template(self, template_name, output_name, context, url_type=None):
...
 
-    def rewrite_links(self, doc, src, lang, url_type=None):
+    def rewrite_links(self, doc, src, lang, url_type=None, is_fragment=False):
         """Replace links in document to point to the right places."""
         # First let lxml replace most of them
         doc.rewrite_links(lambda dst: self.url_replacer(src, dst, lang, url_type), resolve_base_href=False)
 
         # lxml ignores srcset in img and source elements, so do that by hand
-        objs = list(doc.xpath('(*//img|*//source)'))
+        objs = list(doc.xpath('({0}//img|{0}//source)'.format('' if is_fragment else '*')))
```
Kwpolska: Please rewrite this as a 4-line `if/else` tree instead of abusing str.format. We should also find out what that asterisk changes (can non-fragment code work without it? Does it reach all `` and `<source>` elements if the asterisk is not present?

felixfontein: It doesn't reach top-level `` tags with the asterisk there. I think (I don't know the xpath syntax well enough to be sure) that the asterisk eats up a top level element, and then it looks for children (or children-of-children etc.) of that top element.

CHANGES.txt
felixfontein: Should I also move the `url_type` part, then?

nikola/nikola.py
felixfontein: The extra `<div></div>` is produced by `lxml.html.tostring(doc, encoding='utf-8', method='html')`.

I don't really like cutting stuff out via text replacements.

nikola/nikola.py
Kwpolska: What about no asterisk and full document? (Please rewrite in 4 lines anyway)

felixfontein: (The original syntax was introduced in 24a45e18af3f086ad1915d486e69ac322d5bed31.)

nikola/nikola.py
```diff
@@ -1334,9 +1337,15 @@ def render_template(self, template_name, output_name, context, url_type=None):
...
         parser = lxml.html.HTMLParser(remove_blank_text=True)
-        doc = lxml.html.document_fromstring(data, parser)
+        if is_fragment:
+            doc = lxml.html.fragment_fromstring(data, parser)
+        else:
+            doc = lxml.html.document_fromstring(data, parser)
         self.rewrite_links(doc, src, context['lang'], url_type)
-        data = b'<!DOCTYPE html>\n' + lxml.html.tostring(doc, encoding='utf8', method='html', pretty_print=True)
+        if is_fragment:
+            data = (doc.text or '').encode('utf-8') + ''.encode('utf-8').join([lxml.html.tostring(child, encoding='utf-8', method='html') for child in doc.iterchildren()])
```
Kwpolska: `doc.text` is all the text in `doc` that appears before a HTML element. Text that appears after HTML elements is taken care of by `tostring()`. (lxml is weird.)

felixfontein: I added some comments with a similar explanation, hope it makes it clearer.

ralsina:
Ok then!

felixfontein:
One more thing I had to try out after seeing that `lxml.html.tostring` has a `doctype` argument. :)

felixfontein:
The one test fails because there were some connection problems to pypi while installing doit. All other tests had no problems.

felixfontein:
Same problem as before: one test fails because of connection problems to pypi...

Kwpolska:
Those tests are pretty worthless anyways.

felixfontein:
Ok. And thanks for the approval. Anyone mind if I merge?

Kwpolska:
Go ahead.

Issue #2536, opened 2016-10-16
Let compiler specify additional targets


felixfontein:
What do you think about a change to the `posts` plugin to allow the post's compiler to specify more targets than the HTML fragment?

With this, post compilers can make sure that cache files which contain dependencies (read by `register_extra_dependencies`) is re-created even when the resulting HTML fragment is there, making the whole mechanism to write dependencies to files less hackish.

ralsina:
But why would extra targets change how dependencies work? Can you give me
an example?

felixfontein:
Assuming there's only a simple list of dependent files (such as created by ReST plugin etc.): assume someone partially deleted their `cache` folder, so that for one post the `.html` fragment is there but the `.dep` file is gone. Then if the dependency is changed and `nikola build` is run, Nikola assumes the post has no dependencies and doesn't recompile it.

Or assume the compiler generates some other files out of the post and somewhere wants to keep track which files it actually generates, so a closely related plugin can make sure that the files are really there. For that, it must be possible to enforce that the compiler is re-run when that file or the generated files mentioned in there suddenly go missing. (For example, because someone deleted `output` but not `cache`.)

felixfontein:
I forgot: if the generated file which keeps track of the dependencies is a target, doit will recompile the post when that file goes missing. So it cannot happen that some change gets lost because the dependency file got deleted (or some other generated files, or whatever else).

ralsina:
OK, I agree that .deps should be a target of the compiler plugin, and that
if a compiler produces more than one file it should mark them all as
targets.

The last part about output/ doesn't sound as good. We need to trust doit.
If we produce a file, and it's deleted, it will be rebuilt. If we need to
manually account for special cases we are doing it wrong.

felixfontein:
This was fixed in #2547.

PR #2547, opened 2016-10-25
Fixing dependencies

felixfontein:
Implementation of #2536.

The first part ensures that the `.dep` file is always written and specified as a second target of the `render_post` task, so that the `.dep` file will be created if it isn't there yet. (Otherwise changing dependencies of the post might be ignored when the `.dep` file is gone.)

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -275,6 +275,10 @@ class PageCompiler(BasePlugin):
         """Add dependency to post object to check .dep file."""
         post.add_dependency(lambda: self._read_extra_deps(post), 'fragment')
 
+    def get_extra_targets(self, post, lang, dest):
+        """Returns a list of extra targets for the render_posts task when compiling the post for the specified language."""
+        return []
+
     def compile(self, source, dest, is_two_file=True, post=None, lang=None):
         """Compile the source file into HTML and save as dest."""
         # For backwards compatibility, call `compile_html`
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -251,7 +251,7 @@ class PageCompiler(BasePlugin):
     friendly_name = ''
     demote_headers = False
     supports_onefile = True
-    use_dep_file = False  # If set to true, the .dep file is always written and added as a target
+    use_dep_file = True  # If set to false, the .dep file is never written and not automatically added as a target
     default_metadata = {
         'title': '',
         'slug': '',
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -280,6 +280,9 @@ class PageCompiler(BasePlugin):
     def register_extra_dependencies(self, post):
         """Add dependency to post object to check .dep file."""
         def create_lambda(lang):
+            # We create a lambda like this so we can pass `lang` to it, because if we didn’t
+            # add that function, `lang` would always be the last language in TRANSLATIONS.
+            # (See http://docs.python-guide.org/en/latest/writing/gotchas/#late-binding-closures)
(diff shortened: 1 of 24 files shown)


Kwpolska:
I’d rather not create empty .dep files. Are there any issues with the current approach?

felixfontein:
Yes: if the `.dep` file was removed for some reasons, and a file which should have been mentioned there was modified, the post won't be recompiled because Nikola cannot distinguish between an empty `.dep` file and the `.dep` file not being there.

felixfontein:
How about letting the compiler decide whether `.dep` must be created (and let it insert the `.dep` target manually via the new function)?

Kwpolska:
Are compilers the only things that can provide .dep files? If yes, we could do that.

felixfontein:
Well, of course any other plugin could also create `.dep` files for whatever it wants :) To my knowledge, `.dep` files are only generated by compilers. Or more precisely, nowadays they are created by the `Post` object.

A problem is that sometimes compilers also create `.dep` files with a different meaning; for example the WordPress compiler. So always writing `.dep` files is probably not a good idea at all. How about: the plugin can indicate that the `Post` object should always write the `.dep` file, and add it to the list of targets. This will then be used by the ReST plugin and which else of the standard plugins uses the `.dep` mechanism.

I'll try that out.

felixfontein:
Ah. There's actually a bug in the `.dep` system: `.dep` files are written for different languages, but only the `.dep` file for the default language is ever read.

nikola/plugin_categories.py
```diff
@@ -250,6 +250,7 @@ class PageCompiler(BasePlugin):
     friendly_name = ''
     demote_headers = False
     supports_onefile = True
+    use_dep_file = False  # If set to true, the .dep file is always written and added as a target
```
felixfontein: Well, some compilers might use `.dep` files for something else. Actually, the WordPress compiler does: it uses the `.dep` file to store a JSON-encoded file which not only contains file dependencies, but potentially other dependencies. (I added that feature when `.dep` files were completely handled by the compilers themselves.) I don't know if any other plugin not shipped with Nikola directly uses `.dep` like this as well.

felixfontein:
Hmm, there seem to be a LOT of failing stuff in the tests which are not caused by this PR. Like all the flake8 warnings. I was only able to get these warnings after upgrading flake8. Also, it seems like the baseline is (again) out-of-date, as the generated thumbnails differ.

nikola/plugin_categories.py
Kwpolska: Or perhaps we could make it so `.dep` always means file dependencies separated by newlines, and the WordPress compiler creates its custom `.wpdep` format?

And I think the default should be `True`.

nikola/plugin_categories.py
```diff
@@ -273,7 +278,18 @@ def _read_extra_deps(self, post):
 
     def register_extra_dependencies(self, post):
         """Add dependency to post object to check .dep file."""
-        post.add_dependency(lambda: self._read_extra_deps(post), 'fragment')
+        def create_lambda(lang):
```
Kwpolska: Add a comment for future readers to explain this:

    # We create a lambda like this so we can pass `lang` to it, because if we didn’t add that function, `lang` would always be the last language in TRANSLATIONS.

felixfontein: Fixed that.

felixfontein: @.dep file name: Yes, we could do that, too. I'll change it.

felixfontein:
I'll wait until the checks are done, and then I'll merge.

PR #2554, opened 2016-11-05
Fix mangled pseudolinks (example patch for #2553)

### LINKED ISSUE #2553: url_replacer() mangles bare hash links
`<a href="#">something</a>` is mangled into `<a href="index.html">something</a>`, as is `<a href="#a">something</a>` unless there exists an element with `id="a"`. This causes a lot of problems for JavaScript-powered links. I have a simple example patch that fixes this.

pydsigner:
Simple patch to fix #2553.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,6 +7,7 @@ Bugfixes
 * Make ``data`` from global context available to templated shortcodes
   as ``global_data`` (Issue #2488)
 * Don't crash if plugins is a file (Issue #2539)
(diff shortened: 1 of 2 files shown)


ralsina:
Optional before merging, can you add a line in CHANGES.txt? If you don't I'll do it.

pydsigner:
@ralsina How does that look?

ralsina:
@pydsigner awesome, merging.

pydsigner:
Thanks much!

PR #2556, opened 2016-11-06
Change Python 2 switch message to say (or 2017)

Kwpolska:
We suck at deadlines, don’t we?

(I would love to get v8 out this year, though! @ralsina, any opinions?)


ralsina:
Mainly I don't see a big reason to release v8 because... what
backwards-incompatible change is an large, actual improvement?

If there isn't one, then it's not worth it.

PR #2563, opened 2016-11-20
Adjusting to updated version of flake8.

felixfontein:
A new version of flake8 contains a new (or updated?) rule (E305) which makes Nikola fail flake8. This PR should fix it.


felixfontein:
The baseline test is (not surprisingly) failing because:
  1. I had to change one of the source files in the example site as well;
  2. thumbnails have a different binary result than in the baseline.

PR #2564, opened 2016-11-22

Kwpolska:


Kwpolska:
We can’t do that because of Python 2 compatibility.

ralsina:
Or we can just stop supporting python 2. Is as good a reason as any :-)
If someone needs python 2 nikola he can keep using 7.8.1 or whatever.

Kwpolska:
Feel free to make a PR to drop python 2.

PR #2565, opened 2016-11-23
Make thumbnail naming configurable with IMAGE_THUMBNAIL_FORMAT.

blubberdiblub:
This will make it possible to deviate from the hard-coded thumbnail naming scheme of `original_image_name.thumbnail.extension`.

Coincidentally, it also allows to force all thumbnails to a particular image format (say JPEG) by putting that extension into the template literally instead of using the `{ext}` placeholder.

The default template `'{stem}.thumbnail{ext}'` corresponds to the current hard-coded naming scheme, so it's backwards compatible.

My personal use case is getting `.jpg` thumbnails out of source `.png` images, as my source material mainly consists of paletted 4 and 16 color images, which compress extremely well as PNG images, but when scaled down as thumbnails, they gain all kinds of extra shades of colors due to interpolation (which I want to keep, this is really fine and makes the thumbnails very nice to look at despite their smaller size), so the thumbnail files get __significantly__ larger than the actual images. Considering that the primary blog article (before following the links to the original images) should be lean and mean, this is somewhat disappointing. So generating JPEG thumbnails is my way of counteracting the file size increase somewhat:

     41818 images/tfmx-tutorial-1/17-tfmx-load-song.png
     21586 output/images/tfmx-tutorial-1/17-tfmx-load-song.png
    148790 output/images/tfmx-tutorial-1/17-tfmx-load-song.thumbnail.png
     86125 output/images/tfmx-tutorial-1/17-tfmx-load-song.thumbnail.jpg

Also, I don't particularly like the `.thumbnail` string, as it's a bit too long for my taste. Making the naming configurable enables me to change that in my blog's `conf.py` later to whatever I desire.

### DIFF
--- a/nikola/plugins/task/scale_images.py
+++ b/nikola/plugins/task/scale_images.py
@@ -45,6 +45,7 @@ class ScaleImage(Task, ImageProcessor):
 
     def process_tree(self, src, dst):
         """Process all images in a src tree and put the (possibly) rescaled images in the dst folder."""
+        thumb_fmt = self.kw['image_thumbnail_format']
         ignore = set(['.svn'])
         base_len = len(src.split(os.sep))
         for root, dirs, files in os.walk(src, followlinks=True):
@@ -60,7 +61,11 @@ class ScaleImage(Task, ImageProcessor):
                     continue
                 dst_file = os.path.join(dst_dir, src_name)
                 src_file = os.path.join(root, src_name)
-                thumb_file = '.thumbnail'.join(os.path.splitext(dst_file))
+                thumb_stem, thumb_ext = os.path.splitext(src_name)
+                thumb_file = os.path.join(dst_dir, thumb_fmt.format(
+                    stem=thumb_stem,
+                    ext=thumb_ext,
+                ))
                 yield {
                     'name': dst_file,
                     'file_dep': [src_file],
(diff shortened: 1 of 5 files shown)


nikola/plugins/task/scale_images.py
```diff
@@ -45,6 +45,7 @@ def set_site(self, site):
 
     def process_tree(self, src, dst):
         """Process all images in a src tree and put the (possibly) rescaled images in the dst folder."""
+        thumb_fmt = self.kw['image_thumbnail_format']
         ignore = set(['.svn'])
```
ralsina: Not really related to this PR but WTF. Also, Thumbs.db special casing a few lines below makes little sense.

blubberdiblub:
Sure, can be argued. I preferred `{stem}` as it's unambiguous (whereas `{name}` could also refer to the whole name including the extension), but I don't have a strong opinion concerning that and am ready to change it, if desired.

Kwpolska:
You decide.

nikola/plugins/task/scale_images.py
Kwpolska: `.svn` is the real WTF. And `Thumbs.db`/`.DS_Store` special handling below. Pre-existing WTF, that is.

(You don’t *need* that variable, you could just take it out of `self.kw` on line 65; that would actually look better IMO

blubberdiblub:
Alright, there's no need to contribute to the proliferation of even more terms for the same thing (there are already like 4 or so ;) ) and there's no danger of someone getting it wrong, as far as I can tell, so I'll change it to `{name}`.

Kwpolska:
Alright, thanks for contributing to Nikola! :tada:

nikola/plugins/task/scale_images.py
blubberdiblub: Yes, I considered that, but I don't like retrieving the same item from a dictionary over and over again during each iteration, that's why I opted for retrieving it once before entering the loop.

Kwpolska:
(I added your name to AUTHORS.txt, by the way)

ralsina:
Thanks for the neat feature @blubberdiblub !!!!

PR #2568, opened 2016-11-24
Fix #2566 -- remove unusual scale_images ignores

### LINKED ISSUE #2566: scale_images has unusal ignoring setup
```py
         ignore = set(['.svn'])
```

WTF, svn? This is 2016. Also, why is this a set?

```py
                 if src_name in ('.DS_Store', 'Thumbs.db'):
```

Should probably be merged with dictionary ignores. (Not sure if we need to make this customizable?)

Kwpolska:
Rationale:

* SVN is basically dead, and the ignore isn’t necessary for modern
versions anyway. (Fixed in 2011)
* .DS_Store and Thumbs.db are not going to be considered image files
due to their extensions anyways.

cc @ralsina, @felixfontein, @blubberdiblub


ralsina:
+1

PR #2577, opened 2016-12-04
Fix #2576 — Send coverage information to Codacy

### LINKED ISSUE #2576: Pay real attention to code coverage
Code coverage measurements do exist, but we don’t really care about it. Perhaps we could make Codacy handle that for us?

Kwpolska:
This is #2576. Codacy needs a PR to analyze a commit, so here goes.


Kwpolska:
It looks pretty useful, so I’m going to merge..

PR #2578, opened 2016-12-04
Unescaping HTML in WordPress tags and categories.

### LINKED ISSUE #2557: Wordpress Importer: Tags: &amp; should be just &
I noticed that imports made through the wordpress importer plugin does not correctly handle ampersands (*&*) in tags and writes them into the new site as &amp;.

This probably applies to other HTML entities aswell.

felixfontein:
Fixes #2557.

@okin: can you test this with your older export?

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -94,7 +94,7 @@ __all__ = ('CustomEncoder', 'get_theme_path', 'get_theme_path_real', 'get_theme_
            'NikolaPygmentsHTML', 'create_redirect', 'TreeNode',
            'flatten_tree_structure', 'parse_escaped_hierarchical_category_name',
            'join_hierarchical_category_path', 'clean_before_deployment', 'indent',
-           'load_data')
+           'load_data', 'html_unescape')
 
 # Are you looking for 'generic_rss_renderer'?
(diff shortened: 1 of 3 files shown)


okin:
Worked for me with Python 3.5 and code looks sane to me aswell.

Thank you @felixfontein!

felixfontein:
Good to hear! I'd like @Kwpolska to also take a look, especially at the function in `utils`, before we merge.

nikola/utils.py
```diff
@@ -1942,3 +1942,20 @@ def load_data(path):
         return
     with io.open(path, 'r', encoding='utf8') as inf:
         return loader.load(inf)
+
+
+# http://stackoverflow.com/a/2087433
+if sys.version_info[0] == 3 and sys.version_info[1] >= 4:
```
Kwpolska: Breaks with Python 4.0.

I’d suggest:

```py
try:
    import html
    html_unescape = html.unescape
except (AttributeError, ImportError):
    try:
        # the whole ordeal
```

Or even don’t use html.unescape and always define our own.

felixfontein: Yeah, that sounds better. About defining my own: doesn't sound like a very appealing idea to me.

PR #2579, opened 2016-12-05
Added basic page range navigation.

### LINKED ISSUE #2299: Add  the possibility of page range navigation
I have seen that some blogging platforms (like Wordpress or Blogger) allow the user to navigate the indexes up to a page range. That is to say, instead of showing "Older posts" /"Newer posts" they usually show a page range to navigate between indexes of posts.
I think this eases navigation a lot (particularly with large blogs). Could an option be implemented for the blogger to choose from classical older / newer style links or page range navigation?

felixfontein:
Fixes #2299.

This PR provides a basic page range navigation as asked for in #2299 (I hope). Needs better styling (both CSS and template).

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2379,8 +2379,8 @@ class Nikola(object):
         displayed_page_numbers = [utils.get_displayed_page_number(i, num_pages, self) for i in range(max(num_pages, 1))]
         page_links = [page_link(i, displayed_page_numbers[i], num_pages, False) for i in range(max(num_pages, 1))]
         if kw['show_index_page_navigation']:
-            map = {page_number - 1: link for page_number, link in zip(displayed_page_numbers, page_links)}
-            page_links_context = [map[i] for i in range(num_pages)]
+            temp_map = {page_number - 1: link for page_number, link in zip(displayed_page_numbers, page_links)}
+            page_links_context = [temp_map[i] for i in range(num_pages)]
         for i, post_list in enumerate(lists):
             context = context_source.copy()
             if 'pagekind' not in context:
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2408,7 +2408,13 @@ class Nikola(object):
         displayed_page_numbers = [utils.get_displayed_page_number(i, num_pages, self) for i in range(num_pages or 1)]
         page_links = [page_link(i, page_number, num_pages, False) for i, page_number in enumerate(displayed_page_numbers)]
         if kw['show_index_page_navigation']:
+            # Since the list displayed_page_numbers is not necessarily
+            # sorted -- in case INDEXES_STATIC is True, it is of the
+            # form [num_pages, 1, 2, ..., num_pages - 1] -- we order it
+            # via a map. This allows to not replicate the logic of
+            # utils.get_displayed_page_number() here.
             temp_map = {page_number - 1: link for page_number, link in zip(displayed_page_numbers, page_links)}
+            # Note that len(temp_map) >= num_pages.
             page_links_context = [temp_map[i] for i in range(num_pages)]
         for i, post_list in enumerate(lists):
             context = context_source.copy()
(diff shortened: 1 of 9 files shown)


Kwpolska:
[Bootstrap has pagination styling.](http://getbootstrap.com/components/#pagination)

felixfontein:
Ah, cool. I'll try that.

felixfontein:
Works great! (Except that there are no more ellipses, which can be annoying if you have a large amount of pages in your index.)

Kwpolska:
Why not? You could just have an ellipsis that links to `#` in the Bootstrap pager as well. (I’d prefer … instead, more pages displayed, and you may use raw Unicode characters. We mandate UTF-8 everywhere.)

felixfontein:
Now the ellipses are back, and the surrounding size can be modified easier (and is by default 5 instead of 3).

nikola/nikola.py
```diff
@@ -2374,11 +2376,16 @@ def generic_index_renderer(self, lang, posts, indexes_title, template_name, cont
                 lists.append(posts[:kw["index_display_post_count"]])
                 posts = posts[kw["index_display_post_count"]:]
         num_pages = len(lists)
+        displayed_page_numbers = [utils.get_displayed_page_number(i, num_pages, self) for i in range(max(num_pages, 1))]
```
Kwpolska: For a simpler comparison: `num_pages or 1`

nikola/nikola.py
```diff
@@ -2374,11 +2376,16 @@ def generic_index_renderer(self, lang, posts, indexes_title, template_name, cont
...
+        page_links = [page_link(i, displayed_page_numbers[i], num_pages, False) for i in range(max(num_pages, 1))]
```
Kwpolska:
```py
page_links = [page_link(i, page_number, num_pages, False) for i, page_number in enumerate(displayed_page_numbers)]
```

nikola/nikola.py
```diff
@@ -2374,11 +2376,16 @@ def generic_index_renderer(self, lang, posts, indexes_title, template_name, cont
...
+        if kw['show_index_page_navigation']:
+            temp_map = {page_number - 1: link for page_number, link in zip(displayed_page_numbers, page_links)}
```
Kwpolska: Why is this a dict? I’d expect the lists to be sorted already, so *this could be a list*. In any case, avoid iterating over `range(num_pages)` more than once — not only you might have missed the case when num_pages = 0, but also `[i[1] for i in sorted(temp_map.values())]` has more relation to the dict.

nikola/nikola.py
felixfontein: Ok.

nikola/nikola.py
felixfontein: Fixed.

nikola/nikola.py
felixfontein: The list is not sorted already: in case `INDEXES_STATIC == True`, `displayed_page_numbers` will be `[num_pages, 1, 2, 3, ..., num_pages - 1]`. In case `INDEXES_STATIC == False`, it will be `[1, 2, ..., num_pages]` (see `utils.get_displayed_page_number`). To avoid having to duplicate that kind of strange logic in here, I'm using the map.

In case `num_pages == 0`, `page_links_context` should have length 0 (and not length 1 as `displayed_page_numbers` and `page_links`). This isn't the case for `[i[1] for i in sorted(temp_map.values())]`, which will have length 1.

Anyway, in case `num_pages == 0`, the whole thing is not working that well anyway. In case `INDEXES_PRETTY_PAGE_URL` is `True`, a redirect to a non-existing page will be created, and in any case, the main index page is not created. How about fixing that here as well?

Kwpolska: If you want to fix this, go ahead. As it stands, this code LGTM, although this comment should appear in the source code to explain the temp_map’s existence.

felixfontein: I added the comment. I'll fix that maybe later in a different PR.

felixfontein:
Thanks for merging :)

PR #2583, opened 2016-12-05
Improving taxonomy plugin deactivation

### LINKED ISSUE #2581: Taxonomies can’t be disabled
Previously, I could do this:

```py
DISABLED_PLUGINS = ['render_indexes']
```

That won’t work anymore, because now everything is handled by `render_taxonomies`. There should be a way to disable taxonomies at will. Via #2535.

felixfontein:
Fixes #2581.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -812,11 +812,11 @@ class Nikola(object):
                 missing_plugins = []
                 for plugin_name in new_plugin_names:
                     if plugin_name not in self.config['DISABLED_PLUGINS']:
-                        missing_plugins.apped(plugin_name)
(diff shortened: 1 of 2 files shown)


nikola/nikola.py
```diff
@@ -351,6 +351,14 @@
     }
 }
 
+# Mapping old pre-taxonomy plugin names to new post-taxonomy plugin names
+PLUGIN_NAME_MAP = {
```
Kwpolska: TAXONOMY_PLUGIN_NAME_MAP?

felixfontein: How about `TAXONOMY_COMPATIBILITY_PLUGIN_NAME_MAP`? That also emphasizes that this is about compatibility.

PR #2586, opened 2016-12-08
Ensuring that generic_index_render always works fine with an empty post list

felixfontein:
When `INDEXES_STATIC == False`, an empty post list results in no generated pages for `generic_index_renderer`. You can see this when building an empty site with `INDEXES_STATIC == False`: no `index.html` is generated.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2404,8 +2404,11 @@ class Nikola(object):
             while posts:
                 lists.append(posts[:kw["index_display_post_count"]])
                 posts = posts[kw["index_display_post_count"]:]
+            if not lists:
+                lists.append([])
         num_pages = len(lists)
-        displayed_page_numbers = [utils.get_displayed_page_number(i, num_pages, self) for i in range(num_pages or 1)]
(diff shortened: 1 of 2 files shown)


nikola/nikola.py
```diff
@@ -2404,8 +2404,11 @@ def generic_index_renderer(self, lang, posts, indexes_title, template_name, cont
             while posts:
                 lists.append(posts[:kw["index_display_post_count"]])
                 posts = posts[kw["index_display_post_count"]:]
+            if not lists:
+                lists.append([])
         num_pages = len(lists)
-        displayed_page_numbers = [utils.get_displayed_page_number(i, num_pages, self) for i in range(num_pages or 1)]
+        assert num_pages >= 1
```
Kwpolska: Assertions are ugly and can be disabled, use an `if` and raise an exception yourself.

felixfontein: In that case, I'd rather remove that line.

felixfontein: (I don't understand why the other `asserts` are OK then, though.)

PR #2585, opened 2016-12-06
Keyword arguments in path handlers

Kwpolska:
This implements the generic functionality needed for #2580. I’ll try to do it for the Taxonomy stuff in a future PR (first, I need to understand the thing)

cc @felixfontein

PR #2589, opened 2016-12-09
Accept a `page` argument for taxonomy paths

Kwpolska:
This is #2585, cc @felixfontein. Note the change in the `if` for pagination — did I just fix a bug or was it intended?

### DIFF
--- a/nikola/plugins/misc/taxonomies_classifier.py
+++ b/nikola/plugins/misc/taxonomies_classifier.py
@@ -269,15 +269,21 @@ class TaxonomiesClassifier(SignalHandler):
         path, append_index, _ = self._parse_path_result(result)
         return self._postprocess_path(path, lang, append_index=append_index, dest_type='list')
 
-    def _taxonomy_path(self, name, lang, taxonomy, dest_type='page'):
+    def _taxonomy_path(self, name, lang, taxonomy, dest_type='page', page=None):
         """Return path to a classification."""
         if taxonomy.has_hierarchy:
             result = taxonomy.get_path(taxonomy.extract_hierarchy(name), lang, dest_type=dest_type)
         else:
             result = taxonomy.get_path(name, lang, dest_type=dest_type)
-        path, append_index, page = self._parse_path_result(result)
+        path, append_index, page_ = self._parse_path_result(result)
+
+        if page is not None:
+            page = int(page)
+        else:
+            page = page_
+
         page_info = None
-        if not taxonomy.show_list_as_index and page is not None:
+        if taxonomy.show_list_as_index and page is not None:
             number_of_pages = self.site.page_count_per_classification[taxonomy.classification_name][lang].get(name)
             if number_of_pages is None:
                 number_of_pages = self._compute_number_of_pages(self._get_filtered_list(taxonomy, name, lang), self.site.config['INDEX_DISPLAY_POST_COUNT'])
--- a/nikola/plugins/misc/taxonomies_classifier.py
+++ b/nikola/plugins/misc/taxonomies_classifier.py
@@ -205,7 +205,7 @@ class TaxonomiesClassifier(SignalHandler):
         """Given a list of posts and the maximal number of posts per page, computes the number of pages needed."""
         return min(1, (len(filtered_posts) + posts_count - 1) // posts_count)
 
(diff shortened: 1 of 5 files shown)


nikola/plugins/misc/taxonomies_classifier.py
```diff
@@ -269,15 +269,21 @@ def _taxonomy_index_path(self, name, lang, taxonomy):
...
+        path, append_index, page_ = self._parse_path_result(result)
+
+        if page is not None:
+            page = int(page)
+        else:
+            page = page_
+
         page_info = None
-        if not taxonomy.show_list_as_index and page is not None:
+        if taxonomy.show_list_as_index and page is not None:
```
felixfontein: Thanks for fixing that bug.

nikola/plugins/misc/taxonomies_classifier.py
```diff
@@ -269,15 +269,21 @@ def _taxonomy_index_path(self, name, lang, taxonomy):
         path, append_index, _ = self._parse_path_result(result)
         return self._postprocess_path(path, lang, append_index=append_index, dest_type='list')
 
-    def _taxonomy_path(self, name, lang, taxonomy, dest_type='page'):
+    def _taxonomy_path(self, name, lang, taxonomy, dest_type='page', page=None):
```
felixfontein: You also need a `page=None` argument for `_taxonomy_atom_path` (which must be passed to `_taxonomy_path`).

felixfontein: Ah, and in line 247 it should be `dest_type in ('page', 'feed')` and not `dest_type == 'page'`.

felixfontein:
There was another problem in `_postprocess_path` which only showed up when allowing page numbers for Atom feeds, so I decided to simply fix both things.

felixfontein:
The last commit shows how this actually can simplify things. (And it is a test case that the whole thing actually works.)

felixfontein:
The last commit () isn't completely good. In case the setting `INDEXES_PRETTY_PAGE_URL` happens to be `True`, it doesn't work correctly since it ignores the argument `force_addition`. That's in fact the only place where `force_addition == True`, and the argument is only ever needed for `page_path` and not for `page_link`.

How about keeping the new code, but only in `page_path` make a special case which handles `force_addition == True`? This avoids repeating more logic from the path handlers in the `TaxonomiesClassifier` plugin, but makes the functions seem more complex.

Kwpolska:
(“make it work” should be the top priority, not “make it look good”)

felixfontein:
Well, before the commit it worked well, so reverting the commit is a simple solution. On the other hand, I also like to avoid writing logic twice, and using `Nikola.link` and `Nikola.path` for the common situation (i.e. when `force_addition == False`) does that, even though it doesn't look so great. It's kind of a style question.

I guess undoing the commit is the simplest solution which makes it work, though.

felixfontein:
I decided to undo the commit. (I also fixed an unrelated typo.)

felixfontein:
Oh, another idea: how to add an additional argument to the path handler (like `alternative_path`), and then use that?

nikola/plugins/misc/taxonomies_classifier.py
```diff
@@ -218,11 +218,15 @@ def _postprocess_path(self, path, lang, append_index='auto', dest_type='page', p
...
         `'feed'`, `'rss'` or `'page'`.
 
-        Finally, if `dest_type` is `'page'`, `page_info` can be `None` or a tuple
+        If `dest_type` is `'page'`, `page_info` can be `None` or a tuple
         of two integers: the page number and the number of pages. This will
         be used to append the correct page number by calling
         `utils.adjust_name_for_index_path_list` and
         `utils.get_displayed_page_number`.
+
+        If `alternative_path` is set to `True`, `utils.adjust_name_for_index_path_list`
```
Kwpolska: You should explain what the difference would be in the docstring.

felixfontein: I added a very short explanation. (I want to avoid repeating the logic of `adjust_name_for_index_path_list` here.)

Kwpolska:
@ralsina, any review?

felixfontein:
Thanks!

PR #2590, opened 2016-12-10
Remove all uses of `assert` in the codebase

Kwpolska:
The `assert` statement can be disabled (`python -O`) and often does not
explain what went wrong. This replaces all uses of `assert` with an `if`
and a `ValueError`.

This needs a logic review, I had to invert conditions manually.

### DIFF
--- a/nikola/plugins/misc/taxonomies_classifier.py
+++ b/nikola/plugins/misc/taxonomies_classifier.py
@@ -66,7 +66,8 @@ class TaxonomiesClassifier(SignalHandler):
                     for lang in site.config['TRANSLATIONS'].keys():
                         # Extract classifications for this language
                         classifications[lang] = taxonomy.classify(post, lang)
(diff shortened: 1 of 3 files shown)


nikola/plugins/misc/taxonomies_classifier.py
```diff
@@ -66,7 +66,8 @@ def _do_classification(self, site):
                     for lang in site.config['TRANSLATIONS'].keys():
                         # Extract classifications for this language
                         classifications[lang] = taxonomy.classify(post, lang)
-                        assert taxonomy.more_than_one_classifications_per_post or len(classifications[lang]) <= 1
+                        if not taxonomy.more_than_one_classifications_per_post and len(classifications[lang]) > 1:
+                            raise ValueError("Too many {0} classifications for post {1}".format(taxonomy.classification_name, post.source_path)
```
felixfontein: There's one `)` missing at the end of the line.

PR #2593, opened 2016-12-10
Fixes #2592

### LINKED ISSUE #2592: Archives (CREATE_MONTHLY_ARCHIVE = True) think all months are "None"
https://irclogs.getnikola.com/2015/

To reproduce locally: get https://github.com/getnikola/irclogs-site create any random files in `posts/*.rst`.

felixfontein:
Forgot one `return`...

### DIFF
--- a/nikola/plugins/task/archive.py
+++ b/nikola/plugins/task/archive.py
@@ -97,7 +97,7 @@ class Archive(Taxonomy):
         elif len(classification) == 1:
             return classification[0]
         elif len(classification) == 2:
-            nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+            return nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
         else:
             # Fallback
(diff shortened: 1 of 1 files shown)


felixfontein:
@Kwpolska: the failing tests are the baseline. I don't see what went wrong here, though; maybe some dependency changed which now generates a slightly different output? Also, the same failing baseline differences show up for #2594, which has completely different code changes.

Kwpolska:
That’s a lxml update, I just rebuilt the baseline.

Kwpolska:
PS. small and obvious fixes do not need to be done through a pull request, you may commit to master directly.

felixfontein:
Nope, that doesn't work anymore. I tried that first, and I got an error message that a review is required when trying to push to master. So I had to convert it to a PR first.

Kwpolska:
Oh, that’s a side effect of requiring reviews and protecting branches. Which does not affect administrators, so I didn’t notice.

PR #2596, opened 2016-12-14
Forcing webassets to really rebuild files.

felixfontein:
I've experiencing a problem with assets bundling: if I create a bundle out of CSS files which are all generated by a plugin, the following happens:
  - when running `nikola build` first, none of the CSS files are there when generating the `create_bundles` task, whence an empty file is written;
  - the `create_bundles` task is executed after executing the tasks which write the CSS files;
  - when running `nikola build` a second time, nikola sees the CSS files and executes the `creates_bundles` task another time;
  - unfortunately, webassets sees that the existing bundle CSS file is older than the source files, and decides not to rebuild it.

So the result is an empty (unchanged) CSS bundle file, even though the input files changed (they are not empty), which is clearly wrong.

From the webassets source, I extracted a way to force it to rebuild the bundles. If that doesn't work in some cases, we could also simply delete the output file before asking it to rebuild; that should have the same result.

### DIFF
--- a/nikola/plugins/task/bundles.py
+++ b/nikola/plugins/task/bundles.py
@@ -79,7 +79,7 @@ class BuildBundles(LateTask):
                 env.register(output, bundle)
                 # This generates the file
                 try:
-                    env[output].urls()
+                    env[output].build(force=True)
                 except Exception as e:
                     self.logger.error("Failed to build bundles.")
                     self.logger.exception(e)


felixfontein:
Also thanks for merging!

Kwpolska:
PS. You could argue `.build()` was the right function all along, the docs assume a different use case than ours, and deep down both call `bundle._build()` nevertheless.

Issue #1639, opened 2015-03-21
Archives should allow moving to next and previous.


magmax:
Context variables injected in archives doesn't allow to manage "next page" or "previous page" links.

It would be great to allow it.

Kwpolska:
What do you mean, exactly?  What should be changed?

magmax:
The same way index.tmpl can manage `${prevlink}` and `${nextlink}` (file data/themes/base/templates/index_helper.tmpl), the template list_posts.tmpl should receive the link for the next and previous pages.

For example, in an environment with archives per year, it could be possible to add a link for "next year" and "previous year" in each of the archive pages. In an environment with `CREATE_MONTHLY_ARCHIVE=True`, it should have a "next month" and "previous month" buttons.

magmax:
An example taken from one of the blogs in the "users" section:

the archive has these years: http://www.curiousefficiency.org/archive.html

So it would be great to be able to move from http://www.curiousefficiency.org/2011/index.html to the previous/next years without having to press on the archive index again.

Thank you!

Kwpolska:
This certainly makes sense.  It could be done, I’ll try to look into it when I have some time.

jjconti:
Can I work on this or is someone already on it?

ralsina:
@jjconti feel free!

Aeyoun:
Should look into including them in the Atom feeds too (for when archives are indexes).

Kwpolska:
Fixed ni #2599.

PR #2598, opened 2016-12-17
Allowing cross-classification navigation

felixfontein:
Inserts links to previous/parent/next pages/Atom/RSS to context for taxonomy classification pages.

This allows to implement #1639 with the following variables:
 - `previous_archive*`: whatever archive entry comes before this one (sorted alphabetically; this is more useful for tags or other classifications than for archives);
 - `next_archive*`: whatever archive entry comes before this one (sorted alphabetically; this is more useful for tags or other classifications than for archives);
 - `previous_archive_sibling*`: for months, the previous month in this year (or `None` if there is none); same for years and days;
 - `next_archive_sibling*`: for months, the next month in this year (or `None` if there is none); same for years and days;
 - `previous_archive_samelevel*`: for months, the previous month in this year or the previous years (or `None` if there is none); same for years and days;
 - `next_archive_samelevel*`: for months, the next month in this year or the previous years (or `None` if there is none); same for years and days;
 - `parent_archive*`: for months, the year; for years, the global archive; and for days, the month.

Here, `*` can be replaced with nothing (to get classification identifier), `_name` for a friendly name (like 'May 2015'), `_link` for a link to the classification page, `_atom` for a link to the Atom feed for this classification, and `_rss` for a link to the RSS feed for this classification.

I haven't modified any templates yet; whoever wants to do this, feel free.

Also, nothing is added to Atom feeds (as @da2x suggested in #1639) since I have no idea how it should/could be added there.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -582,6 +582,14 @@ class Taxonomy(BasePlugin):
         Whether to include all classifications for all languages in every
         language, or only the classifications for one language in its language's
         pages.
+
+    insert_classification_navigation_links = False:
+        If set to True, inserts links to previous and following
+        classifications (based on the order induced by natsort and
+        modified by Taxonomy.sort_classifications) and, for hierarchical
+        taxonomies, links to the parent (if exists), previous and following
+        siblings, and previous and following classifications on the same
+        hierarchy level.
     """
 
     name = "dummy_taxonomy"
@@ -607,6 +615,7 @@ class Taxonomy(BasePlugin):
     minimum_post_count_per_classification_in_overview = 1
     omit_empty_classifications = False
     also_create_classifications_from_other_languages = True
+    insert_classification_navigation_links = False
 
     def is_enabled(self, lang=None):
         """Return True if this taxonomy is enabled, or False otherwise.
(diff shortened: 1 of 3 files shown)


Kwpolska:
This is far too overcomplicated.

felixfontein:
(The code concerning translations for `'{month} {year}'` and `'{month} {day}, {year}'` should be adjusted and these variables should be added to transiflex before this PR is merged.)

felixfontein:
Why? It is too complicated for only archives maybe, but not too complicated if you want to use it for other taxonomies as well (like sections, tags, categories, ...).

Kwpolska:
How useful would that be though? I don’t think it would. You **can’t** reasonably do it all for any other taxonomies, since those are unordered. For hierarchies (categories), getting a link to the parent is perfectly reasonable (and should be really simple), but even then there’s no “next category” or whatever

The solution should be much simpler. To me, this looks like 141 lines of gibberish code.

felixfontein:
There's always the same order which will appear on the taxonomy overview page, which is perfectly fine for me.

Kwpolska:
What happens when you add a new tag, which can happen pretty often on some blogs? Rebuilding pages just to fix a link only a handful of people would use, and that makes no sense for anything larger?

felixfontein:
We could enable it optionally for taxonomies, so that by default that isn't added. Then the default tag plugin could disable it (or have a setting for that, or whatever).

felixfontein:
Rebased to account for the commit already merged. Closing for now as this doesn't have a chance to be merged.

PR #2599, opened 2016-12-18
Archive navigation

Kwpolska:
This is a simple solution for #1639, done in 26 readable lines. This is not finished, template stuff and perhaps friendly names need to be worked in.

cc @felixfontein, @magmax

(PS. I also did some minor cleanups re TreeNodes on master in c79164b, 6027bf5, 41162b2)


nikola/plugins/task/archive.py
```diff
@@ -150,10 +152,34 @@ def provide_context_and_uptodate(self, classification, lang, node=None):
...
+                context["up_level"] = self.site.link('/'.join(hierarchy[:-1]), lang)
+            else:
+                context["up_level"] = None
+
+            flat_hierarchy = self.site.flat_hierarchy_per_classification['archive'][lang]
+            nodelevel = len(hierarchy)
+            flat_samelevel = [node.classification_name for node in flat_hierarchy if len(node.classification_path) == nodelevel]
+            # We need to sort it. Natsort means it’s year 10000 compatible!
+            flat_samelevel = natsort.natsorted(flat_samelevel, alg=natsort.ns.F | natsort.ns.IC)
+            idx = flat_samelevel.find(classification)
```
felixfontein: Why not precompute a LUT in `postprocess_posts_per_classification` instead of recomputing this every time this function is called?

nikola/plugins/task/archive.py
```diff
@@ -150,10 +152,34 @@ def provide_context_and_uptodate(self, classification, lang, node=None):
...
+            if idx == -1:
+                raise Exception("Cannot find classification {0} in flat hierarchy!".format(classification))
+            previdx, nextidx = idx - 1, idx + 1
+            # If the previous index is -1, or the next index is 1, the previous/next archive does not exist.
+            context["previous_archive"] = self.site.link(flat_samelevel[previdx], lang) if previdx != -1 else None
+            context["next_archive"] = self.site.link(flat_samelevel[nextidx], lang) if nextidx != -1 else None
```
felixfontein: I think it would make sense to also insert the names of the archives. Links for the Atom feeds (see da2x' comments in #1639) would be nice, too.

nikola/plugins/task/archive.py
```diff
@@ -150,10 +152,34 @@ def provide_context_and_uptodate(self, classification, lang, node=None):
...
+        if kw["create_archive_navigation"]:
+            if hierarchy:
+                # Up level link makes sense only if this is not the top-level
+                # page (hierarchy is empty)
+                context["up_level"] = self.site.link('/'.join(hierarchy[:-1]), lang)
+            else:
+                context["up_level"] = None
+
+            flat_hierarchy = self.site.flat_hierarchy_per_classification['archive'][lang]
+            nodelevel = len(hierarchy)
```
felixfontein: You should also insert `nodelevel` into the context so that link texts (next year/month/day) can be chosen appropriately without unnecessary complicated code (like manually counting the number of slashes in the classification etc.).

nikola/plugins/task/archive.py
```diff
@@ -150,10 +152,34 @@ def provide_context_and_uptodate(self, classification, lang, node=None):
...
             "title": title,
             "pagekind": [page_kind, "archive_page"],
         }
+
+        # Generate links for hierarchies
+        if kw["create_archive_navigation"]:
+            if hierarchy:
+                # Up level link makes sense only if this is not the top-level
+                # page (hierarchy is empty)
+                context["up_level"] = self.site.link('/'.join(hierarchy[:-1]), lang)
```
felixfontein: Your calls to `self.site.link` (this one and the two below) must specify the taxonomy name (here: `archive`).

nikola/plugins/task/archive.py
Kwpolska: Those links do not sound anywhere near useful to me. If someone really needs an Atom link to the archives for a *previous* level (why?), they can just click the link they want and then the Atom link.

nikola/plugins/task/archive.py
Kwpolska: Do you mean a lookup table for `flat_samelevel`, or for `idx` too?

Kwpolska:
I tried adding some support in templates, but for some reason, it doesn’t work in the monthly archives (2012/03 for example). What template do they use? It looks really strange, because the right variables are in the context.

nikola/plugins/task/archive.py
felixfontein: For both of them. You can also make a lookup for the resulting nodes / classifications if you want.

felixfontein:
`list_post.tmpl` for post lists, `archiveindex.tmpl` for archives, and `list.tmpl` for list of years or months.

Kwpolska:
Why are there so many?! I’ll have to move some things then, guess I’ll go back to depending on `pagekind`.

felixfontein:
I haven't tested it, but I think it looks good.

Kwpolska:
I didn’t finish the theme support yet, not merging.

felixfontein:
My review was (mostly) for the code, not for the theme support :)

Kwpolska:
Alright, theming is done, could use a review of that and some testing (although I did my best to make it work, maybe there’s some edge case I didn’t notice)

felixfontein:
How about also inserting the names of the linked to archives into the context? That would give themes more flexibility (i.e. incorporating the archive's name instead of just naming the links next/previous/up).

Kwpolska:
Sure, do you want to implement that?

felixfontein:
That should do the job.

Kwpolska:
Thanks for the help!

felixfontein:
You're welcome. I was just about to write that I tested this (with my own themes) and it works great!

PR #2601, opened 2016-12-20
Improving friendly archive names.

felixfontein:
Something from #2598 that should be in master.

@Kwpolska: how do I add new strings? Insert them into transifex first? In the `messages_xx.py` files? The `.po` files?

### DIFF
--- a/nikola/plugins/task/archive.py
+++ b/nikola/plugins/task/archive.py
@@ -97,10 +97,20 @@ class Archive(Taxonomy):
         elif len(classification) == 1:
             return classification[0]
         elif len(classification) == 2:
-            return nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+            month = nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+            if only_last_component:
+                return month
+            else:
+                year = classification[0]
+                return self.site.MESSAGES[lang].get('{month} {year}', '{month} {year}').format(year=year, month=month)
         else:
-            # Fallback
-            return '/'.join(classification)
+            day = int(classification[2])
+            if only_last_component:
+                return str(day)
+            else:
+                year = classification[0]
+                month = nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+                return self.site.MESSAGES[lang].get('{month} {day}, {year}', '{month} {day}, {year}').format(year=year, month=month, day=day)
 
     def get_path(self, classification, lang, dest_type='page'):
         """A path handler for the given classification."""
--- a/nikola/plugins/task/archive.py
+++ b/nikola/plugins/task/archive.py
@@ -102,7 +102,7 @@ class Archive(Taxonomy):
                 return month
             else:
                 year = classification[0]
-                return self.site.MESSAGES[lang].get('{month} {year}', '{month} {year}').format(year=year, month=month)
+                return self.site.MESSAGES[lang]['{month} {year}'].format(year=year, month=month)
         else:
             day = int(classification[2])
             if only_last_component:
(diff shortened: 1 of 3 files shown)


nikola/plugins/task/archive.py
```diff
@@ -97,10 +97,20 @@ def get_classification_friendly_name(self, classification, lang, only_last_compo
         elif len(classification) == 1:
             return classification[0]
         elif len(classification) == 2:
-            return nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+            month = nikola.utils.LocaleBorg().get_month_name(int(classification[1]), lang)
+            if only_last_component:
+                return month
+            else:
+                year = classification[0]
+                return self.site.MESSAGES[lang].get('{month} {year}', '{month} {year}').format(year=year, month=month)
```
Kwpolska: You don’t really need .get(), `[]` will do it because you are allowed to assume all locales are correct.

felixfontein: I wanted to keep this code there until the strings are in transifex.

felixfontein:
Ok, this PR is ready from my side.

felixfontein:
Very interesting. I've just replicated what's used for generating the archive titles. So they've been broken for a long time.
I'll try to fix that...

felixfontein:
One possibility would be the `babel` library (http://babel.pocoo.org/en/latest/api/dates.html):
 * `babel.dates.get_month_names('wide', context='stand-alone', locale='pl_PL')[12]` yields `grudzień`;
 * `babel.dates.get_month_names('wide', context='format', locale='pl_PL')[12]` yields `grudnia`;
 * `babel.dates.format_date(now, babel.dates.get_date_format('long', locale='pl_PL'), locale='pl_PL')` yields `23 grudnia 2016`.

With `locale` I seem to get `23 grudzień 2016` (format string `'%d %B %Y'` for locale `pl_PL.UTF-8`).

Also, the whole situation is complicated by the fact that some translations of the strings for `Posts for {month}, {day} {year}` etc. already do some grammar adjustment when these are regular, like `Kirjoituksia ajalta {day}. {month}ta {year}` in Finnish where you always add `ta` after the month's name when having it in a date, while the regular `Posts for {month} {year}` string doesn't have this adjustment because you don't have the suffix there (`Postauksia ajalle {month} {year}`).

Changing this behavior in a way which doesn't break things will be hard (or introduce a lot of extra things).

felixfontein:
How about providing additional formatting variables `month_babel_standalone` and `month_babel_format` for all month-related messages? Then translators can decide which approach to use.

Also, should `babel` be a hard requirement for Nikola, or a soft requirement? If soft, what should happen if it is not available?

Kwpolska:
So, there are two ways we could go:

* ignore the thing and get slightly ugly output in some cases, or
* use Babel, add *yet another* dependency, and get _pretty_ output, along with fixing some other language woes (plural forms)

I’m not sure which one would be better, nicer to the world (users, distributors), and so on. Comments and opinions welcome…

felixfontein:
Where do we need plural form handling at the moment? In the strings, only `%d min remaining to read`, `old posts, page %d` and `page %d` incorporate numbers. But it might be more useful in the future. But it might be more used in the future, and then it will be pretty handy. On the other hand, adding support for this will require some redesign, as the simple `messages_xx.py` format cannot handle that without some additional work.

felixfontein:
Hmm, how about merging this PR and continuing the discussion in a new issue, as this affects many more parts than just archive names?

felixfontein:
Do you want to start a discussion, or should I?

Kwpolska:
I’m on it.

felixfontein:
@Kwpolska: can you push the new string to transifex?

Kwpolska:
Sure, done.

felixfontein:
Thanks!

Issue #1696, opened 2015-05-07
Some features should be moved to plugins


ralsina:
Specially features that are not widely used, or "bad" for some reason. Feel free to add more ;-)
- [x] Annotations via annotateit.org (removed)
- [x] Post password-protection (removed)
- [ ] Tag cloud
- [x] Slides

Kwpolska:
:+1:

felixfontein:
The tag cloud JSON file is now written by a dedicated plugin `render_tag_cloud` (since PR #2535). Moving to to https://github.com/getnikola/plugins should be straightforward.

felixfontein:
Moving the post encryption feature into a plugin should be no problem by adding a signal handler for the `compiled` signal emitted by `Post.compile`.

gwax:
The commenting systems could probably be carved out as plugins as well. That way you could install the one you want and never consider the rest.

Kwpolska:
You don’t need to care about other comment systems. Those templates will simply be ignored, and won’t ever appear in output (plus making them plugins would be hard).

gwax:
I care about the commenting systems less from the perspective of using nikola and more from the perspective of theme development; they are annoyingly entangled with the rest of the base templates. Though, maybe that can be better solved by rearranging the base templates slightly.

tritium21:
It is a major refactor to move comments into their own plugins (its been requested before)

PR #2604, opened 2016-12-25
Moving post encryption into plugin.

felixfontein:
Implements part of #1696.


Kwpolska:
Meh, I’d prefer to just delete this misfeature here and now. Over 2 years have passed since we deprecated that thing.

felixfontein:
I can live with that, too. Getting it out of `post.py` is a good start, though, since it shows people who really want to use this (for whatever reason) how they can still do that even though the feature is gone from Nikola.

felixfontein:
(Rebased to remove conflicts and get rid of change which was already included in #2607.)

Kwpolska:
Why should we help people do outright stupid things?

felixfontein:
Because we allowed to do that in the first place, and we want to avoid backwards-compatibility breaking changes in minor versions?

Kwpolska:
We should probably get v8 going already.

nikola/plugins/misc/post_encryption.py
```diff
@@ -0,0 +1,118 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+# PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+# OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"""Render the taxonomy overviews, classification pages and feeds."""
```
Kwpolska: not really

felixfontein: LOL

felixfontein:
Next step: removing that plugin in v8.

Issue #2602, opened 2016-12-23
Make Post objects sortable/comparable


liffiton:
I'd like to be able to [re]order the posts in a template (e.g., ``tag.tmpl``, and specifically in Jinja in my case) by their titles.

I'm imagining I'd write something like ``for post in posts|sort``.  That does work after adding an ``__lt__`` operator to the Post class:

```python
def __lt__(self, other):
   return self.title() < other.title()
```

I realize that sorting by title is not the only way to sort posts, though, so making the sort filter order in this way may not be the best implementation.  An alternative would be to leverage the "attribute" parameter of Jinja's [sort filter](http://jinja.pocoo.org/docs/dev/templates/#sort), which lets one tell it what to sort by; this would require exposing the Post object's title as an attribute:

```python
@property
def title_str(self):
    return self.title()
```

Then this would work: ``for post in posts|sort(attribute="title_str")``

Kwpolska:
I can think of a better solution: a smarter `sort_posts` filter/function, available for templates and general use, that would handle all those edge cases. I’ll work on it later.

PR #2608, opened 2016-12-27
Add a sort_posts function

Kwpolska:
This is #2602 — it adds a `sort_posts` function (available as Jinja filter in global context), which allows general-purpose timeline sorting. Example usage:

```python
sort_posts(timeline, 'title', 'date')
sort_posts(timeline, 'author', reverse=True)
```

cc @liffiton — does this match your needs?


felixfontein:
From `sort_posts(timeline, 'title', 'date')` I'd assume that the posts are sorted by title, and if they have the same title, then they are sorted by date. But the code actually sorts them the other way around: first by date, and only for posts on the same date (i.e. with identical timestamp, as date means timestamp) by title.

Kwpolska:
That’s a sane assumption to make, I added a little `reversed()` call to account for this.

felixfontein:
How about being able to sort some keys reversed, and some not, by prepending `reverse_` to the key name? Standard use case is sorting by date and title: you might want to sort by date reversed to get the newest posts first, but by title alphabetically.

Kwpolska:
@felixfontein done — prepend with `-` to get a key reversed (same syntax as used in Django). Review welcome.

nikola/utils.py
```diff
@@ -1947,6 +1950,35 @@ def clean_before_deployment(site):
...
+    """
+    # We reverse the keys to get the usual ordering method: the first key
+    # provided is the most important sorting predicate (first by 'title', then
+    # by 'date' in the first example)
+    for key in reversed(keys):
+        try:
+            # An attribute (or method) of the Post object
+            a = getattr(posts[0], key)
+            if callable(a):
+                keyfunc = operator.methodcaller(key)
```
felixfontein: Does this actually work when `key[0] == '-'`?

Kwpolska: Oh, right. Fixed.

PR #2614, opened 2017-01-01
Fix #2613 — allow enabling pretty URLs with per-post setting

### LINKED ISSUE #2613: Per page pretty url config is not working
I have disabled pretty urls in `conf.py`

```py
PRETTY_URLS = False
```

For some posts, I don't want this, so I have set 

```
.. pretty_url: True
```

But Nikola is ignoring this.

Kwpolska:
This is #2613. This is an example implementation, and it may not be merged if deemed unnecessary.

cc @felixfontein, @ChillarAnand

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -277,12 +277,10 @@ class Post(object):
         return '<Post: {0!r} {1}>'.format(self.source_path, m.hexdigest())
 
     def _has_pretty_url(self, lang):
-        if self.pretty_urls and \
-                self.meta[lang].get('pretty_url', '') != 'False' and \
-                self.meta[lang]['slug'] != 'index':
-            return True
-        else:
-            return False
+        m = self.meta[lang].get('pretty_url', '')
+        return ((self.pretty_urls and m != 'False' and
+                 self.meta[lang]['slug'] != 'index') or
+                (not self.pretty_urls and m == 'True'))
 
     @property
     def is_mathjax(self):
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -278,9 +278,12 @@ class Post(object):
 
     def _has_pretty_url(self, lang):
         m = self.meta[lang].get('pretty_url', '')
-        return ((self.pretty_urls and m != 'False' and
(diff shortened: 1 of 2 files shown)


nikola/post.py
```diff
@@ -277,12 +277,10 @@ def __repr__(self):
...
     def _has_pretty_url(self, lang):
-        if self.pretty_urls and \
-                self.meta[lang].get('pretty_url', '') != 'False' and \
-                self.meta[lang]['slug'] != 'index':
-            return True
-        else:
-            return False
+        m = self.meta[lang].get('pretty_url', '')
+        return ((self.pretty_urls and m != 'False' and
+                 self.meta[lang]['slug'] != 'index') or
```
felixfontein: Shouldn't `self.meta[lang]['slug'] != 'index'` be `and`ed with both sides of `or`? I.e. something like
```
    return self.meta[lang]['slug'] != 'index' and ((self.pretty_urls and m != 'False') or (not self.pretty_urls and m == 'True'))
```

Kwpolska: It could be, but if you are specifying `pretty_url` as a special case, you probably don’t want it overridden by some magic. In fact, we could go the other way: allow `index` to create a pretty URL if there is an override, even if self.pretty_urls is True. Which way should I go?

nikola/post.py
```diff
@@ -277,12 +277,10 @@ def __repr__(self):
...
+                (not self.pretty_urls and m == 'True'))
```
felixfontein: Why not replace the whole logic with:
``` .py
    m = self.meta[lang].get('pretty_url', repr(self.pretty_urls))
    return self.meta[lang]['slug'] != 'index' and m == 'True'
```
?

ChillarAnand:
@Kwpolska mentioned [an alternate solution here here](https://github.com/getnikola/nikola/issues/2613#issuecomment-269910518). With that solution, users have to change `posts` directory structure whenever pretty urls is necessary. 

I prefer this fix as directory structure can remain same. But its up to you to decide.

PR #2617, opened 2017-01-07
Work around conflicts between posts and section indexes (via #2613)

Kwpolska:
This is a workaround, because it uses the same mechanism `PAGE_INDEXES`
use to prevent generating pages. A real solution would need to figure
out that those pages are not sections.

cc @felixfontein, @ChillarAnand. Via #2613.


ChillarAnand:
Thanks @Kwpolska

PR #2620, opened 2017-01-08
Fix #2619 — promote Nikola in a meta generator tag

### LINKED ISSUE #2619: Suggestion for adding a generator meta tag
I have been using [wappalyzer](https://www.wappalyzer.com/) for a number of years now and what I personally find impressive about it is that it gives me the opportunity to discover new programs or projects every time I browse a website; thus making me a new user to that particular set of tools, like many other users of it.

I have noticed that projects like Hugo, Blogger, Jekyll, and so forth are detected by wappalyzer, whereas Nikola websites are not.

After a couple of seconds of investigation I have noticed that Nikola does not use a `<meta name="generator" content="Nikola 7.8.1">` meta tag.

In my humble, yet honest opinion, if we added this specific meta tag in Nikola, it would broaden its user base even further.

What do you guys think?

Kwpolska:
This is #2619. cc @stefanos82.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -24,6 +24,8 @@ Bugfixes
 Features
 --------
 
+* Add meta generator tag to default templates to promote Nikola
+  (Issue #2619)
 * Add ``nikola new_post -d`` and ``NEW_POST_DATE_PATH`` to allow
   automatic creation of year/month/day (date-based) directory
   structures (Issue #2513)
--- a/nikola/data/themes/base/templates/base_helper.tmpl
+++ b/nikola/data/themes/base/templates/base_helper.tmpl
@@ -35,7 +35,8 @@ lang="${lang}">
     %endif
 
     ${html_stylesheets()}
-    <meta content="${theme_color}" name="theme-color">
+    <meta name="theme-color" content="${theme_color}">
+    <meta name="generator" content="Nikola https://getnikola.com/">
     ${html_feedlinks()}
     <link rel="canonical" href="${abs_link(permalink)}">
 
--- a/nikola/data/themes/base/templates/base_helper.tmpl
+++ b/nikola/data/themes/base/templates/base_helper.tmpl
@@ -36,7 +36,7 @@ lang="${lang}">
 
     ${html_stylesheets()}
     <meta name="theme-color" content="${theme_color}">
-    <meta name="generator" content="Nikola https://getnikola.com/">
+    <meta name="generator" content="Nikola (getnikola.com)">
     ${html_feedlinks()}
     <link rel="canonical" href="${abs_link(permalink)}">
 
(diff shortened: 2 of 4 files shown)


stefanos82:
Actually, you don't need Nikola's URL inside `content`; just the name is alright.

Kwpolska:
It isn’t enough, because the name alone might not drive traffic to us (name is pretty generic, and Google might not help us here)

stefanos82:
Allow me to disagree so I can explain my reasoning. 

The `generator` meta element 

> must not be used on pages whose markup is not generated by software, e.g. pages whose markup was written by a user in a text editor.

You can read more about it from W3's [the meta element](https://www.w3.org/TR/html5/document-metadata.html#the-meta-element) page.

Also, about traffic...you don't need to worry about it. 

Here's an excerpt from Nikola's own generated HTML source code:

```html
<meta name="author" content="The Nikola Team">
<meta property="og:site_name" content="Nikola">
<meta property="og:title" content="Static Site Generator">
<meta property="og:url" content="https://getnikola.com/index.html">
<meta property="og:description" content="Nikola — Static Site Generator
        
            
                In goes content, out comes a website, ready to deploy.
                Get started with Nikola » Learn about Nikola features »
    ">
<meta property="og:type" content="article">
<meta property="article:published_time" content="2015-07-08T09:39:34Z">
<meta name="twitter:card" content="summary">
<meta name="twitter:site" content="@GetNikola">
</head>
```

As you can see, the `<meta property="og:url" content="https://getnikola.com/index.html">` tag directs all engines at the right direction.

Kwpolska:
`og:url` is the permalink to *current* page. It does not link to `getnikola.com` outside of `getnikola.com`.

But what do you mean with the standard excerpt? It does not say anything about linking.

stefanos82:
It's not about linking, but about indicating properly what program has been used to generate the current page.

Apart from the aforementioned meta tags, inside `<head>` we have `<link rel="canonical" href="https://getnikola.com/index.html">` that behaves the same as `og:url` permalink to the current page.

So, I'm pretty much sure, let alone confident  that search engine bots will be able to parse the entire website and attract the appropriate traffic to Nikola.

Anyhow, if you think by adding the hyperlink inside content meta tag would help the traffic, by all means use it.

felixfontein:
I think `getnikola.com` should definitely be included in the generator, just to resolve possible ambiguities. I like the version `Nikola (getnikola.com)`.

Kwpolska:
Merging, this goes into v7.8.2 which will be out soon.

stefanos82:
@Kwpolska I forgot to ask: is this option going to be included in `conf.py` as a flag option?

I'm asking because I'm sure some folks will ask in the future how to turn it off, the same way we use `expose_php = Off;` in `php.ini`.

Not everyone want to let the public know what set of tools have they used to build their websites.

felixfontein:
Currently it is enabled by default; the only way to turn it off is to edit the theme resp. create a child theme which doesn't have that meta tag set.

stefanos82:
That's fine for now, but maybe in the near future we could add this option so we could avoid unnecessary headaches.

Other than that, it looks good to me. Well done mate :+1: !

felixfontein:
I've created an issue so we won't forget...

PR #2622, opened 2017-01-08
Allowing to compile MarkDown posts from string to string.

felixfontein:
Allows to compile MarkDown from string to string (for getnikola/plugins#201).

### DIFF
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -64,14 +64,14 @@ class CompileMarkdown(PageCompiler):
         self.config_dependencies.append(str(sorted(site.config.get("MARKDOWN_EXTENSIONS"))))
 
     def compile_string(self, content, source_path=None, is_two_file=True):
-        """Compile MarkDown into HTML strings."""
+        """Compile Markdown into HTML strings."""
         if markdown is None:
             req_missing(['markdown'], 'build this site (compile Markdown)')
         self.extensions += self.site.config.get("MARKDOWN_EXTENSIONS")
         if not is_two_file:
             _, content = self.split_metadata(content)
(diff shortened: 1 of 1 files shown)


nikola/plugins/compile/markdown/__init__.py
```diff
@@ -63,6 +63,15 @@ def set_site(self, site):
 
         self.config_dependencies.append(str(sorted(site.config.get("MARKDOWN_EXTENSIONS"))))
 
+    def compile_to_string(self, content):
```
Kwpolska: But please match the reST plugin API.

```python
    def compile_string(self, data, source_path=None, is_two_file=True):
        """Compile Markdown into HTML strings."""
```

felixfontein: Done.

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -64,14 +64,14 @@ def set_site(self, site):
...
-        """Compile MarkDown into HTML strings."""
+        """Compile Markdown into HTML strings."""
         if markdown is None:
             req_missing(['markdown'], 'build this site (compile Markdown)')
         self.extensions += self.site.config.get("MARKDOWN_EXTENSIONS")
         if not is_two_file:
             _, content = self.split_metadata(content)
         output = markdown(content, self.extensions, output_format="html5")
-        output, shortcode_deps = self.site.apply_shortcodes(output, filename=source, with_dependencies=True, extra_context=dict(post=post))
+        output, shortcode_deps = self.site.apply_shortcodes(output, filename=source_path, with_dependencies=True, extra_context={'post': post})
```
felixfontein: Thanks for fixing that!

felixfontein:
Thanks!

PR #2623, opened 2017-01-08
Move shortcode handling to reST compiler’s compile_string

Kwpolska:
This is an API change, and I’m well aware of that. I think we can allow it, because nobody else seems to be using it.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -304,7 +304,10 @@ class PageCompiler(BasePlugin):
         self.compile_html(source, dest, is_two_file)
 
     def compile_string(self, data, source_path=None, is_two_file=True, post=None, lang=None):
-        """Compile the source file into HTML strings (with shortcode support)."""
+        """Compile the source file into HTML strings (with shortcode support).
+
+        Returns a tuple of at least two elements (html string, shortcode dependencies).
+        """
         # This function used to have some different APIs in different places.
         raise NotImplementedError()
 
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -303,6 +303,11 @@ class PageCompiler(BasePlugin):
         # ignore `compile_html`
         self.compile_html(source, dest, is_two_file)
 
+    def compile_string(self, data, source_path=None, is_two_file=True, post=None, lang=None):
+        """Compile the source file into HTML strings (with shortcode support)."""
+        # This function used to have some different APIs in different places.
+        raise NotImplementedError()
+
     # TODO remove in v8
     def compile_html(self, source, dest, is_two_file=True):
         """Compile the source, save it on dest (DEPRECATED)."""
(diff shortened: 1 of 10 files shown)


felixfontein:
Fine for me. I've been using the `compile_string` function so far, but I don't mind it processing shortcodes as well.

nikola/plugin_categories.py
```diff
@@ -303,6 +303,11 @@ def compile(self, source, dest, is_two_file=True, post=None, lang=None):
         # ignore `compile_html`
         self.compile_html(source, dest, is_two_file)
 
+    def compile_string(self, data, source_path=None, is_two_file=True, post=None, lang=None):
+        """Compile the source file into HTML strings (with shortcode support)."""
```
felixfontein: Maybe the documentation should say that this has to return a tuple whose first component is the HTML fragment?

nikola/plugins/compile/html.py
```diff
@@ -41,6 +41,10 @@ class CompileHtml(PageCompiler):
     name = "html"
     friendly_name = "HTML"
 
+    def compile_string(self, data, source_path=None, is_two_file=True, post=None, lang=None):
+        """Compile HTML into HTML strings, with shortcode support."""
+        return self.site.apply_shortcodes(data, with_dependencies=True, extra_context={'post': post})
```
felixfontein: The `if not is_two_file:` part from `compile` should be moved here, too, shouldn't it?

felixfontein:
About ipynb: looks ok, but neither I'm not familiar with its internal API nor do I use it in my blogs, so no idea how good it works...

nikola/plugins/compile/html.py
Kwpolska: Sure, fixed.

nikola/plugin_categories.py
Kwpolska: Done.

nikola/plugin_categories.py
```diff
@@ -304,7 +304,10 @@ def compile(self, source, dest, is_two_file=True, post=None, lang=None):
         self.compile_html(source, dest, is_two_file)
 
     def compile_string(self, data, source_path=None, is_two_file=True, post=None, lang=None):
-        """Compile the source file into HTML strings (with shortcode support)."""
+        """Compile the source file into HTML strings (with shortcode support).
+
+        Returns a tuple of at least two elements (html string, shortcode dependencies).
```
felixfontein: Someone might interpret this that shortcode dependencies should always be the second element in the list (which is not true). Should we change the implementations to always return this as second, or adjust the documentation?

Kwpolska: > Returns a tuple of at least two elements: HTML string [0] and shortcode dependencies [last].

felixfontein:
LGTM

Kwpolska:
Merging, and let’s hope nobody else breaks. (It wasn’t a public API before anyway)

Issue #2627, opened 2017-01-08
Category sorting is not sufficient


kayhayen:
Hello there,

I am using tags in "posts", to make my content easily browse able on http://speedcenter.nuitka.net/categories/index.html

With a "post", I am forced to give a date, presumably for ordering
purposes. I am currently using an old date, that I would even like
to be invisible. Or maybe I am going to use a reasonable one in
the future, e.g. the commit date of the test case. But it's going to
be the same one in a lot of cases if not always all cases.

The individual tags page is currently randomly sorted, listing the
cases in the RSS feed in arbitrary order, instead of sorting by name,
which would be better.

Is there a way, or any chance, you could add the slug or title as a
tie breaker when sorting posts with identical date? I presume this
must also affect 2 blog articles dated identically, order will be pseudo
random, file system order maybe?

Definitely not necessarily reproducible.

Thanks in advance,
Kay

felixfontein:
Currently, posts are sorted first by priority (which doesn't do anything if you didn't define priorities manually), then by date, and then by `source_path` of the post. Since I cannot see from the generated pages what the source paths were: do you have any "random" ID/hash/... in the source paths which could cause this?

felixfontein:
Oh: and which version of Nikola are you running?

kayhayen:
This was 7.8.1, but soon will be 7.8.2 which I have read about. If you consider e.g. this link (might go away, due to renames of tags coming up): http://speedcenter.nuitka.net/categories/27_develop_steady.html

You can see that I am using the default slug, and it appears not sorted. Or do you mean with source_path, the file system path? They all live in "constructs/construct-TitleCasedName", no hashes are used in paths.

Yours,
Kay

felixfontein:
Hmm, the sorting didn't change between 7.8.1 and 7.8.2, so 7.8.2 should have the same problem/bug. `source_path` is the path in the file system, so yeah, they should be sorted. I'll investigate that.

felixfontein:
I can't reproduce this. Would it be possible for you to upload the source files somewhere? Please use `tar` or something else which preserves file system ordering of files and doesn't sort the filenames in a directory.

kayhayen:
Here you go, this is a tar dump of my work directory, with "nuitka" and ".git" removed. It's not that huge. I am using ext4 file system in case that matters:

http://nuitka.net/volatile/issue_2627.tar.bz2

[...]

PR #2630, opened 2017-01-11
Improving post sorting.

felixfontein:
The current sorting of posts for `site.timeline`, `site.posts`, `site.all_posts`, `site.pages` and the lists of posts per taxonomy has some downsides:

  1. The sorting code is repeated in two different files (`nikola.py` and `taxonomies_classifier.py`) and cannot be replaced easily with monkey patching (you have to catch signals and post-process the lists and hope nobody relies on the order not changing since before you began post-processing).
  2. The sorting is done by priority, date, and the post's source path, and all of this reversed; it would make more sense to sort by the post's source path non-reversed, and even better to sort by the post's title (if possible because the language is known, i.e. for taxonomy post lists which are language-dependent) before resorting to the post's source path.

The current PR fixes both things:

  1. It provides a single function in the Nikola site object (which can be easily monkey-patched if someone wants a different sorting order) which sorts posts and returns a new, sorted list.
  2. All the above named places use this function.
  3. It doesn't reverse sort by the post's source path, and it does sort (in correct order) by the title in case priority and date match before resorting to the post's source path in case the language is known (which is the case for taxonomy post lists, as mentioned above).

This should fix the remaining problems in #2627 (i.e. category listings are not random anymore, but in reversed order as one would expect).

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1984,7 +1984,8 @@ class Nikola(object):
         # Next, flatten the hierarchy
         self.category_hierarchy = utils.flatten_tree_structure(root_list)
 
-    def sort_posts_chronologically(self, posts, lang=None):
+    @staticmethod
+    def sort_posts_chronologically(posts, lang=None):
         """Return sorted list of posts."""
         # Last tie breaker: sort by source path
         posts = sorted(posts, key=lambda p: p.source_path)
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1984,7 +1984,7 @@ class Nikola(object):
         # Next, flatten the hierarchy
         self.category_hierarchy = utils.flatten_tree_structure(root_list)
 
-    def sort_posts(self, posts, lang=None):
+    def sort_posts_chronologically(self, posts, lang=None):
         """Return sorted list of posts."""
         # Last tie breaker: sort by source path
         posts = sorted(posts, key=lambda p: p.source_path)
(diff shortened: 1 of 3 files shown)


nikola/nikola.py
```diff
@@ -1984,6 +1984,20 @@ def create_hierarchy(cat_hierarchy, parent=None):
         # Next, flatten the hierarchy
         self.category_hierarchy = utils.flatten_tree_structure(root_list)
 
+    def sort_posts(self, posts, lang=None):
```
Kwpolska: This is a name clash with `nikola.utils.sort_posts`. It should be named something distinctive, eg. `sort_posts_chronologically` (its main purpose)

Moreover, the sort should be done in place (that’s what we did before)

felixfontein:
The failing tests are due to the sorting order mattering in the invariance test: until now `/pages/bootstrap-demo.html` came after `/posts/welcome-to-nikola.html` (because of reverse sorting by post's source path), but now `"Bootstrap Demo"` comes before `"Welcome to Nikola"` (because of correct sorting by post's title).

nikola/nikola.py
felixfontein: It is intentionally not a static method, because it might want to use the site's config in the future.

Kwpolska: What for?

felixfontein: I first wanted to do it inplace, but that's kind of annoying with `natsort` (you need to use `natsort.natsort_keygen`, but I'm not sure whether that leads to the same order as `natsort.natsorted`). But if you insist I can change it back.

Kwpolska: There seems to be no way to end up with the unsorted list, so you can leave it as-is.

felixfontein: Because maybe someone wants an option to sort things differently. Just a precaution.

Kwpolska: That’s going too far into options-for-five-people territory.

felixfontein: Ok. Fixed.

felixfontein: Fixed the name clash and added a changelog entry.

felixfontein:
Thanks for improving the documentation!

PR #2634, opened 2017-01-14
Fix #2628 - allow meta generator to be diabled

adarshdec23:

PR #2635, opened 2017-01-14
Still missing from #2634.

felixfontein:


felixfontein:
I can't merge directly or override the missing reviews, so here we go ;)

Kwpolska:
(That would get done on release anyways.)

PR #2637, opened 2017-01-14
Fix #2537 - Build fails on bad image file

### LINKED ISSUE #2537: build fails on bad image file: OSError: cannot identify image file
I'm wondering if a try / except could be rigged up in the image_processing.py functions to better handle issues with image files?

I have an automated system which downloads images and builds pages with them embedded. I'd much prefer logging the error and continuing with the build process (with the image not included in the post), than noticing that my builds are crashing and having to fix the image file.

adarshdec23:
Catch OSError and return. 
Log the exception as a warning. The warning message is consistent with that in Pillow docs: 	

IOError – If the file cannot be found, or the image cannot be opened and identified.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Features
 Bugfixes
 --------
 
+* Catch OSError while opening a damaged image file. (Issue #2537)
+
 New in v7.8.3
 =============
 
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -97,7 +97,13 @@ class ImageProcessor(object):
         if not Image or os.path.splitext(src)[1] in ['.svg', '.svgz']:
             self.resize_svg(src, dst, max_size, bigger_panoramas)
             return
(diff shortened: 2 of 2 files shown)


Kwpolska:
:-1:

If you have a broken file, Nikola crashing *is* the correct behavior, because that file probably won’t be recognized by browsers either, and you will end up with garbage output. In my opinion, this can be left as-is for the user to fix themselves.

adarshdec23:
I see the value of what you're saying. 
Bad images should not reach the browser. So isn't it better to log an error and not process the image further. This ensures that the image does not reach the output and also users won't have to see an ugly back-trace.

Kwpolska:
A hard crash works better here IMO than just a warning and continuing with a build.

felixfontein:
How about improving the error message by printing it with `self.logger.error()` and then calling `sys.exit()`? Then the user won't see the ugly stack trace, and the build will still stop.

PR #2650, opened 2017-01-28
Compiling ipynb now looks for default jupyter configuration

knowsuchagency:
For those, like me, who use the [jupyter extensions](https://github.com/ipython-contrib/jupyter_contrib_nbextensions), specifically [python markdown](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/python-markdown), this makes it so Nikola will automatically look to see if users have edited the configuration files nbconvert normally looks at when compiling notebooks itself.

Thus if you have a markdown cell that jupyter normally converts to 2 from {{ 1 + 1 }} in your notebook, you'll see the former rather than the later.


nikola/plugins/compile/ipynb.py
```diff
@@ -82,6 +82,7 @@ def _compile_string(self, nb_json):
         if flag is None:
             req_missing(['ipython[notebook]>=2.0.0'], 'build this site (compile ipynb)')
         c = Config(self.site.config['IPYNB_CONFIG'])
+        c.update(get_default_jupyter_config())
```
Kwpolska: Do not create a one-off function in nikola.utils, instead put it in ipynb.py.

Kwpolska:
@knowsuchagency, could you please address the review comments?

ralsina:
I followed up on the requested changes in #2688, thanks a lot for the fix @knowsuchagency :-)

Closing this one.

Issue #2649, opened 2017-01-27
`<base href="">` generated incorrectly


OvermindDL1:
The `<base href>` in the `<head>` is being generated wrong on some pages.

For example, in my `/tags/index.html` file (via `conf.py` entry of `TAGS_INDEX_PATH = "tags/index"`) the base element is:
```html
<base href="http://mysite.com/tags%5Cindex.html">
```
The `%5C` is causing all relative links on this page to fail, no tag sections can be accessed from here.

```
$ nikola --version
Nikola v7.8.3
```

Kwpolska:
I cannot reproduce with v7.8.3 on macOS, this is what I get:

```html
<base href="https://example.com/tags/index.html">
```

But do you really need the `<base>` tags? Experience shows you don’t.

OvermindDL1:
> But do you really need the `<base>` tags? Experience shows you don’t.

Not really and I have it disabled, it was just a bug that I ran across and was able to reliably reproduce.  Feel free to `git clone https://github.com/OvermindDL1/overminddl1.github.io.git`, adjust the conf.py to re-enable the base element, then `nikola build`, the `tags/index.html` that is generated is wrong.  Also, why do I have to mark it as `tags/index` to generate a `tags/index.html`?  The full name would be better in the config.

ralsina:
```
TAGS_INDEX_PATH = "tags/index"
```

That doesn't look right to me. TAGS_INDEX_PATH is supposed to be either a filename or a full path relative to TAG_PATH

Kwpolska:
I still can’t reproduce with your site:

```html
<base href="http://blog.overminddl1.com/tags/">
```

Make sure lxml and other dependencies of Nikola are up-to-date.

OvermindDL1:
> That doesn't look right to me. TAGS_INDEX_PATH is supposed to be either a filename or a full path relative to TAG_PATH

That was my thinking, you'd think it would be a full path so I should put `tags/index.html`, however if I do that then it generates a `tags/index.html.html` in the output directory...

> Make sure lxml and other dependencies of Nikola are up-to-date.

Hmm, I did that this morning, and just tried again now but it is currently saying "Requirement already up-to-date", I've tried to upgrade every package I see through `pip` that I have, but no change.

I am on Windows here (though I usually use linux), perhaps that is affecting things (maybe it thinks the `/` is an invalid path-sep on Windows so it is encoding it instead of leaving it alone as it should?).

[...]

PR #2652, opened 2017-01-29
Preventing .html to be added when it shouldn't be.

felixfontein:
Fixes bug described in #2649, when `.html` is added to `TAGS_INDEX_PATH`.

### DIFF
--- a/nikola/plugins/misc/taxonomies_classifier.py
+++ b/nikola/plugins/misc/taxonomies_classifier.py
@@ -240,10 +240,12 @@ class TaxonomiesClassifier(SignalHandler):
                 path = ['rss']
             elif len(path) == 0 or append_index == 'always':
                 path = path + [os.path.splitext(self.site.config['INDEX_FILE'])[0]]
+            elif len(path) > 0 and append_index == 'never':
+                path[-1] = os.path.splitext(path[-1])[0]
             path[-1] += force_extension
         elif (self.site.config['PRETTY_URLS'] and append_index != 'never') or len(path) == 0 or append_index == 'always':
             path = path + [self.site.config['INDEX_FILE']]
-        else:
+        elif append_index != 'never':
             path[-1] += '.html'
         # Create path
         result = [_f for _f in [self.site.config['TRANSLATIONS'][lang]] + path if _f]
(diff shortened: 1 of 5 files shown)


Kwpolska:
That will break the five sites which used `tags/index` to work around the bug.

felixfontein:
So, what do you suggest?

ralsina:
I know it's ugly but ... maybe check if it already ends with ".html" and if so, don't?

Or maybe check if the user has configured something like "tags/index" and in that case do a fix to "tags/index.html", show a warning?

felixfontein:
Changing input parameters sounds bad to me. What if someone really wants `output/tag/index` without any file extension? It won't be possible anymore.

How about printing a warning if the path ends with `/index`? People can then choose to ignore it, or fix the problem when they notice the warning.

Kwpolska:
No, nobody would want `output/tag/index`. It’s nonsense.

felixfontein:
How about the solution in 590736c18efb1f4ffdf5fbb8ae342d630b791811?

nikola/plugins/misc/taxonomies_classifier.py
```diff
@@ -240,10 +240,12 @@ def _postprocess_path(self, path, lang, append_index='auto', dest_type='page', p
                 path = ['rss']
             elif len(path) == 0 or append_index == 'always':
                 path = path + [os.path.splitext(self.site.config['INDEX_FILE'])[0]]
+            elif len(path) > 0 and append_index == 'never':
+                path[-1] = os.path.splitext(path[-1])[0]
```
Kwpolska: Is this fixing outdated code that does provides eg. `['rss.xml']` instead of `['rss']`?

felixfontein: You mean like `sections.py` which had `return result + ['rss'], 'never'`, which now is `return result + ['rss.xml'], 'never'`?

Kwpolska: I’ll ask another way: why do you need splitext here and you didn’t in the past?

felixfontein: Because in the past, part of the code assumed that `'never'` never had the extension there, while other parts assumed that it had. This code assumed it wasn't there so it could simply add it. Now `'never'` means "change extension if necessary", so now it needs `splitext` to get rid of the old one.

Kwpolska: Got it, merging.

Issue #2660, opened 2017-02-14
Markdown compiler has performance issues


living180:
As noted in https://getnikola.com/blog/markdown-can-affect-performance.html, the markdown compiler has very poor performance when compiling a large number of markdown files.

PR #2661, opened 2017-02-14
Improve markdown compiler performance

living180:
This PR fixes issue #2660.

The first commit eliminates pathological performance when compiling a large number of markdown files at one time but preventing the markdown extensions list from growing each time a markdown file was compiled.  On my machine this reduced the time required to compile 1000 markdown files from over 11 minutes to under 55 seconds.

The second commit provides a further incremental improvement by creating and reusing a single Markdown object instead of creating a separate Markdown object for each file that is compiled. This further reduces the compilation time for 1000 markdown files to under 34 seconds.

### DIFF
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -47,7 +47,8 @@ from nikola.utils import makedirs, req_missing, write_metadata
 class ThreadLocalMarkdown(threading.local):
     """Convert Markdown to HTML using per-thread Markdown objects.
 
-    See discussion in #2661."""
+    See discussion in #2661.
+    """
 
     def __init__(self, extensions):
         """Create a Markdown instance."""
--- a/nikola/plugins/compile/markdown/__init__.py
+++ b/nikola/plugins/compile/markdown/__init__.py
@@ -30,11 +30,12 @@ from __future__ import unicode_literals
 
 import io
 import os
+import threading
 
 try:
-    from markdown import markdown
+    from markdown import Markdown
 except ImportError:
-    markdown = None  # NOQA
+    Markdown = None  # NOQA
     nikola_extension = None
     gist_extension = None
     podcast_extension = None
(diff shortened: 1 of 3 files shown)


nikola/plugins/compile/markdown/__init__.py
```diff
@@ -49,34 +49,38 @@ class CompileMarkdown(PageCompiler):
...
             self.config_dependencies.append(plugin_info.name)
-            self.extensions.append(plugin_info.plugin_object)
+            extensions.append(plugin_info.plugin_object)
             plugin_info.plugin_object.short_help = plugin_info.description
 
-        self.config_dependencies.append(str(sorted(site.config.get("MARKDOWN_EXTENSIONS"))))
+        site_extensions = self.site.config.get("MARKDOWN_EXTENSIONS")
+        self.config_dependencies.append(str(sorted(site_extensions)))
+        extensions.extend(site_extensions)
+        self.markdown = Markdown(extensions=extensions, output_format="html5")
```
Kwpolska: This will crash if `Markdown` does not exist.

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -49,35 +49,38 @@ class CompileMarkdown(PageCompiler):
...
             self.config_dependencies.append(plugin_info.name)
-            self.extensions.append(plugin_info.plugin_object)
+            extensions.append(plugin_info.plugin_object)
             plugin_info.plugin_object.short_help = plugin_info.description
 
         site_extensions = self.site.config.get("MARKDOWN_EXTENSIONS")
         self.config_dependencies.append(str(sorted(site_extensions)))
-        self.extensions.extend(site_extensions)
+        extensions.extend(site_extensions)
+        self.markdown = Markdown(extensions=extensions, output_format="html5")
```
felixfontein: Are `Markdown` objects thread-safe? (You can tell doit to process tasks in parallel using threads, so it could be that two Markdown posts are compiled simultaneously.) I somehow doubt this as you call `self.markdown.reset()` after `self.markdown.convert(data)`. If this is the case, you should better use thread-local storage for the `Markdown` object.
(That would also indirectly solve Kwpolska's remark.)

ralsina:
I agree, either make it thread local or protect it with a mutex.

felixfontein:
Since this is about performance, making it thread-local sounds like a much better choice than a mutex :)

nikola/plugins/compile/markdown/__init__.py
living180: Good catch, thanks.

living180:
@Kwpolska wrote
> Any benchmarks?

In the PR description, I included some timings I measured (basically more than 11 minutes before my changes, then under 55 seconds with just the first commit, and finally under 34 seconds with both commits).  Are you looking for something more than that?  If so, could you please describe what you would like to see?  I'm happy to try to accomodate

Kwpolska:
@living180 Ah, that’s good enough. But please address all our comments so we can proceed with this.

living180:
Updated version - should not fail ungracefully if the `markdown` module is not installed, and is now thread-safe for parallel builds.  (Testing with `nikola build -n 4` reduces the runtime to 14s on my machine.)

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -43,40 +44,56 @@
 from nikola.utils import makedirs, req_missing, write_metadata
 
 
+class ThreadLocalMarkdown(threading.local):
+    """Convert to markdown using per-thread Markdown objects."""
+
+    def __init__(self, extensions):
+        self.markdown = Markdown(extensions=extensions, output_format="html5")
```
felixfontein: I'm not sure whether this is the correct way to use `threading.local`. Every thread creates a copy of this initial value, but depending on how this copy is created (maybe even none at all, by just copying the reference to the `Markdown` object) the resulting object is not thread-safe.

It might be better to set `self.markdown` to `None`, and to check `self.markdown` in `convert()` and initialize it there when its current value is `None`.

felixfontein: Ok, after reading some more about this, I realize I'm wrong and your approach is correct. `threading.local` is a really strange beast, it seems. (For anyone interested, it seems the only "real" documentation can be found in the docstring here: https://svn.python.org/projects/python/trunk/Lib/_threading_local.py.)

Kwpolska: Here’s some test code:

```py
import threading
from markdown import Markdown

class ThreadLocalMarkdown(threading.local):
    """Convert to markdown using per-thread Markdown objects."""

    def __init__(self, extensions):
        self.markdown = Markdown(extensions=extensions, output_format="html5")
        self.counter = 0

    def convert(self, data):
        """Convert data to HTML and reset internal state."""
        result = self.markdown.convert(data)
        self.markdown.reset()
        self.counter += 1
        print(self.counter)
        return result

    def show(self):
        print(self.counter)

converter = ThreadLocalMarkdown([])

def run_convert():
    converter.convert("Hello, **world!**")

threads = [threading.Thread(None, run_convert) for i in range(10)]
[t.start() for t in threads]
converter.show()
```

says a bunch of ones and a zero at the end, as expected.

felixfontein: With an integer it wouldn't be a problem anyway, because integers are immutable. If what I first thought would happen would really happen, you'd need a list or some other mutable object to test for it.

After reading the docstring for `_threading_local`, I'm happy with the code. I still think that `threading.local` viloates the principle of least astonishment.

Kwpolska: I’m taking this as an approval and merging.

Kwpolska:
Thanks for contributing to Nikola! :tada:

nikola/plugins/compile/markdown/__init__.py
felixfontein: Yep, I'm happy with this PR. I'm just not happy how `threading.local` was designed ;)

PR #2663, opened 2017-02-15
Fix #2651 -- append file name if new_post gets directory name

### LINKED ISSUE #2651: allow specifying of section only via path, then let nikola append filename
There is currently no easy way to specify a section for a new post.
`nikola new_post posts/sectionname/here-i-have-to-think-of-the-slug-of-the-title-already.md` and then specifying the Title in nikola seems error prone. It would be much nicer if one could specify the section-path only and the filename would then be appended: `nikola new_post posts/sectionname`.

Kwpolska:
cc @t-animal


felixfontein:
Wouldn't it be better to only treat the argument as a path if it ends with `os.sep` or `/`? Otherwise it is impossible to tell what `nikola new_post bla` does without knowing whether `posts/bla` is a directory or something else (a file or not even existing).

Kwpolska:
Most people would know what their directory structure is, and Nikola won’t create new directories for you.

PR #2664, opened 2017-02-15
New require_all_tags for post-list directive

andredias:
The default behaviour of post-list directive is to show posts that have at least one of the tags. It could be better if it could be more restrictive.

I suggest this new option 'require_all_tags'. If declared, the post-list will show only posts that have all tags.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,8 +4,10 @@ New in master
 Features
 --------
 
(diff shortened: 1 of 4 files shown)


andredias:
I've just noticed that I have used the wrong branch. Should I resend it on the right one?

andredias:
I'll fix this.

PR #2665, opened 2017-02-16
add option require_all_tags to post-list directive.

andredias:
If declared it changes the tag filter behaviour to show only posts that have *all* specified tags.

### DIFF
--- a/nikola/plugins/compile/rest/post_list.py
+++ b/nikola/plugins/compile/rest/post_list.py
@@ -31,6 +31,7 @@ from __future__ import unicode_literals
 import os
 import uuid
 import natsort
+import operator
 
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
@@ -254,9 +255,9 @@ def _do_post_list(start=None, stop=None, reverse=False, tags=None, require_all_t
     if tags:
         tags = {t.strip().lower() for t in tags.split(',')}
         if require_all_tags:
-            compare = lambda x, y: x.issubset(y)
+            compare = set.issubset
         else:
-            compare = lambda x, y: x & y
+            compare = operator.and_
         for post in timeline:
             post_tags = {t.lower() for t in post.tags}
             if compare(tags, post_tags):
--- a/nikola/plugins/compile/rest/post_list.py
+++ b/nikola/plugins/compile/rest/post_list.py
@@ -261,6 +261,8 @@ def _do_post_list(start=None, stop=None, reverse=False, tags=None, require_all_t
             post_tags = {t.lower() for t in post.tags}
             if compare(tags, post_tags):
                 filtered_timeline.append(post)
+    else:
+        filtered_timeline = timeline
 
     if sort:
         filtered_timeline = natsort.natsorted(filtered_timeline, key=lambda post: post.meta[lang][sort], alg=natsort.ns.F | natsort.ns.IC)
--- a/nikola/plugins/compile/rest/post_list.py
+++ b/nikola/plugins/compile/rest/post_list.py
@@ -104,6 +104,10 @@ class PostList(Directive):
         Filter posts to show only posts having at least one of the ``tags``.
         Defaults to None.
 
+    ``require_all_tags`` : flag
+        Change tag filter behaviour to show only posts that have all specified ``tags``.
+        Defaults to False.
+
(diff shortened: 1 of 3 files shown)


nikola/plugins/compile/rest/post_list.py
```diff
@@ -246,18 +251,16 @@ def _do_post_list(start=None, stop=None, reverse=False, tags=None, categories=No
...
+    if tags:
+        tags = {t.strip().lower() for t in tags.split(',')}
+        if require_all_tags:
+            compare = lambda x, y: x.issubset(y)
+        else:
+            compare = lambda x, y: x & y
+        for post in timeline:
+            post_tags = {t.lower() for t in post.tags}
+            if compare(tags, post_tags):
+                filtered_timeline.append(post)
```
Kwpolska: If tags are not specified, `filtered_timeline` is empty.

tritium21: technically, if you are making a post list, and require all the tags to be present, and don't offer any tags, you are semantically asking for a list of no posts, so... not incorrect behavior >.>

Kwpolska: Uh, no? If no `tags` are specified, then `filtered_timeline` stays empty, which means the post list will also be empty. And not every post list is tag-based.

tritium21: This is already in an if tags: conditional; that case will never actually trigger, or I cant see how it will trigger (now that i see it in more context)

Kwpolska: That’s not what I’m talking about. Read the old code: if no tags were specified, `filtered_timeline` would just be filled with all posts from `timeline`, otherwise there is a way for `continue` to fire and not put a post in the timeline. In this code, `filtered_timeline` may be appended to only if there are tags specified. But if there are no tags specified, the entire block does not fire, and `filtered_timeline` stays empty. After the tags block, `filtered_timeline` is used in place of `timeline`. So, to reiterate: `if not tags: filtered_timeline == []` and post list is *empty*.

tritium21: Welp, now I feel a little silly.  Yeah, this needs an else clause on if tags: ... else: filtered_timeline = timeline[:]  # or just = timeline

andredias: Ok. I'll fix that.

Kwpolska: Copying the list does not seem necessary, just setting it to `timeline` is enough.

nikola/plugins/compile/rest/post_list.py
```diff
@@ -246,18 +251,16 @@ def _do_post_list(start=None, stop=None, reverse=False, tags=None, categories=No
...
-                    cont = False
-
-            if cont:
-                continue
-
-        filtered_timeline.append(post)
+    if tags:
+        tags = {t.strip().lower() for t in tags.split(',')}
+        if require_all_tags:
+            compare = lambda x, y: x.issubset(y)
```
tritium21: `def compare(x, y)` is preferred over assigning lambdas (flake8 test)

Kwpolska: Personally, I’d be fine with `# NOQA` for the first function and `compare = operator.and_` for the second.

tritium21: sounds good to me

Kwpolska: Or wait, the first one can be done with `compare = set.issubset`.

andredias: yeah, but it would require defining two additional functions:

```python
if require_all_tags:
    def compare(x, y):
        return x.issubset(y)
else:
    def compare(x, y):
        return x & y
```
I don't think it is worth it. "Beautiful is better than ugly."

tritium21: you could do

```
    if require_all_tags:
        compare = set.issubset
    else:
        compare = operator.and_

andredias: Ok. Using operator and set.issubset seems more elegant.

Kwpolska:
Thanks for contributing! I simplified the option detection a little in commit 5c2cd6f0.

PR #2668, opened 2017-02-16
fix #2666 | The post where the post-list directive is declared shouldn't be listed

### LINKED ISSUE #2666: Post-list plugin shouldn't list the post where the post-list directive is declared
Post: `posts/xyz.rst` with tags: `version control`, `mercurial`

At the end, I declare:

```rst
.. post-list::
    :tags: version control, mercurial
```

The problem is that the post `xyz.rst` is listed at the post-list. This is redundant.

I have an idea how to fix this, but I need some feedback about the changes I'd make.
I know the post path being compiled is stored at `self.state.document.settings._nikola_source_path`.
I could pass this piece of information to `_do_post_list` via the `filename` parameter and then remove the post from the `timeline`. However, I'm not sure about the impact in `self_post`. Would it still be necessary if the post is removed?

Any suggestions?

andredias:
Fix #2666

### DIFF
--- a/nikola/plugins/compile/rest/post_list.py
+++ b/nikola/plugins/compile/rest/post_list.py
@@ -178,9 +178,10 @@ class PostList(Directive):
         template = self.options.get('template', 'post_list_directive.tmpl')
         sort = self.options.get('sort')
         date = self.options.get('date')
+        filename = self.state.document.settings._nikola_source_path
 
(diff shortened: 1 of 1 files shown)


ralsina:
Changelog, dudes :-)

Kwpolska:
I’ll do it.

Issue #2659, opened 2017-02-13
Support posts without source file


remram44:
#1708 added a PostScanner plugin that allows pluggable post sources, noting that "the only currently existing plugin of this kind reads them from disk".

However all info is read from Post.source_path across Nikola, so it seems that having a file backing each post is the only way to go.

It would be nice to be able to create a Post object with metadata & source and feed them to Nikola rather than autogenerating files (and in fact, if that is what you have to do, no PostScanner is required).

felixfontein:
Well, that would require some work on the `Post` class. There's already an issue for that: #1709.

You can of course also create your own subclass of `Post` and do in there whatever you want, or provide a "mockup" which behaves as Nikola expects it. It is just more complicated without the refactoring in #1708.

remram44:
The problem is that a lot of code, outside of the Post class, open the `source_path` file directly.

felixfontein:
Actually, most of the times `source_path` is only used to have something to sort posts by, or to show in error/status messages for the user of Nikola.

There is only one place (outside `Post`) which assumes that `source_path` is actually a real file on disk: namely `sources.py`, the plugin which tries to copy the source files to the output when `COPY_SOURCES` is `True`. So as long as you don't have `COPY_SOURCES` enabled and put something sensible into `source_path` which allows to identify the generated post, there's no need for `source_path` to actually point to an existing file.

remram44:
It is definitely doable 👍

I'll take a crack at this myself if I can find the time.

tritium21:
Perhaps, one should consider adding a member to the Post object strictly for sorting, defaulting to the path with the native post scanner, and copy source be adjusted to accept None as a valid option for post path (and therefor not copy the source, and not try and link to the source).  Its a bigger diff, but a more flexible one.

felixfontein:
It would also be good if `copy_sources` wouldn't have to use some logic to guess the correct source paths for translated posts. Currently it uses `post.source_path` together with `utils.get_translation_candidate`. It would probably be best if there's a function in `Post` which returns the source path for a given language, or `None` if there is no source (`copy_sources` could then ignore it).

Actually, I just noticed that there is already such a function: `translated_source_path`. I'll make a PR to use it (and to allow return value `None`).

remram44:
Having (overridable) Post methods that return file-like objects, instead of
externally deriving filenames and opening the file, would go a long way
towards extensibility!

[...]

PR #2669, opened 2017-02-16
Don't use Post.source_path directly to determine translated post's source file name.

felixfontein:
Allowing Post.translated_source_path() to return None.

Helping with one aspect of #2659.

### DIFF
--- a/nikola/plugins/task/sources.py
+++ b/nikola/plugins/task/sources.py
@@ -61,12 +61,8 @@ class Sources(Task):
                     # do not publish PHP sources
                     if post.source_ext(True) == post.compiler.extension():
                         continue
-                    source = post.source_path
-                    if lang != kw["default_lang"]:
-                        source_lang = utils.get_translation_candidate(self.site.config, source, lang)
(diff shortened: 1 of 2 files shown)


nikola/plugins/task/sources.py
```diff
@@ -61,12 +61,8 @@ def gen_tasks(self):
                     # do not publish PHP sources
                     if post.source_ext(True) == post.compiler.extension():
                         continue
-                    source = post.source_path
-                    if lang != kw["default_lang"]:
-                        source_lang = utils.get_translation_candidate(self.site.config, source, lang)
-                        if os.path.exists(source_lang):
-                            source = source_lang
-                    if os.path.isfile(source):
```
Kwpolska: Why was all this code removed? Was it all redundant?

felixfontein: This code essentially does what `post.translated_source_path(lang)` is supposed to be doing. Even worse, it uses a slightly different logic, which could cause incorrect source files to be copied.

PR #2670, opened 2017-02-16
Avoid replicating compiler detection logic.

felixfontein:
The compiler is already decided in the `scan_posts` plugin and shouldn't be re-decided here.


felixfontein:
(Also for #2659.)

Issue #1709, opened 2015-05-09
Refactor the Post object


ralsina:
The Post object is way too complex at this point.

1) post.py is over 1000 LOC
2) `Post.__init__` is over 150 LOC
3) The `Post` class has 36 methods
4) We have 5 (five!) different ways to extract metadata

So, proposed solutions:
- [ ] Create an abstract Post class
- [ ] Move some things into plugins [1]
- [x] Deprecate stuff (example: old-old metadata!)
- [x] Move some things out of core directly (encryption)
- [x] Maybe refactor metadata extraction into compiler plugins? (Get docinfo from docutils, etc) (~~This is #618~~ no it's not)

[1] For example, estimated read time could be implemented as metadata. We could create a mechanism for adding metadata to posts. Could be separated into plugins. Those will use the Post class to add data to it.  Makes for a very simple way to extend the Post class, without commiting to core.

felixfontein:
How about two new plugin classes for points 2 and 4:
 - post preprocessing:
   - they get the `Post` object and `lang` and can add additional metadata or dependencies
 - post postprocessing:
   - receives the compiler's output (or the previous postprocessor's output), the `Post` object and `lang`
   - the encryption plugin is a natural candidate for this

Perhaps in the following form:
``` .py
class PostPreprocessor(BasePlugin):
    """Preprocess a Post object before compiling."""

    name = "dummy_post_preprocessor"

    def preprocess(self, post, lang):
        """Preprocess the Post object before compiling."""
        raise NotImplementedError()

class PostPostprocessor(BasePlugin):
    """Postprocesses a compiled Post."""

    name = "dummy_post_postprocessor"

    def postprocess(self, data, post, lang):
        """Postprocess the compiled post.

        Obtains the compiled post ``post`` in language ``lang`` as ``data``
        as a decoded string. Must return the new post contents as a string.
        """
        raise NotImplementedError()
```
What is to be decided is whether the postprocessing will be done before or after the `compiled` signal. (Obviously, one could implement postprocessing also a `SignalHandler` listening to the `compiled` signal. It doesn't receive `lang`, though. And I think having an own plugin category is cleaner.)

felixfontein:
The post postprocessing plugin type isn't really needed, this can be handled by a signal handler listening for `compiled` (emitted by `Post.compile`).

gwax:
[...]

PR #1708, opened 2015-05-09
Refactor scanposts (Issue #1700)

ralsina:
See Issue #1700 for details.

This branch should not cause any visible change in behaviour :-)

PR #2671, opened 2017-02-16
Creating abstract base class BasePost.

felixfontein:
First shot at #2659 and #1709. Tried to move everything which is not input file dependent into base class.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1751,7 +1751,7 @@ class Nikola(object):
                 # PyRSS2Gen's pubDate is GMT time.
                 'pubDate': (post.date if post.date.tzinfo is None else
                             post.date.astimezone(dateutil.tz.tzutc())),
-                'categories': post._tags.get(lang, []),
+                'categories': post.tags_for_language(lang),
                 'creator': post.author(lang),
                 'guid': post.guid(lang),
             }
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -75,8 +75,8 @@ TEASER_REGEXP = re.compile('<!--\s*TEASER_END(:(.+))?\s*-->', re.IGNORECASE)
 _UPGRADE_METADATA_ADVERTISED = False
 
 
-class Post(object):
-    """Represent a blog post or site page."""
+class BasePost(object):
+    """Abstract base class for representing a blog post or site page."""
 
     def __init__(
         self,
@@ -91,9 +91,8 @@ class Post(object):
     ):
         """Initialize post.
 
-        The source path is the user created post file. From it we calculate
-        the meta file, as well as any translations available, and
-        the .html fragment file path.
+        The source path is an identifier shown to the user to identify the post
+        in case of errors, or used as a last resort tie-breaking for sorting posts.
 
         destination_base must be None or a TranslatableSetting instance. If
         specified, it will be prepended to the destination path.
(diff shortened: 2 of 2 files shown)


felixfontein:
@Kwpolska: so what's bad about refactoring the `Post` class?

Kwpolska:
@felixfontein Overengineering.

felixfontein:
Do you really think that refactoring the `Post` class in general is overengineering? Or do you think it's a bad idea to support something like #2659?

Kwpolska:
I think that `BasePost` and the possibility to have different subclasses of Post is a bad idea, and a waste of time and resources. #2659 is a bit of a problematic thing. Nikola is tied to writing files to disk in many places, and you would need to mess with dependencies if you changed the meaning of `source_path` — and then there might be plugins which depend on that being a real file on disk, and would need a lot of special casing.

But perhaps the file-less posts could write to `cache/`, where a file has to exist anyway?

cc @remram44

felixfontein:
Aside from the point that some other plugins might assume that `source_path` is a real file on disk (which might be a problem, but which shouldn't be a reason not to do this), Nikola doesn't need the post files to exist on disk except for `scan_post.py` and the functions that I didn't move to `BasePost`. If you want a pure in-memory post (generated by whatever source), you can simply put the post's content into an `uptodate` dependency and emit that, and write it to disk in the overridden `compile` method. That would be a pretty simple implementation of another subclass of `BasePost`, and something rather impossible (without a lot of copy'n'pasting) with the current `Post` class from Nikola master; the main problem being all the logic done in `__init__` which cannot be moved out now without causing incompatibility with existing plugins creating `Post` objects.

Creating artificial post files in `cache/` is something I don't really like. Especially because then you have to make sure there aren't any collisions or other files which will be interpreted by Nikola as other-language versions of the same post, or as metadata. (If you have a post file containing metadata, and for some reason a file with the same name but `.meta` at the end is there containing metadata, I think Nikola will use that one and threat the metadata in the post file as content.) This would be a *really* messy solution.

remram44:
If you think supporting file-less posts is not something that Nikola wants to support, I totally understand. You are building on top of doit, and most people use files. I respect wanting to keep the codebase simple.

I only requested #2659 because work on a PostScanner plugin had already been introduced in #1708, so I thought this was a direction you were considering -- after all, there is not much that can be done in terms of plugins if all you can do is readdir() a different directory.

I am in no way blocked by #2659 personally.

Kwpolska:
> There is not much that can be done in terms of plugins if all you can do is readdir() a different directory.

You can, but a post scanner is usually not enough to do it. For example, our Plugins site uses [pkgindex_scan](https://plugins.getnikola.com/v7/pkgindex_scan/) and [pkgindex_compiler](https://plugins.getnikola.com/v7/pkgindex_compiler/) — the scan plugin creates Post elements from README.md files found in the filesystem, but then the scanner adds meta data from .plugin files and handles compiling those README files. I don’t think there is a hard dependency on `source_path` being unique, so maybe this architecture (custom scanner + compiler) will be enough for you?

felixfontein:
Well, file-less posts are no problem even with doit, as long as you can efficiently create an `uptodate` object every run of Nikola which describes whether the source is still up-to-date. For an RSS feed, that should be no problem (assuming that the feed file isn't huge and parsing it takes a considerable amount of time).

The only way to implement something like this in Nikola is at the moment to completely rewriting the `Post.__init__` method when subclassing the `Post` class, as its current implementation assumes there are files on disk.

Actually, there's one trick which would simplify the procedure without creating an abstract base class: move all the file-specific methods of the constructor into a method called by the constructor. Then subclasses of `Post` could override that method (and other methods, as they like to). This would still require to repeat some code (especially in the functions retrieving dependencies), but it would be much less bad than the current situation.

I'd still prefer an abstract base class, though. And I still disagree that this is overengineering; the `Post` class is one of the least modular/extensible parts of Nikola.

gwax:
I think the principle of #1709 is sound but I'm not sure this is the right architectural direction. I'm on the fence about #2659.

It seems to me, if we're going to break up Post, we should start by pulling out some of the complex functionality first, rather than starting by abstracting everything.

Looking at `Post.__init__` and some of the other code, it seems like creating a `PostMetadata` class and encapsulating the metadata might be a better solution than abstracting the metadata processing.

felixfontein:
I guess we should then close this and continue the discussion in #1709. (Also related is #2830.)

PR #2674, opened 2017-02-19
[WIP] Basic support for itunes rss tags

anderbubble:
I'm working on enhancing Nikola to support publishing a podcast RSS feed. Most of this work is going into a "postcast" plugin I'm working on, and will submit or publish eventually; but some of what I'm trying to do requires support in Nikola itself.

This PR aims to support [iTunes-specific tags](https://help.apple.com/itc/podcasts_connect/#/itcb54353390) in RSS feeds.

Other related PRs I'm working on:

* #2676 (copyright tag in rss feeds)
* #2675 (explicit / static guids)

I'd also be interested in direction for where and how to publish a plugin, when that's ready.

## todo

* remove references to postcast

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1715,9 +1715,7 @@ class Nikola(object):
                 for tag in ('subtitle', 'duration', 'explicit'):
                     key = 'itunes_{}'.format(tag)
                     args[key] = post.meta[lang].get(key)
-                args['itunes_author'] = (
-                    post.meta[lang].get('itunes_author')
-                    or post.author(lang))
+                args['itunes_author'] = post.meta[lang].get('itunes_author') or post.author(lang)
                 args['itunes_summary'] = post.meta[lang].get('itunes_summary') or data
                 if 'itunes_image' in post.meta[lang]:
                     args['itunes_image'] = self.url_replacer(post.permalink(), post.meta[lang]['itunes_image'], lang, 'absolute')
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1712,15 +1712,15 @@ class Nikola(object):
                 rss_obj.rss_attrs["xmlns:dc"] = "http://purl.org/dc/elements/1.1/"
 
             if itunes:
+                for tag in ('subtitle', 'duration', 'explicit'):
+                    key = 'itunes_{}'.format(tag)
+                    args[key] = post.meta[lang].get(key)
                 args['itunes_author'] = (
                     post.meta[lang].get('itunes_author')
                     or post.author(lang))
-                args['itunes_subtitle'] = post.meta[lang].get('itunes_subtitle')
                 args['itunes_summary'] = post.meta[lang].get('itunes_summary') or data
                 if 'itunes_image' in post.meta[lang]:
-                    args['itunes_image'] = urljoin(self.config['BASE_URL'], post.meta[lang]['itunes_image'])
-                args['itunes_duration'] = post.meta[lang].get('itunes_duration')
-                args['itunes_explicit'] = post.meta[lang].get('itunes_explicit')
+                    args['itunes_image'] = self.url_replacer(post.permalink(), post.meta[lang]['itunes_image'], lang, 'absolute')
 
             if enclosure:
                 # enclosure callback returns None if post has no enclosure, or a
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1739,6 +1739,7 @@ class Nikola(object):
         if itunes:
             rss_obj.rss_attrs["xmlns:itunes"] = "http://www.itunes.com/dtds/podcast-1.0.dtd"
             rss_obj.itunes_author = self.config['BLOG_AUTHOR'](lang)
+            rss_obj.itunes_name = self.config['BLOG_AUTHOR'](lang)
             rss_obj.itunes_email = self.config['BLOG_EMAIL']
             rss_obj.itunes_summary = description
             if 'POSTCAST_IMAGE' in self.config:
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1649,7 +1649,7 @@ class Nikola(object):
 
     def generic_rss_renderer(self, lang, title, link, description, timeline, output_path,
                              rss_teasers, rss_plain, feed_length=10, feed_url=None,
-                             enclosure=_enclosure, rss_links_append_query=None):
+                             enclosure=_enclosure, rss_links_append_query=None, itunes=None):
         """Take all necessary data, and render a RSS feed in output_path."""
         rss_obj = utils.ExtendedRSS2(
             title=title,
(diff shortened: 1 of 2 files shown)


nikola/utils.py
```diff
@@ -1272,13 +1286,66 @@ def publish_extensions(self, handler):
             })
             handler.endElement("atom:link")
 
+        if self.itunes_author:
```
Kwpolska: You could duplicate less code, eg. with `getattr` or some iTunes data dictionary.

Kwpolska:
Looks good (apart from a lot of duplicated code). Please update CHANGES.txt, AUTHORS.txt, and the manual.

nikola/utils.py
```diff
@@ -1255,6 +1255,20 @@ class ExtendedRSS2(rss.RSS2):
...
+                 itunes_explicit=None, **kwargs):
+        self.itunes_author = itunes_author
+        self.itunes_summary = itunes_summary
+        self.itunes_name = itunes_name
+        self.itunes_email = itunes_email
+        self.itunes_image = itunes_image
+        self.itunes_categories = itunes_categories
+        self.itunes_explicit = itunes_explicit
+
+        return rss.RSS2.__init__(self, **kwargs)
```
Kwpolska: Don’t use `return` here, instead call `rss.RSS2.__init__` at the start of this function. (old-style class in Python 2, so `super()` does not work(

anderbubble:
First, thank you for reviewing this pull request!

I returned `__init__` from `__init__` because the code for `Item` already did.[^1] Should I clean that up at the same time?

As for the duplicated code: can you provide a pointer to somewhere else in the codebase that does something similar better? (It'd just be nice to have an example of what idiom is considered most normal in this codebase.)

I'm also not happy with my handling of the `Image` url. Any suggestions for how to do it better?

[^1]: https://github.com/getnikola/nikola/pull/2674/files#diff-cca99a67a7b16498786d42c7f2f89444L1283

Kwpolska:
1. Yes, you can do that as well.
2. I don’t think we have anything like this. Here’s one example:

```py
itunes = {'itunes:subtitle': 'foo'}
# in publish_extensions
for name in ['itunes:author', 'itunes:subtitle', 'itunes:summary', 'itunes:duration']:
    if name in itunes:
        handler.startElement(name, {})
        handler.characters(itunes[name])
        handler.endElement(name)
# special handling for image and explicit stays
```

3. You could ask `url_replacer` with `url_type='absolute'`.

nikola/nikola.py
```diff
@@ -1649,7 +1649,7 @@ def apply_shortcodes(self, data, filename=None, lang=None, with_dependencies=Fal
 
     def generic_rss_renderer(self, lang, title, link, description, timeline, output_path,
                              rss_teasers, rss_plain, feed_length=10, feed_url=None,
-                             enclosure=_enclosure, rss_links_append_query=None):
+                             enclosure=_enclosure, rss_links_append_query=None, itunes=None):
```
felixfontein: Souldn't the default value for `itunes` be `False` and not `None`? After all, it is only used like a boolean.

anderbubble:
It is my habit to use `None` for a default value so that the underlying
code can distinguish between "unspecified" and "explicitly false."

But f that's counter to expectation here, I'm fine to change it.

Kwpolska:
`None` is good in this case IMO.

anderbubble:
I see now that I've added explicit references to my "postcast" plugin in this Nikola core PR; so I'm going to have to factor that out.

nikola/nikola.py
```diff
@@ -1711,6 +1711,17 @@ def generic_rss_renderer(self, lang, title, link, description, timeline, output_
...
+            if itunes:
+                for tag in ('subtitle', 'duration', 'explicit'):
+                    key = 'itunes_{}'.format(tag)
+                    args[key] = post.meta[lang].get(key)
+                args['itunes_author'] = (
+                    post.meta[lang].get('itunes_author')
+                    or post.author(lang))
+                args['itunes_summary'] = post.meta[lang].get('itunes_summary') or data
+                if 'itunes_image' in post.meta[lang]:
+                    args['itunes_image'] = self.url_replacer(post.permalink(), post.meta[lang]['itunes_image'], lang, 'absolute')
```
Kwpolska: Just keep in mind that the permalink won’t affect parsing of `itunes_image` paths, hey should be relative to site root (in your posts).

nikola/nikola.py
```diff
@@ -1711,6 +1711,17 @@ def generic_rss_renderer(self, lang, title, link, description, timeline, output_
             if post.author(lang):
                 rss_obj.rss_attrs["xmlns:dc"] = "http://purl.org/dc/elements/1.1/"
 
+            if itunes:
+                for tag in ('subtitle', 'duration', 'explicit'):
+                    key = 'itunes_{}'.format(tag)
+                    args[key] = post.meta[lang].get(key)
+                args['itunes_author'] = (
```
Kwpolska: Write that on one line, will look better and fix a style checker complaint.

nikola/utils.py
```diff
@@ -1255,6 +1255,25 @@ class ExtendedRSS2(rss.RSS2):
...
+        self.itunes_name = itunes_name
+        self.itunes_email = itunes_email
+        self.itunes_image = itunes_image
+        self.itunes_categories = itunes_categories
+        self.itunes_explicit = itunes_explicit
+
+        # It's an old style class
+        rss.RSS2.__init__(self, **kwargs)
+
+    def _itunes_attributes (self):
```
Kwpolska: Remove the space before `(self)` here.

nikola/utils.py
```diff
@@ -1272,15 +1291,49 @@ def publish_extensions(self, handler):
...
+                    handler.startElement("itunes:category", {'text': subcategory})
+                    handler.endElement("itunes:category")
+                handler.endElement("itunes:category")
+
 
 class ExtendedItem(rss.RSSItem):
     """Extended RSS item."""
 
-    def __init__(self, **kw):
+    def __init__(self, itunes_author=None, itunes_subtitle=None,
```
Kwpolska: It might be better to use the `kw` dict as well for your itunes attributes.

nikola/nikola.py
anderbubble: That hasn't been my experience. With this code, the image is located relative to the post's location, and then an absolute url to that location is generated.

nikola/utils.py
anderbubble: Oops. That's a minor stylistic affectation I use in my own code, but don't intend to inflict on anyone. Just a habit. :)

nikola/utils.py
anderbubble: Doesn't that kind of reading-into-kwargs defeat static code analysis? If you don't declare what arguments you take in your function definition, you're unnecessarily hiding information.

At least, that's been my take on it in the past.

Kwpolska: I would prefer to keep it this way if it was started this way — perhaps there’s a reason for using `kw` there?

anderbubble:
I'm actually going a different way on this; leave this open for now, but I'd prefer to merge #2686 rather than this, and do my iTunes-specific stuff in my plugin rather than pollute core.

felixfontein:
Since now #2686 is merged, can we close this @anderbubble?

anderbubble:
@felixfontein for sure. Thanks.

PR #2675, opened 2017-02-19
[WIP] Optionally read rss guid from the post

anderbubble:
I'm working on enhancing Nikola to support publishing a podcast RSS feed. Most of this work is going into a "postcast" plugin I'm working on, and will submit or publish eventually; but some of what I'm trying to do requires support in Nikola itself.

This PR aims to support optional explicit (and, thus, static) GUIDs from posts, so that if I move where the podcast posts are they don't flood podcast apps with old episodes.

> [Assign the GUID to an episode only once and never change it. Assigning new GUIDs to existing episodes can cause issues with your podcast’s listing and chart placement in the iTunes Store.](https://help.apple.com/itc/podcasts_connect/#/itcb54353390)

Other related PRs I'm working on:

* #2676 (copyright tag in rss feeds)
* #2674 (iTunes rss tags)

I'd also be interested in direction for where and how to publish a plugin, when that's ready.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1705,7 +1705,7 @@ class Nikola(object):
                             post.date.astimezone(dateutil.tz.tzutc())),
                 'categories': post._tags.get(lang, []),
                 'creator': post.author(lang),
-                'guid': post.permalink(lang, absolute=True),
+                'guid': post.meta[lang].get('guid') or post.permalink(lang, absolute=True),
             }
 
             if post.author(lang):


Kwpolska:
Sorry, but this has been rejected before in #2378.

anderbubble:
First, thank you for reviewing this pull request.

But I strongly disagree with the reasons given in #2381.

1. I *have* "flooded the planets" before. It was embarrassing and, as a result, I started applying uuid4 guids to all my posts (in the ikiwiki iteration of this site). It hasn't been an issue since.

2. As such, I *have* seen readers use GUID to track read status; and iTunes (which is the consumer I'm currently targeting) explicitly states that GUID must not change.

3. I'm not trying to support moving domains; but, as-is, I can't move them *at all* without making them re-appear in a reader. That's incorrect and embarrassing.

4. I don't see how an optional parameter complicates Nikola's user interface over practicality. As-is, this *cannot* be done in a plugin without duplicating the RSS renderer. Show me how to do it in a plugin, and I'll move the code there.

I'm totally ok with leaving the derived permalink as the default; but since it's not *actually* a permalink (it changes if the post moves) it should be possible to specify something actually permanent if I want.

Kwpolska:
@da2x, @ralsina:

Kwpolska:
@da2x, @ralsina: opinions welcome.

anderbubble:
Thanks for being willing to have the discussion! It's a pleasure to
contribute to Nikola!

felixfontein:
How about instead having the possibility to customize RSS/Atom feed generation? Like having a signal which passes the XML object, so that signal handler plugins can hook up to the signal and add stuff? That would allow to add such specific things without modifying Nikola core.

anderbubble:
Would such a signal handler plugin have access to the post? (I haven't made
a signal handler plugin before.)

If so, would that be a better way to handle the iTunes tags I'm adding in
#2674?

On Sun, Feb 19, 2017, 9:21 AM Felix Fontein <notifications@github.com>
wrote:

How about instead having the possibility to customize RSS/Atom feed
generation? Like having a signal which passes the XML object, so that
signal handler plugins can hook up to the signal and add stuff? That would
allow to add such specific things without modifying Nikola core.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/getnikola/nikola/pull/2675#issuecomment-280929458>, or mute
the thread

felixfontein:
Depends a bit what the signal call passes as its arguments. Passing the XML object and the post list would be a good idea. (You can of course find the posts based on the generated XML, like by using the `guid`, but that's kind of hacky.)

(There's also a different solution which doesn't require any support in Nikola core: you could monkey-patch the `Nikola` object by replacing its `generic_rss_renderer` method by your own function, which does whatever you want. I wouldn't recommend doing that, though.)

anderbubble:
Right: by the time you monkey-patch the RSS function, you're no better than
implementing your own RSS feed in a plugin.

I've gotta say, though: parsing the posts back out of the generated feed
seems like the wrong thing; and I think you'd have to do that even if you
pass the posts to the signal handler, to know what part of the feed to
edit, right? Seems the opposite of the simplicity that is being argued for.

felixfontein:
I fully agree. I think it makes sense to provide the signal handler with a list of pairs `(post, extended_item)` next to the `rss_obj` (and next to `site`). That should allow to make modifications very simple.

Regarding iTunes tags: that would only partially work, since you don't know in the signal handler whether this is a feed where `itunes == True` is passed or not. Also, you'd have to monkey-patch `ExtendedItem` to actually generate the needed XML.

felixfontein:
(Signal handlers are well suited for changes which affect ALL feeds, not for specific ones.)

ralsina:
Sigh. It's a valid issue. I suppose if the post has a guid metadata it could be used, defaulting to what is used now. If it breaks something else when used, you get to keep both pieces :-P

The RSS/Atom generation is unwieldly, and it could be improved, but backwards compatibility is a must. If we change things the wrong way we'll make *everyone* flood the planets.

anderbubble:
While the discussion is happening, can I get the "duplicate" and "wontfix" labels removed from the PR? I'm happy to continue discussing the best way to do this, and learning about signal handlers is probably good; but those labels are vexing me. :)

fwiw, I liked the idea of a signal handler that would receive `(post, extended_item)` pairs. Can `extended_item` pairs be edited in-place? I've only used them to append.

felixfontein:
I removed the labels. For your other question: I guess so, but I don't really know. I've never worked with the RSS feed code.

anderbubble:
So what's the current thinking? Does this need to become "enhance RSS feed generation to support signal handlers" or is the existing implementation path acceptable?

felixfontein:
The last time someone suggested this feature, it was opposed by the argument that most feed readers ignore the GUID anyway. Since we now have one very specific example where the ability to specify a GUID is crucial (iTunes podcasts), *and* this fix is very simple and doesn't add much complexity, I suggest the original concerns do not apply anymore. @ralsina, @da2x, @tritium21 (you three opposed the previous PR): do you agree to that? Or do you still have concerns against this?

Kwpolska:
I’m for merging this as-is.

tritium21:
Already merged, but because of the implementation of this version of the feature doesn't break history like the old one did, I have no objection to it.

felixfontein:
(I think the old implementation of this feature also didn't break anything. It added a function to `Post` to retrieve the GUID and also inserted it into Atom feeds, but its default fallback when the GUID wasn't explicitly specified was also to use the absolute permalink as before.)

tritium21:
I looked at the old implementation....at least the version of it I saw simply swapped the guid from the permalink to a metadata field.  That broke stuff.  I didn't check, just now, if that was changed in later versions, and I don't remember offhand.

felixfontein:
I just looked in the commit referenced in the old PR: https://github.com/julianbrost/nikola/commit/40ba642155baea5a4c68ca3980301941c65696da That one only uses the GUID from metadata if it is specified there, and otherwise uses `post.permalink(lang, absolute=True)`.

anderbubble:
Fwiw, PR #2378 looks better than mine / this one to me.

PR #2679, opened 2017-02-21
Fix #2645 (h/t @benaiah)

### LINKED ISSUE #2645: `nikola auto` serves CSS with wrong MIME type on livereload
I'm using Nikola to develop a static site, and the live-reload functionality of `nikola auto` is not working for CSS. When I save my SASS file, it successfully builds a new CSS file and sends it to the browser, but it sends it with the MIME type "application/octet-stream" instead of "text/css", which makes the page lose *all* styling. The only way to get styling back is to refresh the page manually. I've tested this in both Chrome and Firefox, both on Debian Stretch.

ralsina:


ralsina:
@benaiah I just adopted your fix, since it looked ok.

nikola/plugins/command/auto/__init__.py
```diff
@@ -298,8 +298,12 @@ def serve_static(self, environ, start_response):
         p_uri = urlparse(uri)
         f_path = os.path.join(self.site.config['OUTPUT_FOLDER'], *[unquote(x) for x in p_uri.path.split('/')])
 
+        # mimetypes.guess_type can't handle query strings, so we need
+        # to strip the URI to only the path for mimetype guessing
+        stripped_uri = urlparse.urljoin(url, p_uri.path)
```
Kwpolska: Where is `url` defined?

ralsina: Sorry, brainfart, the right thing is to just use p_uri.path there. Pushed.

PR #2685, opened 2017-02-25
Fix #2684 -- prevent writing cache files outside of the cache folder

### LINKED ISSUE #2684: Posts with paths not relative to cwd put their cache in unusual places
If posts use absolute paths, or paths that refer to parent directories (eg. when provided by custom scanners), the cache ends up appearing in weird places, eg. alongside the source files.

Kwpolska:
This is #2684. Since those are edge cases, changing the post name/base path won’t hurt anybody.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -18,6 +18,8 @@ Features
 Bugfixes
 --------
 
+* Prevent writing cache files outside of the cache folder
+  (Issue #2684)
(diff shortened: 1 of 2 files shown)


Kwpolska:
Thanks! (This was needed for `nikola-themes`, of course. If you have built that site already, nuke `cache`, `.doit.db` and try until it works.)

PR #2688, opened 2017-03-04
Compiling ipynb now looks for default jupyter configuration

ralsina:
This is a followup from #2650

For those, like @knowsuchagency, who use the jupyter extensions, specifically python markdown, this makes it so Nikola will automatically look to see if users have edited the configuration files nbconvert normally looks at when compiling notebooks itself.

Thus if you have a markdown cell that jupyter normally converts to 2 from {{ 1 + 1 }} in your notebook, you'll see the former rather than the later.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Add Jupyter config as dependency for jupyter posts (by @knowsuchagency)
 * Add ``RSS_COPYRIGHT``, ``RSS_COPYRIGHT_PLAIN``, and
   ``RSS_COPYRIGHT_FORMATS`` options in conf.py which can be disabled
   by specifying ``copyright_=False`` to ``generic_rss_renderer``, or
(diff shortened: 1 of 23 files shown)


CHANGES.txt
```diff
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Add Jupyter config as dependency for jupyter posts (by @knowsuchagency)
```
Kwpolska: What’s the purpose of this attribution?

nikola/plugins/compile/ipynb.py
```diff
@@ -190,3 +192,30 @@ def create_post(self, path, **kw):
...
+        return config
+
+    for parent in jupyter_config_path():
+        try:
+            for file in os.listdir(parent):
+                if 'nbconvert' in file and file.endswith('.json'):
+                    abs_path = os.path.join(parent, file)
+                    with open(abs_path) as config_file:
+                        config.update(json.load(config_file))
+        except FileNotFoundError:
```
Kwpolska: `except OSError:` for py2 compatibility

CHANGES.txt
ralsina: New contributor encouragement :-)

nikola/plugins/compile/ipynb.py
ralsina: Done

PR #2693, opened 2017-03-10
Fix #2692 -- make `nikola plugin --list-installed` more readable

### LINKED ISSUE #2692: `nikola plugin --list-installed` is hard to read
The readability of the output of `nikola plugin --list-installed` can be improved by adding colors, e.g., plugin names have different color than plugin paths.

Kwpolska:
* Align plugin list
* Make the plugin name bold

This is #2692 — cc @xuhdev

### DIFF
--- a/nikola/plugins/command/plugin.py
+++ b/nikola/plugins/command/plugin.py
@@ -188,7 +188,7 @@ class CommandPlugin(Command):
             print(formatstring.format(name, path, maxlength))
         dp = self.site.config['DISABLED_PLUGINS']
         if dp:
-            print('\n\nAlso, you have disabled these plugins: {}'.format(dp))
+            print('\n\nAlso, you have disabled these plugins: {}'.format(', '.join(dp)))
         else:
             print('\n\nNo plugins are disabled.')
         return 0
--- a/nikola/plugins/command/plugin.py
+++ b/nikola/plugins/command/plugin.py
@@ -179,9 +179,18 @@ class CommandPlugin(Command):
         plugins.sort()
         print('Installed Plugins:')
         print('------------------')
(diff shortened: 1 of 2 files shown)


xuhdev:
:+1:

nikola/plugins/command/plugin.py
```diff
@@ -179,9 +179,18 @@ def list_installed(self):
         plugins.sort()
         print('Installed Plugins:')
         print('------------------')
+        maxlength = max(len(i[0]) for i in plugins)
+        if self.site.colorful:
+            formatstring = '\x1b[1m{0:<{2}}\x1b[0m  at {1}'
```
ralsina: Does this work on windows?

Kwpolska: On Windows, we just disable the colors everywhere. But Windows 10 *does* support ANSI escape codes, so we can enable colors if the user is on 10.

nikola/plugins/command/plugin.py
```diff
@@ -179,9 +179,18 @@ def list_installed(self):
...
+        else:
+            formatstring = '{0:<{2}}  at {1}'
         for name, path in plugins:
-            print('{0} at {1}'.format(name, path))
-        print('\n\nAlso, you have disabled these plugins: {}'.format(self.site.config['DISABLED_PLUGINS']))
+            print(formatstring.format(name, path, maxlength))
+        dp = self.site.config['DISABLED_PLUGINS']
+        if dp:
+            print('\n\nAlso, you have disabled these plugins: {}'.format(dp))
```
felixfontein: Why not use `', '.join(dp)` to improve readability for the disabled plugin list?

Kwpolska: Will do.

PR #2695, opened 2017-03-12
Attempting to fix #2496.

### LINKED ISSUE #2496: PostScanner plugins have problems due to compilers being disabled if not used by default post scanning plugin

felixfontein:
fixes #2496. closes #2498 (superseeded by this).

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -990,7 +990,7 @@ class Nikola(object):
         for data in plugin_list:
             plugin_dict[data[2].name].append(data)
         result = []
-        for name, plugins in plugin_dict.items():
+        for _, plugins in plugin_dict.items():
             if len(plugins) > 1:
                 # Sort by locality
                 plugins.sort(key=plugin_position_in_places)
@@ -1109,7 +1109,7 @@ class Nikola(object):
                         to_add.append(p)
                     for p in self.disabled_compiler_extensions.pop(k, []):
                         to_add.append(p)
-            for name, p in self.disabled_compilers.items():
+            for _, p in self.disabled_compilers.items():
                 utils.LOGGER.debug('Not loading unneeded compiler {}', p[-1].name)
             for _, plugins in self.disabled_compiler_extensions.items():
                 for p in plugins:
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1089,6 +1089,7 @@ class Nikola(object):
                     file_extensions.update(exts)
                 else:
                     # Stop scanning for more: once we get None, we have to load all compilers anyway
+                    utils.LOGGER.debug("Post scanner {0!r} does not implement `supported_extensions`, loading all compilers".format(post_scanner))
                     file_extensions = None
                     break
             to_add = []
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1111,12 +1111,17 @@ class Nikola(object):
                         to_add.append(p)
             for name, p in self.disabled_compilers.items():
                 utils.LOGGER.debug('Not loading unneeded compiler {}', p[-1].name)
-            for name, plugins in self.disabled_compiler_extensions.items():
+            for _, plugins in self.disabled_compiler_extensions.items():
                 for p in plugins:
                     utils.LOGGER.debug('Not loading compiler extension {}', p[-1].name)
             if to_add:
                 self.plugin_manager._candidates = self._filter_duplicate_plugins(to_add)
                 self.plugin_manager.loadPlugins()
+
+        # IPython theme configuration.  If a website has ipynb enabled in post_pages
+        # we should enable the IPython CSS (leaving that up to the theme itself).
+        if 'needs_ipython_css' not in self._GLOBAL_CONTEXT:
+            self._GLOBAL_CONTEXT['needs_ipython_css'] = 'ipynb' in self.config['COMPILERS']
 
         self._activate_plugins_of_category("Taxonomy")
         self.taxonomy_plugins = {}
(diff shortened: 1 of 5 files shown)


felixfontein:
The last commit should fix the texts, but also induces a potentially backwards compatibility breaking change in case a post scanner plugin needs the final `site._GLOBAL_CONTEXT` in its `set_site`.

The only way to avoid this is probably to get rid of the bad compilers optimization completely. This would also make all the new code superfluous and simplify logic a lot.

felixfontein:
@ralsina: any thoughts and comments?

Kwpolska:
@felixfontein I might not have enough time for a review until the weekend.

nikola/plugins/misc/scan_posts.py
```diff
@@ -106,3 +106,7 @@ def scan(self):
                         raise err
 
         return timeline
+
+    def supported_extensions(self):
+        """Return a list of supported file extensions, or None if such a list isn't known beforehand."""
+        return list(set([os.path.splitext(x[0])[1] for x in self.site.config['post_pages']]))
```
Kwpolska: This can be made simpler. Since we dropped 2.6 support, you can use set comprehensions (`{i for i in x}`). And even if we didn’t, using a generator expression inside `set(i for i in x)` would look better.

```py
        return list({os.path.splitext(x[0])[1] for x in self.site.config['post_pages']})
```

nikola/nikola.py
```diff
@@ -1729,7 +1759,7 @@ def generic_rss_feed(self, lang, title, link, description, timeline,
                             post.date.astimezone(dateutil.tz.tzutc())),
                 'categories': post._tags.get(lang, []),
                 'creator': post.author(lang),
-                'guid': post.guid(lang),
+                'guid': post.permalink(lang, absolute=True),
```
Kwpolska: Was this intentional or are those just some leftovers?

felixfontein: These are definitely leftovers!

nikola/plugins/misc/scan_posts.py
felixfontein: Fixed.

nikola/nikola.py
felixfontein: Looks like that was the only one which got in. I've removed it by now.

felixfontein:
@Kwpolska: take your time! A couple of days more don't really matter ;)

nikola/nikola.py
```diff
@@ -1032,6 +1032,26 @@ def init_plugins(self, commands_only=False, load_all=False):
...
+
+        # Avoid redundant compilers
+        # Remove compilers that match nothing in POSTS/PAGES
+        # And put them in "bad compilers"
+        self.config['COMPILERS'] = {}
+        self.disabled_compilers = {}
+        self.disabled_compiler_extensions = defaultdict(list)
+        self.bad_compilers = set([])
+        for k, v in compilers.items():
+            # self.config['COMPILERS'][k] = sorted(list(v))
```
Kwpolska: This requires a comment to explain (and shouldn’t have commented-out code)

nikola/nikola.py
```diff
@@ -1062,30 +1081,47 @@ def init_plugins(self, commands_only=False, load_all=False):
...
+                    for p in self.disabled_compiler_extensions.pop(k, []):
+                        to_add.append(p)
+            for _, p in self.disabled_compilers.items():
+                utils.LOGGER.debug('Not loading unneeded compiler {}', p[-1].name)
+            for _, plugins in self.disabled_compiler_extensions.items():
+                for p in plugins:
+                    utils.LOGGER.debug('Not loading compiler extension {}', p[-1].name)
+            if to_add:
+                self.plugin_manager._candidates = self._filter_duplicate_plugins(to_add)
+                self.plugin_manager.loadPlugins()
```
Kwpolska: Are plugins loaded through here activated?

nikola/nikola.py
felixfontein: Looks like a leftover from the original code. I've removed it and simplified the thing.

nikola/nikola.py
felixfontein: Yes. Until that point, only plugins of type `PostScanner` have been activated, and all others not. Since we only added plugins of type `PageCompiler` or compiler extensions (plugins having `compiler` set) no plugin should be forgotten.

(Except if a post scanner plugin has `compiler` set in its `.plugin` file. I assumed such plugins don't exist.)

nikola/nikola.py
```diff
@@ -1032,6 +1032,23 @@ def init_plugins(self, commands_only=False, load_all=False):
...
+        # Also add aliases for combinations with TRANSLATIONS_PATTERN
+        for compiler, exts in self.config['COMPILERS'].items():
+            for ext in exts:
+                compilers[compiler].add(ext)
+                for lang in self.config['TRANSLATIONS'].keys():
+                    candidate = utils.get_translation_candidate(self.config, "f" + ext, lang)
+                    compilers[compiler].add(candidate)
+
+        # Avoid redundant compilers
+        # Remove compilers that match nothing in POSTS/PAGES
```
Kwpolska: Could use an explanation of how we gather the compilers, and perhaps a better name than `bad_compilers`.

felixfontein: I removed `bad_compilers`, since the same information is already contained in `disabled_compilers` and nothing in core or `plugins` was using it.

Is the explanation in 53e5f6e good enough, or do you want something more detailed?

Kwpolska:
LGTM. How much testing have you done? Have you tested it with some other PostScanner plugin?

felixfontein:
I've tested it with my pages and blogs; they all use the hierarchical pages post scanner plugin (https://plugins.getnikola.com/v7/hierarchical_pages/), and one of them another post scanner plugin I'm currently working on. So far it seems to work fine.

felixfontein:
(I'm using a slightly adjusted version of the hierarchical pages plugin which defines `supported_extensions`; and used the other plugin first without and later with such a function as well. So it looks like it works both for non-updated as well as updated plugins.)

Kwpolska:
It works fine in both cases (tested with the plugins site). Merging.

felixfontein:
Thanks for merging!

Issue #2646, opened 2017-01-24
KeyError: page_index_folder_index


ericcotelnu:
If setting PAGE_INDEX = True, with prety much default settings (except usual config), then I get:

```
$ nikola build
Traceback (most recent call last):
  File "/usr/local/bin/nikola", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/dist-packages/nikola/__main__.py", line 171, in main
    _ = DN.run(oargs)
  File "/usr/local/lib/python3.5/dist-packages/nikola/__main__.py", line 339, in run
    self.nikola.init_plugins()
  File "/usr/local/lib/python3.5/dist-packages/nikola/nikola.py", line 1077, in init_plugins
    self._activate_plugins_of_category("SignalHandler")
  File "/usr/local/lib/python3.5/dist-packages/nikola/nikola.py", line 1233, in _activate_plugins_of_category
    plugin_info.plugin_object.set_site(self)
  File "/usr/local/lib/python3.5/dist-packages/nikola/plugins/misc/taxonomies_classifier.py", line 328, in set_site
    self._register_path_handlers(taxonomy)
  File "/usr/local/lib/python3.5/dist-packages/nikola/plugins/misc/taxonomies_classifier.py", line 316, in _register_path_handlers
    doc = taxonomy.path_handler_docstrings[name]
KeyError: 'page_index_folder_index'
```

It seems silly, but I'd rather post this for others who might make the same mistake.

ericcotelnu:
Ace, thanks folks. Love that label ("brown paper bag") too, made me LOL when I saw it.

sinewalker:
Applied the patch locally myself, which fixes the crash.  But it doesn't render the index.  Setting STORY_INDEX = True gives a WARNING, but also doesn't render the index.  Here's my PAGES dict:

```python
PAGES = (
    ("pages/*.rst", "pg", "story.tmpl"),
    ("pages/*.txt", "pg", "story.tmpl"),
    ("pages/*.ipynb", "pg", "story.tmpl"),
    ("pages/*.md", "pg", "story.tmpl"),
    ("old_site/stories/*.wp", "pg", "story.tmpl"),
)
```

So I'm expecting output/pg/index.html  (which was being rendered at least in 7.7.1) but it's not there.

Kwpolska:
@felixfontein, any ideas about why PAGE_INDEX doesn’t work at all?

felixfontein:
[...]

PR #2701, opened 2017-03-24
Fixing pages index

felixfontein:
Fixes remaining issue in #2646 and makes the pages index plugin actually work.


Kwpolska:
Doesn’t work yet:

```html
<article class="listpage"><header><h1>Demo Site</h1>
    </header><p>Nothing found.</p>
</article>
```

I’ll try to debug myself.

Kwpolska:
I had to add some special casing for `list.tmpl`, because this is what page indexes wanted. Merging and releasing v7.8.4 if no-one objects.

felixfontein:
Thanks for fixing this! I just checked whether `index.html` was created without errors, but forgot to check whether it actually has content...

PR #2703, opened 2017-03-26
Don’t use doctests anymore

Kwpolska:
Doctests are pretty fragile, require string trickery, and are frowned upon by many other people. Let’s drop them. This commit disables doctests and re-adds all the useful ones as regular unit tests.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -953,26 +953,26 @@ def get_crumbs(path, is_file=False, index_folder=None, lang=None):
     >>> crumbs = get_crumbs('galleries')
     >>> len(crumbs)
     1
-    >>> print('|'.join(crumbs[0]))
-    #|galleries
+    >>> crumbs[0]
+    ['#', 'galleries']
 
     >>> crumbs = get_crumbs(os.path.join('galleries','demo'))
     >>> len(crumbs)
     2
-    >>> print('|'.join(crumbs[0]))
-    ..|galleries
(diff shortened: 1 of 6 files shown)


nikola/utils.py
```diff
@@ -469,9 +469,8 @@ class TemplateHookRegistry(object):
     >>> r = TemplateHookRegistry('foo', None)
```
felixfontein: Isn't this still a doctest? Or do you want to keep it as documentation?

Kwpolska: I didn’t remove any doctests, because they’re still pretty good documentation.

Kwpolska:
Doctest suck because they’re just stringifying the return value and comparing it using string equality. Therefore:

1. If the code is compatible with Python 2 and 3, b'' or u'' might appear in `repr()` of lists (cf. `str.join()` hacks in `get_crumbs` I just removed) — that will not pass `==`.
2. You can’t reasonably do environment-dependent things (cf. paths in `get_asset_path` which had to use a special ELLIPSIS flag and `/.../` — the leading slash was needed because `...` means code continuation line.
3. And because dicts/sets have no order, printing either can mean random doctest failures. You need to compare with `foo() == {'some': 'dict'}` or use `sorted` in the doctest. Which beats the point of doctests.

Merging. (Ignore the test failures — that’s invariance, caused by this branch being created pre-v7.8.4)

felixfontein:
Ah, so nothing really new.

PR #2705, opened 2017-03-27
Make PAGE_INDEX work with PRETTY_URLS

Kwpolska:
This should end the PAGE_INDEX saga for good. PRETTY_URLS and some corner cases, including `.php` are supported (and I also unearthed a bug in #2706).

### DIFF
--- a/nikola/plugins/task/page_index.py
+++ b/nikola/plugins/task/page_index.py
@@ -64,7 +64,7 @@ class PageIndex(Taxonomy):
     def classify(self, post, lang):
         """Classify the given post for the given language."""
         destpath = post.destination_path(lang, sep='/')
-        if post._has_pretty_url(lang):
+        if post.has_pretty_url(lang):
             idx = '/index.html'
             if destpath.endswith(idx):
                 destpath = destpath[:-len(idx)]
--- a/nikola/plugins/task/page_index.py
+++ b/nikola/plugins/task/page_index.py
@@ -64,6 +64,10 @@ class PageIndex(Taxonomy):
     def classify(self, post, lang):
         """Classify the given post for the given language."""
         destpath = post.destination_path(lang, sep='/')
+        if self.site.config["PRETTY_URLS"]:
+            index_len = len(self.site.config["INDEX_FILE"])
+            if destpath[-(1 + index_len):] == '/' + self.site.config["INDEX_FILE"]:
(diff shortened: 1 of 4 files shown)


felixfontein:
If you replace `if self.site.config["PRETTY_URLS"]:` by `if post._has_pretty_url(lang):`, it works flawlessy. (Too bad it's kind of a private function...)

Also, we shouldn't use `self.site.config["INDEX_FILE"]` here, since `PRETTY_URL` will *always* result in `index.html` to be appended, no matter what `self.site.config["INDEX_FILE"]` is (see definition of `Post.destination_path`).

Kwpolska:
Done. @felixfontein, @ralsina: reviews welcome. I’d like to get this out in a release soon.

nikola/plugins/task/page_index.py
```diff
@@ -64,6 +64,8 @@ def is_enabled(self, lang=None):
     def classify(self, post, lang):
         """Classify the given post for the given language."""
         destpath = post.destination_path(lang, sep='/')
+        if post._has_pretty_url(lang):
+            destpath = destpath.rstrip('/index.html')
```
felixfontein: If, for whatever obscure reason, your path happens to be `https://example.com/something/index.html/index.html/index.html`, this will not result in what you want ;-)

nikola/plugins/task/page_index.py
```diff
@@ -64,6 +64,8 @@ def is_enabled(self, lang=None):
     def classify(self, post, lang):
         """Classify the given post for the given language."""
         destpath = post.destination_path(lang, sep='/')
+        if post._has_pretty_url(lang):
```
felixfontein: Should we maybe turn this in a version not prefixed by `_`?

nikola/plugins/task/page_index.py
felixfontein: Actually, `rstrip` is not a good idea here at all: it removes all letters from the right of the string which are *contained* in `/index.html`. So also `/end/html/nix.html` would be removed, for example.

nikola/plugins/task/page_index.py
Kwpolska: Can do.

felixfontein: Cool, thanks!

PR #2714, opened 2017-04-11
Update MathJax CDN link.

### LINKED ISSUE #2713: Update MathJax links
Mathjax is shutting down its CDN end of April: https://www.mathjax.org/cdn-shutting-down/

The links in the templates must be updated to point to the new locations for MathJax.

Anteru:
### Description

This fixes #2713 by updating the MathJax links to cdnjs and fixing it to version 2.7.0 (instead of latest).

### DIFF
--- a/nikola/data/themes/base-jinja/templates/index_helper.tmpl
+++ b/nikola/data/themes/base-jinja/templates/index_helper.tmpl
@@ -37,7 +37,7 @@
                 </script>
             {% endif %}
         {% else %}
-            <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"> </script>
+            <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_HTMLorMML"> </script>
             {% if mathjax_config %}
             {{ mathjax_config }}
             {% else %}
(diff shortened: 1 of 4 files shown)


Anteru:
There's no latest on the new CDN.

Anteru:
Sphinx updated to 2.7.0 as well: https://github.com/sphinx-doc/sphinx/commit/78148e9943c29e83d09ffa5ecd960c76185b7674 -- there's no latest any more. Either you update to 2.7.0, or it just stops working for everyone in two weeks.

felixfontein:
It won't (or at least, shouldn't) stop because of the redirect. But yes, it would be good to get this fixed ASAP.

@Kwpolska: how about injecting the MathJax version into the templates as a variable? (Maybe via global context?) Then at least there's only one place to update, not four. The big downside would be that everything rebuilds at the next update.

I also noticed that we have the same problem with KaTeX. There, version 0.6.0 is hardcoded in the templates (while the latest version is 0.7.1).

Kwpolska:
Putting it in `GLOBAL_CONTEXT` will result in rebuilding too much. I’ll bump KaTeX for now as well and merge.

Also, @Anteru: if you have a patch ready, just send a PR — otherwise, discussion might split between the two (I had to look for the announcement link in the other issue)

Anteru:
Sure, I just did this because the CONTRIBUTING.rst says "First, make sure there's an open issue", then create the branch and make a pull request. I wasn't sure if it's ok to just do a PR without the issue.

Kwpolska:
It’s okay, I’ve just reworded that section (commit a953fbb0)

PR #2718, opened 2017-04-13
Fix #2328, Fix #2712 -- Create a list of template variables

### LINKED ISSUE #2328: Can we get a list of objects available in global context?
It would be extremely handy when creating a theme. Something like [this doc page](http://gohugo.io/templates/variables/) for Hugo.

### LINKED ISSUE #2712: Where can I find what default fields are available in templates?
For example, in templates, we have `${header.html_site_title()}` for the site title. Where can I find the whole list?

Kwpolska:
This is #2328 and #2712. This should make it a lot easier to create themes.

Reviews, comments, insights, automation (hah!) welcome. @felixfontein, how should I phrase `cat_items` contents in `tags.tmpl`?

### DIFF
--- a/docs/template-variables.rst
+++ b/docs/template-variables.rst
@@ -1,0 +1,393 @@
+.. title: Template variables
+.. slug: template-variables
+.. date: 2017-04-13 12:00:00
+.. author: The Nikola Team
+
+:Version: 7.8.4
+:Author: Chris Warrick <chris@getnikola.com>
+
+Variables available in templates are listed below.
+
+* This list is maintained by humans, so it may not always be perfect.
+* Variables whose types are marked with ``?`` may not always be available or may be None in some cases.
+* Templates usually do not have access to the original TranslatableSetting
+  variables, only to the current locale version (except ``NAVIGATION_LINKS``).
+* For function and setting documentation, please consult `code documentation
+  <https://docs.getnikola.com/en/latest/modules/>`_ and `default configuration
+  <https://getnikola.com/conf.html>`_ respectively.
+* Templates often create their own functions (macros), and import macros from
+  other templates. Those macros are not listed here.
+
+Variables and functions come from three places:
+
+* the global context
+* the local context of a page
+* the templates themselves and the templates they import
+
+.. class:: alert alert-info
+.. contents::
+
+Global variables
+----------------
+
+Some variables on the global variables list may be None (the ``?`` symbol is not used).
+
+.. class:: table table-bordered table-striped
+
+==================================  ==================================  ================================================================================
+Name                                Type                                Description
+==================================  ==================================  ================================================================================
+``_link``                           function                            ``Nikola.link`` function
+``abs_link``                        function                            ``Nikola.abs_link`` function
+``annotations``                     bool                                ``ANNOTATIONS`` setting
+``author_pages_generated``          bool                                False
+``blog_author``                     TranslatableSetting<str>            ``BLOG_AUTHOR`` setting
+``blog_description``                TranslatableSetting<str>            ``BLOG_DESCRIPTION`` setting
+``blog_title``                      TranslatableSetting<str>            ``BLOG_TITLE`` setting
+``blog_url``                        str                                 ``SITE_URL`` setting
+``body_end``                        TranslatableSetting<str>            ``BODY_END`` setting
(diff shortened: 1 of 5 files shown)


docs/template-variables.rst
```diff
@@ -0,0 +1,393 @@
...
+``prevlink``                    str             Link to previous page
+``nextlink``                    str             Link to next page
+``prevfeedlink``                str             Link to previous page as an Atom feed
+``nextfeedlink``                str             Link to next page as an Atom feed
+``prev_next_links_reversed``    bool            Whether or not previous and next links should be reversed (``INDEXES_STATIC``)
+==============================  ==============  ===============================================================================
+
+Variables available in taxonomies
+---------------------------------
+
```
felixfontein: Another useful list would be which taxonomies use which templates (in which cases). Should I compile a list, or do you want to do that?

docs/template-variables.rst
```diff
@@ -0,0 +1,393 @@
...
+Variables available in the “Tags and categories” page (``tags.tmpl``)
+---------------------------------------------------------------------
+
+.. class:: table table-bordered table-striped
+
+==============  ======  ===========================================================================================================
+Name            Type    Description
+==============  ======  ===========================================================================================================
+``items``       list    Tags *(name, link)*
+``cat_items``   list    Categories *(name, full name, path, link, indent levels, indent to change before, indent to change after)*
```
felixfontein: I think that looks good. There should be somewhere (either here or above where `<overview_page_hierarchy_variable_name>` and `<overview_page_hierarchy_variable_name + "_with_postcount">` are defined) an explanation what indent levels, indent to change before, and indent to change after means. (Or is that what you meant with "how should I phrase"?)

I would write something like:

The indenting information can be render the items as a tree. The values have the following meanings:
 * `indent levels` is a list of pairs `(current_i, count_i)` giving the current position (`0`, ..., `count_i-1`) and maximum (`count_i`) in the hierarchy level `i`;
 * `indent to change before` is the difference of hierarchy levels between the previous and the current item; positive values indicate that the current item is indented further in and can be used to open HTML tags before the item;
 * `indent to change after` is the difference of hierarchy levels between the current and the next item; negative values indicate that the current item is indented further in and can be used to close HTML tags after the item.

Example:
```
 +--- levels:[(0,3)], before:1, after:0
 +-+- levels:[(1,3)], before:0, after:1
 | +--- levels:[(1,3), (0,2)], before:1, after:0
 | +-+- levels:[(1,3), (1,2)], before:0, after:1
 |   +--- levels:[(1,3), (1,2), (0, 1)], before:1, after:-2
 +-+- levels:[(2,3)], before:-2, after:1
   +- levels:[(2,3), (0,1)], before:1, after:-2
```
See `tags.tmpl` in the base themes for examples on how to render a tree as nested unordered lists in HTML.

docs/template-variables.rst
felixfontein: More generally, it would be good to have a list of which templates are used by which parts of Nikola. This is also helpful for people who want to create a template from scratch instead of deriving from the base templates.

Kwpolska: I’ll try building those lists.

docs/template-variables.rst
Kwpolska: Added that.

docs/template-variables.rst
```diff
@@ -0,0 +1,393 @@
...
+.. class:: table table-bordered table-striped
+
+==================================================================  ======  ============================================
+Name                                                                Type    Description
+==================================================================  ======  ============================================
+``<overview_page_variable_name>``                                   str     List of classifications
+``<overview_page_items_variable_name>``                             str     List of items **(name, link)**
+``<overview_page_items_variable_name + "_with_postcount">``         str     List of items **(name, link, count)**
+``<overview_page_hierarchy_variable_name>``                         str?    List of hierarchies
+``<overview_page_hierarchy_variable_name + "_with_postcount">``     str     List of hierarchies, with added counts
```
Kwpolska: Added all. And yes, `str` was a copy-paste error.

docs/template-variables.rst
Kwpolska: Done; a template list is already in `theming.txt`.

felixfontein: Ah, perfect!

PR #2735, opened 2017-04-24
Avoid empty subarchives

### LINKED ISSUE #2734: Archive hierarchy navigation generates broken links in some cases
If a blog has more than one language and not all posts are available in all languages (and `SHOW_UNTRANSLATED_POSTS` is `False`), and if the archive plugin generates archives per day and `CREATE_ARCHIVE_NAVIGATION` is `True`, broken links can appear:

Say you have posts on 2017/04/22 and 2017/04/23, but on 2017/04/22 you have posts in all languages and on 2017/04/23 there's no translation for a specific language, then the archive page for 2017/04/23 for that language is not created, but the archive navigation for 2017/04/22 for that language has a next link to that non-existing archive.

This only happens on day level, since on month or year level, there's special code which generates the empty archives (so you get links to empty archives). This special logic was added in 650affef3ceb1895b18fcc7f8ef25165d76ffce9, which replicates the behavior of the old archives plugin (https://github.com/getnikola/nikola/blob/v7.8.1/nikola/plugins/task/archive.py).

There are two possible solutions:

  1. We modify the behavior of the archives taxonomy to not generate empty indexes (except for the main archive), i.e. we revert 650affef3ceb1895b18fcc7f8ef25165d76ffce9. Additionally we fix `Archive.postprocess_posts_per_classification` to ignore empty archives as follows:
``` .py
                for node in flat_hierarchy:
                    if not self.site.config["SHOW_UNTRANSLATED_POSTS"]:
                        if not [x for x in posts_per_section_per_language[lang][node.classification_name] if x.is_translation_available(lang)]:
                            continue
                    self.archive_navigation[lang][len(node.classification_path)].append(node.classification_name)
```

  2. We keep this strange logic, and modify `Archive.postprocess_posts_per_classification` to ignore empty archives only on the day level:
``` .py
                for node in flat_hierarchy:
                    if not self.site.config["SHOW_UNTRANSLATED_POSTS"] and len(node.classification_path) == 3:
                        if not [x for x in posts_per_section_per_language[lang][node.classification_name] if x.is_translation_available(lang)]:
                            continue
                    self.archive_navigation[lang][len(node.classification_path)].append(node.classification_name)
```

So the main question is: do we want to keep the strange behavior of the archive (and generate empty archives for year and month level, but not for day level), or just generate a empty global archive and don't generate other empty archives?

Or do we want to generate empty archives (so that there's a symmetry between different languages and you can easily point from the archive in one language to the same archives in other languages), but then generate all of them, and somehow fix the navigation links so you don't get links to empty archives? (Links to empty archives kind of suck from my point of view.)

felixfontein:
Fixes #2734.

Some archive pages are no longer created (more precisely: empty pages on month and year level), so strictly speaking this is breaking backwards compatibility.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -24,6 +24,8 @@ Features
 Bugfixes
 --------
 
+* No longer creates empty subarchive pages, and no longer create broken
+  archive navigation links on day level (Issue #2734)
 * Fixes post scanner plugin order (Issue #2720)
 * Rename ``POSTS_SECTION_ARE_INDEXES`` to ``POSTS_SECTIONS_ARE_INDEXES``
 * Make date ranges work in shortcode-based post lists (Issue #2690)
--- a/nikola/plugins/task/archive.py
+++ b/nikola/plugins/task/archive.py
@@ -219,7 +219,7 @@ class Archive(Taxonomy):
         kw.update(context)
         return context, kw
 
-    def postprocess_posts_per_classification(self, posts_per_section_per_language, flat_hierarchy_per_lang=None, hierarchy_lookup_per_lang=None):
+    def postprocess_posts_per_classification(self, posts_per_archive_per_language, flat_hierarchy_per_lang=None, hierarchy_lookup_per_lang=None):
         """Rearrange, modify or otherwise use the list of posts per classification and per language."""
         # Build a lookup table for archive navigation, if we’ll need one.
         if self.site.config['CREATE_ARCHIVE_NAVIGATION']:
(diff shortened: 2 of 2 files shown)


felixfontein:
I agree. Should we wait for @ralsina, or merge now?

Kwpolska:
Let’s wait.

felixfontein:
@ralsina: can you take a look at this PR?

ralsina:
LGTM

felixfontein:
Thanks!

PR #2737, opened 2017-04-28
Use shortcodes in a more robust manner

ralsina:
### Pull Request Checklist

- [x] I tested my changes.

### Description

This changes how shortcodes are applied:

* Before compiling the file, shortcodes are extracted and replaced by UUIDs
* After compiling, shortcodes are processed and re-inserted where the UUID markers are

This avoids problems like rst mangling URLs and such.

Still needs work:

* apply_shortcodes_2 is not a proper function name ;-)
* All compilers can be switched to this approach gradually, but I would prefer if they all switched at once.

The older API can be kept for backwards compatibility since it has not been touched.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1692,6 +1692,17 @@ class Nikola(object):
             lang = utils.LocaleBorg().current_lang
         return shortcodes.apply_shortcodes(data, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
 
+    def apply_shortcodes2(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context={}):
+        """Apply shortcodes from the registry on data."""
+        if lang is None:
+            lang = utils.LocaleBorg().current_lang
+        deps = []
+        for k, v in shortcodes:
+            replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
+            data.replace(k, replacement)
+            deps.extend(_deps)
+        return data
+
     def _get_rss_copyright(self, lang, rss_plain):
         if rss_plain:
             return (
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1703,7 +1703,7 @@ class Nikola(object):
         deps = []
         for k, v in _shortcodes.items():
             replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
-            data.replace(k, replacement)
+            data = data.replace(k, replacement)
             deps.extend(_deps)
         return data, deps
 
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1694,7 +1694,7 @@ class Nikola(object):
             lang = utils.LocaleBorg().current_lang
         return shortcodes.apply_shortcodes(data, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
 
-    def apply_shortcodes2(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
+    def apply_shortcodes_uuid(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
         """Apply shortcodes from the registry on data."""
         if lang is None:
             lang = utils.LocaleBorg().current_lang
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1705,7 +1705,7 @@ class Nikola(object):
             replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
             data.replace(k, replacement)
             deps.extend(_deps)
-        return data
+        return data, deps
 
     def _get_rss_copyright(self, lang, rss_plain):
         if rss_plain:
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1694,14 +1694,14 @@ class Nikola(object):
             lang = utils.LocaleBorg().current_lang
         return shortcodes.apply_shortcodes(data, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
 
-    def apply_shortcodes_uuid(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
+    def apply_shortcodes_uuid(self, data, _shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
         """Apply shortcodes from the registry on data."""
         if lang is None:
             lang = utils.LocaleBorg().current_lang
         if extra_context is None:
             extra_context = {}
         deps = []
-        for k, v in shortcodes:
+        for k, v in _shortcodes.items():
             replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
             data.replace(k, replacement)
             deps.extend(_deps)
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1686,16 +1686,20 @@ class Nikola(object):
         self.shortcode_registry[name] = f
 
     # XXX in v8, get rid of with_dependencies
-    def apply_shortcodes(self, data, filename=None, lang=None, with_dependencies=False, extra_context={}):
+    def apply_shortcodes(self, data, filename=None, lang=None, with_dependencies=False, extra_context=None):
         """Apply shortcodes from the registry on data."""
+        if extra_context is None:
+            extra_context = {}
         if lang is None:
             lang = utils.LocaleBorg().current_lang
         return shortcodes.apply_shortcodes(data, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
 
-    def apply_shortcodes2(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context={}):
+    def apply_shortcodes2(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
         """Apply shortcodes from the registry on data."""
         if lang is None:
             lang = utils.LocaleBorg().current_lang
+        if extra_context is None:
+            extra_context = {}
         deps = []
         for k, v in shortcodes:
             replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -58,6 +58,7 @@ import lxml.html
 from yapsy.PluginManager import PluginManager
 from blinker import signal
 
+
 from .post import Post  # NOQA
 from .state import Persistor
 from . import DEBUG, utils, shortcodes
--- a/nikola/plugins/compile/rest/__init__.py
+++ b/nikola/plugins/compile/rest/__init__.py
@@ -82,7 +82,7 @@ class CompileRest(PageCompiler):
         }
 
         from nikola import shortcodes as sc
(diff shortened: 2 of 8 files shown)


nikola/shortcodes.py
```diff
@@ -209,14 +209,58 @@ def _parse_shortcode_args(data, start, shortcode_name, start_pos):
...
+                for c in buffer[1:]:  # Dump buffered text
+                    new_data.append(c[1])
+            # Need to buffer to know if this SC closes
+            buffer = [chunk]
+            in_sc = True
+        elif chunk[0] == 'SHORTCODE_END':
+            # Previous shortcode closes here
+            buffer.append(chunk)
+            sc_id = str(uuid.uuid4())
+            new_data.append(sc_id)
```
Kwpolska: The UUID might not be enough, perhaps add some special marker that it’s a shortcode. Also, I assume it’s guaranteed not to appear in the final cache/ file?

nikola/plugins/task/categories.py
```diff
@@ -116,6 +116,9 @@ def slugify_tag_name(self, name, lang):
 
     def slugify_category_name(self, path, lang):
         """Slugify a category name."""
+        if not path:  # Probably used link://category
```
Kwpolska: ?

nikola/nikola.py
```diff
@@ -1686,12 +1686,27 @@ def register_shortcode(self, name, f):
...
-    def apply_shortcodes(self, data, filename=None, lang=None, with_dependencies=False, extra_context={}):
+    def apply_shortcodes(self, data, filename=None, lang=None, with_dependencies=False, extra_context=None):
         """Apply shortcodes from the registry on data."""
+        if extra_context is None:
+            extra_context = {}
         if lang is None:
             lang = utils.LocaleBorg().current_lang
         return shortcodes.apply_shortcodes(data, self.shortcode_registry, self, filename, lang=lang, with_dependencies=with_dependencies, extra_context=extra_context)
 
+    def apply_shortcodes2(self, data, shortcodes, filename=None, lang=None, with_dependencies=False, extra_context=None):
```
Kwpolska: Eh, even the best people have functions named like this. `apply_shortcodes_uuid`?

nikola/plugins/task/categories.py
felixfontein: I guess this wasn't intended to be part of this PR

nikola/shortcodes.py
```diff
@@ -83,6 +80,9 @@ def _skip_whitespace(data, pos, must_be_nontrivial=False):
 
 def _skip_nonwhitespace(data, pos):
     """Return first position not before pos which contains a non-whitespace character."""
+    for i, x in enumerate(data[pos:]):
+        if x.isspace():
+            return pos + i
```
felixfontein: Now you are iterating over the rest of the string twice in case no space is found.

nikola/shortcodes.py
felixfontein: Also some compiler might mangle the `-` in the UUID. I'd remove the dashes to be on the safe side.
(And add some marker, like `sc_id = 'SHORTCODE{0}REPLACEMENT'.format(uuid.uuid4().replace('-', ''))`.)

nikola/shortcodes.py
```diff
@@ -209,14 +209,58 @@ def _parse_shortcode_args(data, start, shortcode_name, start_pos):
     raise ParsingError("Shortcode '{0}' starting at {1} is not terminated correctly with '%}}}}'!".format(shortcode_name, _format_position(data, start_pos)))
 
 
+def _extract_shortcodes(data):
```
felixfontein: Why use a private function name? Isn't this function supposed to be used by plugins? It shouldn't start with an underscore IMO.

nikola/shortcodes.py
```diff
@@ -209,14 +209,58 @@ def _parse_shortcode_args(data, start, shortcode_name, start_pos):
...
     """Given input data, splits it into a sequence of texts, shortcode starts and shortcode ends.
 
     Returns a list of tuples of the following forms:
 
-        1. (_TEXT, text)
-        2. (_SHORTCODE_START, text, start, name, args)
-        3. (_SHORTCODE_END, text, start, name)
+        1. ("TEXT", text)
+        2. ("SHORTCODE_START", text, start, name, args)
+        3. ("SHORTCODE_END", text, start, name)
```
felixfontein: Why do you do these replacements? Before, you'd get a runtime error in case you did a typo (because the symbol with the typo doesn't exist). Now you won't notice anything until something behaves strange, and then you have to some debugging to find out that you did a typo.

nikola/shortcodes.py
ralsina: No, it's only meant to be used right here.

nikola/shortcodes.py
ralsina: Because reading the parsed code made my head hurt.

nikola/shortcodes.py
ralsina: The UUID should be unique enough, considering there are more UUIDs than atoms in the universe. The stuff about dashes sounds good.

nikola/plugins/task/categories.py
ralsina: Yeah, it's not.

nikola/nikola.py
ralsina: Sounds good.

nikola/shortcodes.py
ralsina: No I am not? Sorry, it's late and I don't see it :-)

felixfontein: The `for` loop tries to find the first non-whitespace. If it doesn't find anything, the `while` loop (which is still there below) does the same thing afterwards, and it also won't find anything, so finally `len(data)` will be returned. You can replace the `while` loop with `return len(data)`.

nikola/shortcodes.py
Kwpolska: UUIDs are unique within each other, but not in the world. If you just use an UUID, especially without dashes, you might conflict with, say, a checksum that is part of the user content (or any other 128+-bit integer in hex format — which is not so rare on security–
 or low-level-coding–oriented blogs).

ralsina:
This has a bug when shortcodes are "nested". Specifically this is used in the manual, with the "raw" shortcode wrapping another shortcode that should not be processed.

nikola/shortcodes.py
ralsina: Ah, I forgot to delete all that. Will fix.

ralsina:
@Kwpolska @felixfontein if you could give this code another look, this is sort-of-done I think.

nikola/shortcodes.py
```diff
@@ -209,14 +205,87 @@ def _parse_shortcode_args(data, start, shortcode_name, start_pos):
...
+                text.append(token[1])
+                return ''.join(text), data[1:]
+            elif token[0] == 'SHORTCODE_END':  # This is malformed
+                raise Exception('Closing unopened shortcode {}'.format(token[3]))
+
+    text = ''
+    tail = splitted
+    while True:
+        new_text, tail = extract_data_chunk(tail)
+        text += new_text
```
felixfontein: Wouldn't it be better to make `text` an array of strings, and return `''.join(text)` at the end?

nikola/shortcodes.py
```diff
@@ -209,14 +205,87 @@ def _parse_shortcode_args(data, start, shortcode_name, start_pos):
...
+def extract_shortcodes(data, _new_sc_id=_new_sc_id):
+    """
+    Return data with replaced shortcodes, shortcodes.
+
+    data is the original data, with the shortcodes replaced by UUIDs.
+
+    a dictionary of shortcodes, where the keys are UUIDs and the values
+    are the shortcodes themselves ready to process.
+
+    >>> c = 0
```
Kwpolska: Doctests are not executed anymore (#2703), it might be better to make those real tests.

nikola/plugins/compile/rest/__init__.py
```diff
@@ -81,14 +81,17 @@ def compile_string(self, data, source_path=None, is_two_file=True, post=None, la
             'language_code': LEGAL_VALUES['DOCUTILS_LOCALES'].get(LocaleBorg().current_lang, 'en')
         }
 
+        from nikola import shortcodes as sc
```
Kwpolska: What about other compilers? Also, I think it might be a slightly better-looking API to put `extract_shortcodes` in the Nikola object as well.

ralsina: @Kwpolska Since the old API still works, later branches can port each one to the new API.

I did not put extract_shortcodes in the Nikola object mostly because there's nothing site-specific in it, and I am half convinced to use a real tokenizer in shortcodes.py which would make things awkward if things need refactoring.

nikola/shortcodes.py
ralsina: Oh, well. I'll do that then.

nikola/shortcodes.py
ralsina: Sure.

ralsina:
@felixfontein the demo site has a bunch of shortcodes usage, and this passes the invariance test, so at least that much works :-)

tests/test_shortcodes.py
```diff
@@ -74,3 +74,23 @@ def test_errors(self):
...
+    ('AAA{{% foo %}} bar {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} bar {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% /bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% /bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% bar %}} quux {{% /bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% bar %}} quux {{% /bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} BBB {{% bar %}} quux {{% /bar %}} CCC', (u'AAASC1 BBB SC2 CCC', {u'SC1': u'{{% foo %}}', u'SC2': u'{{% bar %}} quux {{% /bar %}}'})),
+])
+def test_extract_shortcodes(input, expected, monkeypatch):
+
+    i = iter('SC%d' % i for i in range(1, 100))
+    if sys.version[0] < "3":
```
Kwpolska:
```py
if sys.version_info[0] < 3:
```

felixfontein: No, you did `sys.version[0] < 3`.

According to the documentation of `sys.version`: "Do not extract version information out of it, rather, use `version_info` [...]" (https://docs.python.org/2/library/sys.html#sys.version_info, https://docs.python.org/3/library/sys.html#sys.version)

ralsina: Ahhhh right.

tests/test_shortcodes.py
```diff
@@ -74,3 +74,23 @@ def test_errors(self):
...
+    ('{{% foo %}} bar {{% /foo %}}', (u'SC1', {u'SC1': u'{{% foo %}} bar {{% /foo %}}'})),
+    ('AAA{{% foo %}} bar {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} bar {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% /bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% /bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} {{% bar %}} quux {{% /bar %}} {{% /foo %}}BBB', (u'AAASC1BBB', {u'SC1': u'{{% foo %}} {{% bar %}} quux {{% /bar %}} {{% /foo %}}'})),
+    ('AAA{{% foo %}} BBB {{% bar %}} quux {{% /bar %}} CCC', (u'AAASC1 BBB SC2 CCC', {u'SC1': u'{{% foo %}}', u'SC2': u'{{% bar %}} quux {{% /bar %}}'})),
+])
+def test_extract_shortcodes(input, expected, monkeypatch):
+
+    i = iter('SC%d' % i for i in range(1, 100))
```
ralsina: I should change this to use itertools.counter anyway

PR #2743, opened 2017-05-04
Sphinx-style permalinks filter

ralsina:
### Pull Request Checklist

- [x] I tested my changes.

### Description

A filter implementing Sphinx-style header links. Bonus: it works for any compiler, since it's done as HTML post-processing.

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -1911,6 +1911,13 @@ jsonminify
 xmlminify
    Minify XML files. Suitable for Nikola’s sitemaps and Atom feeds.
 
+add_header_permalinks
+   Add links next to every header, Sphinx-style. You will need to add styling for the `headerlink` class,
+   in `custom.css`, for example::
+
+        .headerlink { opacity: 0.1 }
+        .headerlink:hover { opacity: 1; text-decoration: none; }
+
 You can apply filters to specific posts or pages by using the ``filters`` metadata field:
 
 .. code:: restructuredtext
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -1913,10 +1913,20 @@ xmlminify
 
 add_header_permalinks
    Add links next to every header, Sphinx-style. You will need to add styling for the `headerlink` class,
-   in `custom.css`, for example::
+   in `custom.css`, for example:
 
-        .headerlink { opacity: 0.1 }
-        .headerlink:hover { opacity: 1; text-decoration: none; }
+   .. code:: css
+      .headerlink { opacity: 0.1; margin-left: 0.2em; }
+      .headerlink:hover { opacity: 1; text-decoration: none; }
+
+   Additionally, you can provide a custom list of XPath expressions which should be used for finding headers (``{hx}}`` is replaced by headers h1 through h6).
+   This is required if you use a custom theme that does not use ``"e-content entry-content"`` as a class for post and page contents.
+
+   .. code:: python
+        # Default value:
+        HEADER_PERMALINKS_XPATH_LIST = ['*//div[@class="e-content entry-content"]//{hx}']
+        # Include *every* header (not recommended):
+        # HEADER_PERMALINKS_XPATH_LIST = ['*//{hx}']
 
 You can apply filters to specific posts or pages by using the ``filters`` metadata field:
 
--- a/nikola/filters.py
+++ b/nikola/filters.py
@@ -403,6 +403,7 @@ normalize_html = apply_to_text_file(_normalize_html)
 
 @apply_to_text_file
 def add_header_permalinks(data):
+    """Post-process HTML via lxml to add header permalinks Sphinx-style."""
     doc = lxml.html.document_fromstring(data)
     for h in ['h1', 'h2', 'h3', 'h4']:
         nodes = doc.findall('*//%s' % h)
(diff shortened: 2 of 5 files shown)


felixfontein:
Permalinks are not supposed to change every time a HTML page is regenerated, right? But isn't that what's happening here because `uuid.uuid4()` always generates a new value which was never generated before?

nikola/filters.py
```diff
@@ -391,3 +392,15 @@ def _normalize_html(data):
...
+
+
+@apply_to_text_file
+def add_header_permalinks(data):
+    """Post-process HTML via lxml to add header permalinks Sphinx-style."""
+    doc = lxml.html.document_fromstring(data)
+    for h in ['h1', 'h2', 'h3', 'h4']:
+        nodes = doc.findall('*//%s' % h)
+        for node in nodes:
+            new_node = lxml.html.fragment_fromstring('<a id="{0}" href="#{0}" class="headerlink" title="Permalink to this headline">&nbsp;&pi;</a>'.format(uuid.uuid4()))
```
Kwpolska: You’re doing UUIDs too much. I would:
* check if the header has an `id` already, and use that
* otherwise, create a slugified form of the header and use that, watching out for conflicts within the same page and appending numbers [(example for Django)](https://github.com/Kwpolska/achieve/blob/master/achieve/helpers.py#L47)

Kwpolska: Also, what @felixfontein said about permalinks being, um, *permanent*.

ralsina: Honestly, I don't care about this enough to do the extra work.

ralsina:
@felixfontein that is a very good point. So, the id generation here is useless.

nikola/filters.py
Kwpolska: I can take over if you want.

ralsina: @Kwpolska that would be great, thanks!

ralsina:
LGTM with @Kwpolska's latest fix

Kwpolska:
I’m not done yet, I want to at least implement the blacklist (or perhaps whitelist)

felixfontein:
Since the filter cannot/doesn't have persistent state, the link generation algorithm cannot be changed anymore after merging this (or at least from the next release on). Otherwise this filter doesn't generate permalinks, but code-dependent links which will change as soon as this code is touched in a non-trivial way! So we really shouldn't merge this one too early!

Kwpolska:
The link generation algorithm I have in mind will be pretty stable, so no worries. (The links being permanent also depend on user actions)

Kwpolska:
@ralsina, @felixfontein: this is ready for review. Header/id deduplication will be implemented in another PR, because it needs to be done for all IDs, not only the few headers we touch here.

nikola/filters.py
```diff
@@ -397,3 +397,39 @@ def _normalize_html(data):
...
+    except KeyError:
+        # Circular import workaround (utils imports filters)
+        from nikola.utils import LocaleBorg
+        lang = LocaleBorg().current_lang
+
+    if not xpath_list:
+        xpath_list = ['*//div[@class="e-content entry-content"]//{hx}']
+    for xpath_expr in xpath_list:
+        for hx in ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']:
+            nodes = doc.findall(xpath_expr.format(hx=hx))
```
felixfontein: In case `xpath_expr` does *not* contain `hx`, wouldn't this add a permalink six times for every matching node?

felixfontein: (Maybe first put the formatted expressions into a set, then convert the set to a sorted list, and iterate over it. That should fix this problem.)

Kwpolska: Yes, it will. This sounds like an edge case, but are you aware of any good solutions for this?

Kwpolska: Made it a set. There’s still the case of the same header being returned by multiple expressions, should I also work around that?

Kwpolska:
> LGTM, but kind of incomplete without that extra deduplication PR.

The code is waiting on the `deduplicate-headers` branch until this lands.

> Also, how permanent these permalinks will be depends a lot on the deduplication algorithm and what kind of page this is. (If it is an index where newer posts will appear before older ones, this will be a big problem.)

The deduplication algorithm will run primarily on indexes. We can discuss whether or not it should be changing links from the bottom-up in that PR.

felixfontein:
Bottom-up would be a good idea, I think. This will at least make the links permanent when `INDEXES_STATIC` is `True` and you're not on the front page. If `INDEXES_STATIC` is `False`, all hope is lost anyway...

ralsina:
+1 to bottom-up

Kwpolska:
@ralsina: review welcome.

@felixfontein: alright, done on `header-deduplication` branch.

docs/manual.txt
```diff
@@ -1911,6 +1911,23 @@ jsonminify
...
 
+add_header_permalinks
+   Add links next to every header, Sphinx-style. You will need to add styling for the `headerlink` class,
+   in `custom.css`, for example:
+
+   .. code:: css
+      .headerlink { opacity: 0.1; margin-left: 0.2em; }
+      .headerlink:hover { opacity: 1; text-decoration: none; }
+
+   Additionally, you can provide a custom list of XPath expressions which should be used for finding headers (``{hx}}`` is replaced by headers h1 through h6).
```
ralsina: Unbalanced braces

Kwpolska:
Merging, thanks for the reviews.

Issue #2752, opened 2017-05-10
Make compilers use new UUID-based shortcode API


ralsina:
The new API is better. So, we should port the compilers to it.

Included

* [x] html
* [x] ipynb
* [x] markdown
* [ ] ~~pandoc~~ too complicated because of the multiple outputs
* [ ] ~~php~~ I am not doing this
* [x] rest

Plugins: v8 only, v7 is frozen.

* [x] asciidoc
* [x] bbcode
* [x] commonmark
* [x] kramdown (in plugins/ PR)
* [ ] latex (has no v8 version)
* [x] ~~markmin~~ obscure and apparently py2-only
* [x] mediawiki
* [x] misaka
* [x] mistune
* [ ] ~~odt~~ not worth the effort
* [ ] orgmode
* [ ] rest_html5 (has no v8 version)
* [x] textile
* [ ] ~~txt2tags~~ is py2 only, and v8 is py3 only so, nope.
* [ ] ~~wiki~~ The underlying library is broken, and nobody uses this
* [ ] ~~wordpress~~

felixfontein:
The WordPress compiler doesn't support Nikola shortcodes, it has its own WordPress-like shortcode system. I don't think it makes sense to include Nikola shortcodes into the WordPress compiler, since it was never thought to be used for new posts, but for processing posts of converted blogs. And adding support now will only break things, in case someone used `{{` ... `}}` in a WordPress post.

ralsina:
Sounds good to me, let's cross it.

ralsina:
The PHP plugin doesn't seem to support shortcodes and the pandoc plugin is weird.

ralsina:
Ok, I made a huuuuuge mess of things trying to do this.

I was working on the v7 plugins using master to test, and then things get messy.

I will have to test everything with v7 and see how it works.

ralsina:
did a hard reset on today's work, and will try to do it better now.

ralsina:
Annnnd I had done the same thing with asciidoc before. ARGGGGGH

Kwpolska:
Seems to be (mostly) done.

PR #2759, opened 2017-05-12
More new-shortcode support

ralsina:
Partial #2752 implementation

### DIFF
--- a/nikola/plugins/compile/html.py
+++ b/nikola/plugins/compile/html.py
@@ -31,6 +31,7 @@ from __future__ import unicode_literals
 import os
 import io
 
+from nikola import shortcodes as sc
 from nikola.plugin_categories import PageCompiler
 from nikola.utils import makedirs, write_metadata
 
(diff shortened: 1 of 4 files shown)


nikola/plugins/compile/html.py
```diff
@@ -45,7 +46,8 @@ def compile_string(self, data, source_path=None, is_two_file=True, post=None, la
         """Compile HTML into HTML strings, with shortcode support."""
         if not is_two_file:
             _, data = self.split_metadata(data)
-        return self.site.apply_shortcodes(data, with_dependencies=True, extra_context={'post': post})
+        new_data, shortcodes = sc.extract_shortcodes(data)
+        return self.site.apply_shortcodes_uuid(new_data, shortcodes, filename=source_path, with_dependencies=True, extra_context=dict(post=post))
```
Kwpolska: Why this `dict()` form instead of `{}`?

ralsina: copy/paste and the place where I copied it from used that one

PR #2763, opened 2017-05-14
Header/id deduplication

### LINKED ISSUE #2570: How to avoid duplicated ids and h1 on the blog index ?
Hi, 

I was trying to optimize my site to fix some issues in HTML and I found out that there were issues on the blog index page because of the duplicated h1 and duplicate id (the ids automatically generated by Nikola on the headers for instance). 

Is there anyway to fix this ?

Thanks

Kwpolska:
Fixes #2570. Also contains a doc fix for #2743, because I didn’t notice which branch I was committing to.

Should I bother with `<a name="">` as well?

cc @wichtounet


nikola/filters.py
```diff
@@ -437,3 +437,38 @@ def add_header_permalinks(data, xpath_list=None):
...
+        else:
+            seen_ids.add(i)
+
+    if duplicated_ids:
+        # Well, that sucks.
+        for i in duplicated_ids:
+            # Results are ordered the same way they are ordered in document
+            offending_elements = doc.xpath('//*[@id="{}"]'.format(i))
+            counter = 2
+            for e in offending_elements[1::-1]:
```
felixfontein: Shouldn't this be `offending_elements[::-1]` without the first `1`?

nikola/filters.py
```diff
@@ -437,3 +437,38 @@ def add_header_permalinks(data, xpath_list=None):
...
+                new_id = '{0}-{1}'.format(i, counter)
```
felixfontein: What happens if this is ID is in use as well?

nikola/filters.py
Kwpolska: Notice `counter = 2`? There will be `foo`, `foo-2`, `foo-3`…

nikola/filters.py
Kwpolska: Will fix.

nikola/filters.py
felixfontein: But this only enumerates elements 1 and 0 of the list, not any other element. `['a', 'b', 'c', 'd', 'e'][1::-1] == ['b', 'a']`.

You probably want `offending_elements[-2::-1]`.

nikola/filters.py
felixfontein: But trying until you find a free one will be a problem if you want to make permalinks (as for the Sphinx permalinks filter). Assume the two oldest entries have headers with IDs `a` and `a`. If there's nothing else around, one will get `a` and the other `a-2`. But now assume a new post is added which uses `a-2` (for whatever reason). Then the old `a-2` will end up as `a-3`, because `a-2` is already taken, and links pointing to `a` in the second oldest post are suddenly pointing to the wrong place.

Kwpolska: This is one of those cases which we can’t reliably fix, at least as a filter. To make it work in every scenario, we’d need to add post names to the IDs, or otherwise do unusual stuff. The thing is, most people are unlikely to use those permalinks on indexes, and this plugin’s aims are (a) to fix HTML validation issues on indexes, (b) to fix IDs for Sphinx permalinks and other uses clashing, on a single page. You can’t protect against changing permalinks if those are maintained by code, not humans, whilst allowing said humans to edit the page contents (an edit to a post/page could trigger a deduplication)

felixfontein: True. You'd need state to track the permalinks. We should mention that somewhere in the documenation, though. Otherwise we'll sooner or later get bug reports for that :)

Kwpolska:
I updated the documentation and used a different approach: going bottom→top in indexes, and top→bottom in posts/pages. This way, we can get a sensible ordering for both use cases.

ralsina:
@Kwpolska I resolved the conflicts, hope I did not mess it, feel free to push merge

nikola/filters.py
```diff
@@ -441,7 +441,54 @@ def add_header_permalinks(fname, xpath_list=None, file_blacklist=None):
...
+            # because new posts appear at the top of pages.
+            # We also leave the first result out, so there is one element with
+            # "plain" ID
+            if any(doc.find_class(c) for c in top_classes):
+                off = offending_elements[1:]
+            else:
+                off = offending_elements[-2::-1]
+            for e in off:
+                new_id = i
+                while doc.xpath('//*[@id="{}"]'.format(new_id)):
```
felixfontein: Why not use (and later update) `seen_ids` instead of hoping that the query runs as fast as possible?

nikola/filters.py
```diff
@@ -441,7 +441,54 @@ def add_header_permalinks(fname, xpath_list=None, file_blacklist=None):
...
+                    new_id = '{0}-{1}'.format(i, counter)
+                    counter += 1
+                e.attrib['id'] = new_id
+                # Find headerlinks that we can fix.
+                headerlinks = e.find_class('headerlink')
+                for hl in headerlinks:
+                    # We might get headerlinks of child elements
```
felixfontein: If one of the header links belongs to a child element with the same ID, you change the link to something wrong.

Kwpolska: How do you suggest to fix that? A `break`?

nikola/filters.py
Kwpolska: Makes sense, fixed.

Kwpolska:
@felixfontein, please explain, and perhaps suggest a solution to:

> If one of the header links belongs to a child element with the same ID, you change the link to something wrong.

felixfontein:
Maybe something like:
```
<div id="the-content">
  [...]
  <h1 id="the-content">The Content</h1><a class="headerlink" href="#the-content">¶</a>
</div>
```
(And there's another `the-content` ID somewhere so that both these end up in `off`.) If the code processes them from top to bottom, it will change it to the following:
```
<div id="the-content-2">
  [...]
  <h1 id="the-content-3">The Content</h1><a class="headerlink" href="#the-content-2">¶</a>
</div>
```
because the `<div>` is processed first.

To avoid this, you probably have to check if there's another `the-content` ID between the `headerlink` element and the element whose ID we're currently processing.

Kwpolska:
It sounds as if doing a `break` when we find the first headerlink will work, because those are supposed to be the first headerlink in a given header. (Using `headerlinks[0]` would work in 99.9% of cases, I believe). @felixfontein, please review again.

felixfontein:
In the case I sketched above it won't work, because the `<div>` has no headerlink (because it is no header), so still the wrong headerlink will be "fixed". The problem here is that the header in the post happens end up with the same ID (because of it's title) than something generated by the theme (the `<div id="the-content">`).

Kwpolska:
Please propose a better solution then? I’m inclined to say “edge case not worth fixing”.

felixfontein:
Is there a way to find the first parent object of `hl` whose ID is `id`? If yes, you could check whether it is `e`.

Kwpolska:
I’m afraid this might not be efficient enough, and that would be needed only for an edge case. This happening in real input is *very* unlikely. (reST won’t ever generate it, for example)

felixfontein:
Just manually walking up the DOM tree shouldn't be very slow. Usually DOM trees aren't excessively deep that you really get a performance hit. And since this search has only to be done for very few elements, it shouldn't really matter.

And yes, while this is an edge case, it is a bug and really shitty for a user if this happens. And if we ever fix this, it might break some existing links where someone might have worked around this bug.

In my opinion, we should get this filter into a form that should never require any changes, because every change might break existing links, which are supposed to be permalinks. But then, I probably won't be using it (in particular the headerlink feature), so I won't really mind if you decide to ignore this edge case :)

felixfontein:
Ok, just thought of one reason where this can happen more realistically in real life: if someone decides to name a subsection the same as a section. Some people tend to do that. And if you don't use a compiler which prevents colliding IDs, you have a problem.

Kwpolska:
Thanks, merged. Will rebuild getnikola.com to use this in a second.

PR #2765, opened 2017-05-14
Don't crash when translating untranslated strings

ralsina:
keturn just reported on IRC that an old theme crashed because it used `messages['Posted']` instead of the current `messages['Posted:']`  ... that's harsh :-)

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -265,6 +265,10 @@ class Functionary(defaultdict):
         """When called as a function, take an optional lang and return self[lang][key]."""
         if lang is None:
             lang = LocaleBorg().current_lang
+        if lang not in self:
+            LOGGER.warning('Unknown language %s', lang)
+        elif key not in self[lang]:
+            LOGGER.warning('Unknown message %s', key)
         return self[lang][key]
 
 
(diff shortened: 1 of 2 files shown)


nikola/utils.py
```diff
@@ -265,6 +265,10 @@ def __call__(self, key, lang=None):
         """When called as a function, take an optional lang and return self[lang][key]."""
         if lang is None:
             lang = LocaleBorg().current_lang
+        if lang not in self:
+            LOGGER.warning('Unknown language %s', lang)
+        elif key not in self[lang]:
+            LOGGER.warning('Unknown message %s', key)
```
Kwpolska: Is this enough to fix it? Perhaps you should add some fallback?

Kwpolska:
This breaks spectacularly and generates 161 error messages for the demo site (from post meta failing to be found). Reverting this.

Kwpolska:
Fixing this right is hard, since functionaries are used for post metadata, post data, and messages. But I’d argue this is too rare to bother fixing anyway. The original issue was using v6 themes, which are already pretty incompatible, and [v7.0.0](https://github.com/getnikola/nikola/releases/tag/v7.0.0) was released almost exactly 3 years ago.

ralsina:
Ok, sorry about that :-P

I could do a subclass that is only for messages and use that only in that place.

PR #2767, opened 2017-05-15
Add a file blacklist for header permalinks

Kwpolska:
Header permalinks are not desirable in some files (eg. output/index.html on getnikola.com — which I just rebuilt using header_permalinks; deduplication is pending until it’s merged). So, here’s a blacklist.


ralsina:
How about a no-permalinks metadata also?

Kwpolska:
That’s doable, but it would require mapping output locations to posts — do we have anything like that?

ralsina:
No we don't, sadly :-(

Kwpolska:
Merging as-is then.

PR #2768, opened 2017-05-15
Fix #2758 -- Add support for theme meta files

### LINKED ISSUE #2758: Meta files for themes
**To do:**

- [x] Add rudimentary support on themes.getnikola.com so at least it builds
- [x] Update documentation (document format; themes repo readme; creating-a-theme + theming docs)
- [ ] Do useful (?) stuff with meta info on themes.getnikola.com:
  - [x] index: conflate 
  - [x] theme page: show new meta
  - [x] theme page: variants
  - [ ] tag pages for stories?
- [x] Write .theme files for existing themes (and make those dependencies in themes site)
- [ ] Perhaps do useful stuff (re compatibility) in Nikola core?

---

Our current way of specifying theme settings is by using two simple text files, engine and parent. This is not flexible enough, so I propose an extensible meta file format for themes.

Some meta fields that could be included:
* parent
* engine
* author, author_url
* website
* license
* tags
* based_on (eg. Bootstrap, Casper, Lanyon for Jekyll)
* family (eg. bootstrap3, zen)
* mako_version, jinja_version
* variants (eg. bootstrap3-gradients, zen-ipython)

Use cases (sample):
* Merging Mako and Jinja2 themes on the site (via @ralsina on IRC: *I think we should collapse the mako themes that have a jinja variant together in the listing*)
* Listing available theme variants on the site
* Adding tags so themes can be browsed by features they support, frameworks they use, etc.
* Attribution for authors, frameworks, inspirations, licenses
	
File format: For consistency with plugins, and simplicity, perhaps INI? (The old files have to stay for backwards compatibility, at least until v8.)

Comments and more ideas welcome.

Kwpolska:
This is #2758. Nikola currently only uses the `parent` and `engine` values, and the themes site does not have support for these files yet. (We could also add something like MinVersion/MaxVersion support).

Old themes (without .meta files) are still supported and should work without changes.


ralsina:
LGTM very superficial review

PR #2774, opened 2017-05-18
Nicer error handling (Fix #2771)

### LINKED ISSUE #2771: Posts without metadata make nikola fail with cryptic error.
When building a Nikola site with the `ipynb` compiler enabled, any notebooks not created or imported using `nikola new_post` will raise a `ValueError` from because of the missing metadata, e.g.:

`ValueError: Invalid date '' in file posts/notes-on-meshgrid-ordering.ipynb`

This is caused by the return of an empty dictionary in [`read_metadata()`](https://github.com/getnikola/nikola/blob/master/nikola/plugins/compile/ipynb.py#L138).

I understand that this can be worked around with `nikola new_post -i` and creation of a separate metadata file, but the workflow I'd personally prefer when using Nikola is to author my notebooks in a way that is entirely Nikola-agnostic. Drawing inspiration from [what's done for page metadata](https://github.com/getnikola/nikola/blob/master/nikola/post.py#L194-L195), I kludged together a solution in `ipynb.py` that works for my use case.

After speaking with @Kwpolska a little in Freenode #python, it appears this is partially a design decision, but I think it's pretty inelegant to just propagate the `ValueError` to the user.

For my workflow, my ideal user story would be:
> I place an IPython notebook in `posts/` and run `nikola build`  
> No Nikola metadata is found  
> The build script prompts me to enter the same info `nikola new_post` would ask for  
> After completing this process, the build script continues as normal

I'm happy to maintain my own version of the ipynb compiler for my personal use, but I figured I'd find out if this is a feature that would add value to the project.  If not, then at the very least it seems appropriate to point the user to the `new_post` command when this error is raised.

ralsina:
Fix #2771 

* Log before raising exceptions
* Eat the exception at the top level unless NIKOLA_DEBUG is set.


nikola/nikola.py
```diff
@@ -1086,8 +1086,7 @@ def init_plugins(self, commands_only=False, load_all=False):
                     # Remove blacklisted plugins
                     if p[-1].name in self.config['DISABLED_PLUGINS']:
                         bad_candidates.add(p)
-                        utils.LOGGER.debug('Not loading disabled plugin {}', p[-1].name)
-                    # Remove compilers we don't use
+                        utils.LOGGER.debug('Not loading disabled plugin {}', p[-1].name)  # Remove compilers we don't use
```
Kwpolska: The removal happens inside the `if`.

nikola/__main__.py
```diff
@@ -275,13 +275,19 @@ def load_tasks(self, cmd, opt_values, pos_args):
...
+                self.nikola.gen_tasks('render_site', "Task", 'Group of tasks to render the site.'))
+            latetasks = generate_tasks(
+                'post_render',
+                self.nikola.gen_tasks('post_render', "LateTask", 'Group of tasks to be executed after site is rendered.'))
+            signal('initialized').send(self.nikola)
+        except Exception:
+            LOGGER.error('Error loading tasks')
+            if self.nikola.debug:
+                raise
+            return {}, {}
```
Kwpolska: Won’t this make .doit.db empty or anything?

ralsina: Good point

nikola/__main__.py
```diff
@@ -363,7 +369,10 @@ def run(self, cmd_args):
                 LOGGER.error("This command needs to run inside an "
                              "existing Nikola site.")
                 return 3
-        return super(DoitNikola, self).run(cmd_args)
+        try:
+            return super(DoitNikola, self).run(cmd_args)
+        except Exception:
+            pass
```
Kwpolska: return 1

nikola/__main__.py
Kwpolska: Perhaps just let the exception bubble upwards and handle it somewhere else.

nikola/__main__.py
```diff
@@ -363,7 +371,41 @@ def run(self, cmd_args):
...
-        return super(DoitNikola, self).run(cmd_args)
+        try:
+
+            # get "global vars" from cmd-line
+            args = self.process_args(cmd_args)
+
+            # get specified sub-command or use default='run'
+            if len(args) == 0 or args[0] not in sub_cmds:
+                specified_run = False
+                cmd_name = 'run'
```
Kwpolska: I don’t think this applies to Nikola… We default to `help` via magic.

ralsina:
Well, it still does. I'll remove this.

nikola/__main__.py
```diff
@@ -363,7 +371,33 @@ def run(self, cmd_args):
                 LOGGER.error("This command needs to run inside an "
                              "existing Nikola site.")
                 return 3
-        return super(DoitNikola, self).run(cmd_args)
+        try:
```
Kwpolska: Is copy-pasting code really necessary for this? Would `sys.exit(1)` hurt here?

ralsina:
That is an alternative, yes. If you prefer that I can change.

nikola/__main__.py
ralsina: Looking at it better, I would have to exit(1) in the next layer, since this run() prints the exceptions.

ralsina: Hell, even that won't work, it will catch exit() !!!!

nikola/__main__.py
ralsina: Turns out it does.

ralsina:
Turns out no, it's an "except Exception:" so it would work. Will do it
tomorrow, probably.

nikola/__main__.py
ralsina: Ok, fixed.

nikola/__main__.py
```diff
@@ -289,7 +285,7 @@ def load_tasks(self, cmd, opt_values, pos_args):
             signal('initialized').send(self.nikola)
         except Exception:
             LOGGER.error('Error loading tasks')
-            raise
+            sys.exit(3)
```
felixfontein: This is terrible: in case something happens, there is no way to see WHAT happened without modifying Nikola's source code! I've defined `NIKOLA_DEBUG`, but I still just get `Error loading tasks`.

I had to undo this change to get some useful output.

Kwpolska: Added a conditional for debug in fc391082.

felixfontein: Thanks!

Kwpolska:
Hm, could we re-discuss this? Because in some cases, we will end up printing “Error loading tasks” without any explanation. How would people know what to fix? How would bug reports look?

ralsina:
In that case, clearly we are raising exceptions that we are not logging, right?

So, for that, we use NIKOLA_DEBUG=1 

Also, I will check a little to see if we still have unlogging raises.

Kwpolska:
It doesn’t have to be Nikola who’s raising those exceptions. My unhelpful errors were caused by Mako syntax errors. We should at least print the exception text, and tell the user how to get full debug info.

felixfontein:
Yes, a bit more information would be really helpful! It can always happen that we forget to write a message at some point in the future, and then some user gets a really unhelpful error message.

Another thing which might be worth discussing: how about printing a short information text that you can get debug info with `NIKOLA_DEBUG=1` in case an exception is caught? Then some more interested persons (power-users, hackers, ...) don't have to search the documentation (or the source code) first before being able to get more information.

PR #2776, opened 2017-05-19
Flowr no jquery

ralsina:
This branch makes Flowr.js work without JQuery, and in the base theme. Therefore:

* The base gallery looks more like the bootstrap gallery
* All themes derived from base get a nicer gallery

I am hesitant to make the bootstrap theme use this, since it may break someone's derived theme.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Ported gallery image layout to base theme (Issue #2775)
 * Use ``.theme`` files to store theme metadata (Issue #2758)
 * New ``add_header_permalinks`` filter, for Sphinx-style header links
   (Issue #2636)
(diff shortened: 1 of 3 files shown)


Kwpolska:
@ralsina flowr.js should be a symlink. Otherwise, it’s not kept up-to-date with Bower, and stuff like this happens.

ralsina:
@Kwpolska updated from where? It's basically a fork at this point.

Kwpolska:
Wait, so you hacked flowr not to require jQuery?

ralsina:
Exactly :-)

Kwpolska:
IMO it’d be better to focus on #2777 than use a stop-gap measure. But if you want this now, feel free to merge with no-reviews override.

ralsina:
That's also happening soon, it's orthogonal. Merging, then.

PR #2782, opened 2017-05-21
Show last exception text and NIKOLA_DEBUG explanation (#2774 #2771)

Kwpolska:
This improves #2771 and #2774. It shows the text of only the last exception, and advertises NIKOLA_DEBUG=1.

(We can’t rely on logging from other places, because not all exceptions come from within Nikola itself)

### DIFF
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -284,9 +284,10 @@ class NikolaTaskLoader(TaskLoader):
                 self.nikola.gen_tasks('post_render', "LateTask", 'Group of tasks to be executed after site is rendered.'))
             signal('initialized').send(self.nikola)
         except Exception:
-            LOGGER.error('Error loading tasks')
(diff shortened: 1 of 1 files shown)


nikola/__main__.py
```diff
@@ -410,5 +413,12 @@ def levenshtein(s1, s2):
     return previous_row[-1]
 
 
+def _print_exception():
+    """Print an exception in a friendlier, shorter style."""
+    etype, evalue, _ = sys.exc_info()
+    LOGGER.error(''.join(traceback.format_exception(etype, evalue, None, limit=0, chain=False)).strip())
+    LOGGER.notice("To see more details, run Nikola in debug mode (set environment variable NIKOLA_DEBUG=1)")
```
felixfontein: Maybe only print this if debug mode isn't already enabled?

felixfontein: Ah, ignore that. This is only called in case `NIKOLA_DEBUG` is not 1.

PR #2787, opened 2017-05-22
Fix 2780

ralsina:
We have a switch to make panoramas look better, but no way to enable it. At least it should the turned on :-P

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -16,6 +16,7 @@ Features
 Bugfixes
 --------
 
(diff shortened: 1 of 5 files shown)


ralsina:
Oops, branched out of the wrong place.

Kwpolska:
Off-topic, that YAML looks better than I thought. The JSON-y dict/array form will save us some headaches.

ralsina:
@Kwpolska it's not awful, I wrote/copied the 1300 lines in like, one hour :-)

PR #2788, opened 2017-05-22
Better panorama thumbnails

### LINKED ISSUE #2780: Thumbnail too small for (very) wide images

**Python Version:** Python 3.5

**Nikola Version:** 7.8.5

### Description:

Sometimes I stich images. The thumbnail generated is too small, for instance, in this case:

http://grulicueva.homenet.org/~mdione/nikola-gallery/galleries/Trips/2016/12-24%20-%20Isola%202000/

The thumbnail is 180px × 24px, but it's being shown (in my computer) with this CSS style:

`<div class="image-block" title="" data-toggle="tooltip" data-original-title="12 24t13 05 23" style="width: 1130px; height: 151px; margin-bottom: 5px; margin-left: 0px;"><a href="2016-12-24T13.05.23.jpg" class="image-reference cboxElement"></a></div>`

so it's scaling it 6x at least. Maybe this could be fixed with a minimum height x width instead of width alone, but I understand it's a complex problem.

ralsina:
Fix #2780

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -16,6 +16,7 @@ Features
 Bugfixes
 --------
 
+* Enable larger panorama thumbnails (Issue #2780)
 * Disable ``archive_rss`` link handler, which was useless because no
   such RSS was ever generated (Issue #2783)
 * Ignore files ending wih "bak" (Issue #2740)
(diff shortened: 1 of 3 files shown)


ralsina:
@Kwpolska they already are kwargs, we are just calling it positionally :-P

I don't want to run git blame because I fear this brokedness is all my doing.

Kwpolska:
Well, we weren’t [*always*][x] calling it positionally… It’s a simple mistake. No hard feelings.

[x]: https://github.com/getnikola/nikola/commit/8059f5e43c98192cc57d673024eb6368268cc0e5

PR #2791, opened 2017-05-22
Use BaguetteBox in base theme

ralsina:
This adds a nice, lightweight lightbox to the base theme.

Very conservative branch: no changes in bootstrap3, no captions, just the very barebones support.


nikola/data/themes/base/templates/base.tmpl
```diff
@@ -20,7 +20,13 @@ ${template_hooks['extra_head']()}
          ${footer.html_footer()}
     </div>
     ${base.late_load_js()}
+    % if use_cdn:
+        <script src="//cdnjs.cloudflare.com/ajax/libs/baguettebox.js/1.8.2/baguetteBox.min.js"></script>
```
Kwpolska: https://

Kwpolska:
Are captions supported if we wanted them?

nikola/data/themes/base/templates/base.tmpl
ralsina: This keeps it protocol-relative, is https better? It used to give a mixed-content warning :-P

Kwpolska: protocol-relative stopped being cool [a while back](https://www.paulirish.com/2010/the-protocol-relative-url/). Using httpS never gave mixed-content warnings, only http did.

ralsina:
@Kwpolska yes, they are

nikola/data/themes/base/templates/base.tmpl
ralsina: Ok, https it is

ralsina:
I have NO IDEA why this is failing.

Kwpolska:
(PS. CSSLint is pretty crazy, so I nuked it.)

PR #2795, opened 2017-05-24
Guess locale using locale.getdefaultlocale()

Kwpolska:
Credit for discovering this goes to @madduck. Will avoid some more “cannot guess locale” errors.


madduck:
Thanks for the quick turnaround. I've tested this with a mix of `en_NZ.UTF-8` and `de_DE.UTF-8` and it does what it should.

PR #2799, opened 2017-05-25
Fix 2798

### LINKED ISSUE #2798: nikola new_post should guess formats from extensions
* Running `nikola new_post -i /path/to/foo.ipynb` will create something like `posts/foo-title.rst`, which will result in completely broken output (turns out reST isn’t really good at rendering JSON not in a code block)
* Running `nikola new_post posts/bar.ipynb` will create reST-style metadata in that file, instead of a JSON file for ipynb.

(Replace *reST, ipynb* with *default compiler, any other compiler*)

ralsina:
Guess content format from path or imported file on `nikola new_post`

Fixes #2798

### DIFF
--- a/nikola/plugins/command/new_post.py
+++ b/nikola/plugins/command/new_post.py
@@ -273,6 +273,14 @@ class CommandNewPost(Command):
                 if extension in extensions:
                     content_format = compiler
 
+        elif not content_format and import_file:
+            # content_format not specified. If import_file was given, use
+            # it to guess (Issue #2798)
+            extension = os.path.splitext(import_file)[-1]
+            for compiler, extensions in self.site.config['COMPILERS'].items():
+                if extension in extensions:
+                    content_format = compiler
+
         elif not content_format:  # Issue #400
             content_format = get_default_compiler(
                 is_post,
--- a/nikola/plugins/command/new_post.py
+++ b/nikola/plugins/command/new_post.py
@@ -265,16 +265,25 @@ class CommandNewPost(Command):
         if "@" in content_format:
             content_format, content_subformat = content_format.split("@")
 
-        if not content_format:  # Issue #400
+        if not content_format and path:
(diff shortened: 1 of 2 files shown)


nikola/plugins/command/new_post.py
```diff
@@ -265,16 +265,33 @@ def _execute(self, options, args):
...
+                    content_format = compiler
+
+        elif not content_format:  # Issue #400
             content_format = get_default_compiler(
                 is_post,
                 self.site.config['COMPILERS'],
                 self.site.config['post_pages'])
 
-        if content_format not in compiler_names:
+        elif content_format not in compiler_names:
```
Kwpolska: Perhaps that `if` was there for a reason? (I can’t see it though.)

nikola/plugins/command/new_post.py
```diff
@@ -265,16 +265,33 @@ def _execute(self, options, args):
...
-        if not content_format:  # Issue #400
+        if not content_format and path:
+            # content_format not specified. If path was given, use
+            # it to guess (Issue #2798)
+            extension = os.path.splitext(path)[-1]
+            for compiler, extensions in self.site.config['COMPILERS'].items():
+                if extension in extensions:
+                    content_format = compiler
+
+        elif not content_format and import_file:
```
Kwpolska: Perhaps we could merge that and the first if into one?

nikola/plugins/command/new_post.py
ralsina: I checked, this seems to work just as well :)

nikola/plugins/command/new_post.py
ralsina: Perhaps, but it saves 3 lines of code at the cost of a non-stupid condition. I like stupid conditions.

Kwpolska:
Thanks for fixing this!

Issue #2801, opened 2017-05-28
Support Hugo/Jekyll/Pelican style metadata in posts/page


ralsina:
Our arbitrary metadata format have some things going for it:

* Since we always wrap it in comments it can be processed by default toolchains
* It has worked for years

OTOH:

* It's confusing to newcomers
* It's incompatible with all other SSGs so switching to Nikola is harder than it needs be
* It's incompatible with all other SSGs so switching OUT of Nikola is harder than it needs be

So, let's make it so at least YAML works.

Hugo and Jekyll detect YAML metadata because it's wrapped in "---" and at the top of the file.
TOML metadata is wrapped in '+++'.
~~JSON metadata is wrapped in "{  }"~~

- [x] Add naïve detection and loading of YAML metadata
- [x] Add naïve detection and loading of TOML metadata
- [x] ~~Add naïve detection and loading of JSON metadata~~ withdrawn
- [ ] Support metadata-in-comments for compilers that use external tools where the metadata can't be easily stripped (nice to have)
- [x] Support metadata from markdown frontmatter, HTML meta headers, reSt field list Pelican style (Issue #1923)
- [x] Handle empty lines in metadata that supports it
- [x] Support generating new posts/pages with YAML/TOML/docinfo/md/~~JSON~~ metadata
- [x] Handle typed data (example: tags as a list instead of a comma-separated string)

ralsina:
Making the last item its own issue and closing this larger one.

PR #2802, opened 2017-05-28
Support YAML metadata (Issue #2801)

ralsina:
First step in #2801 : Naïve support for YAML metadata.

- Optional yaml import
- Fail only if required by posts
- Parse just like our traditional metadata

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -51,6 +51,10 @@ try:
     import pyphen
 except ImportError:
     pyphen = None
+try:
+    import yaml
+except ImportError:
+    yaml = None
 
 from math import ceil  # for reading time feature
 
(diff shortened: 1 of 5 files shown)


nikola/post.py
```diff
@@ -1023,6 +1027,18 @@ def _get_metadata_from_file(meta_data):
     if not meta_data[0]:
         meta_data = meta_data[1:]
 
+    # If 1st line is '---', then it's YAML metadata
+    if meta_data[0] == '---':
+        if yaml is None:
+            utils.req_missing('pyyaml', 'use YAML metadata', optional=True)
+            raise ValueError('Error parsing metadata')
+        idx = meta_data.index('---', 1)
+        meta = yaml.load('\n'.join(meta_data[1:idx]))
```
Kwpolska: Always use `yaml.safe_load`.

ralsina: Ok

felixfontein:
Shouldn't you also adjust the `split_metadata` function in `PageCompiler`? The new YAML (and TOML?) metadata allows empty lines between starting and ending `---` (resp. `+++`?).

ralsina:
@felixfontein that's why this is a naive implementation, I don't want to spend two days thinking of corner cases :-)

PR #2803, opened 2017-05-28
Toml metadata

ralsina:
2nd branch in the #2801 story: naive support for TOML metadata

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -51,6 +51,10 @@ try:
     import pyphen
 except ImportError:
     pyphen = None
+try:
+    import yaml
+except ImportError:
+    yaml = None
 
 from math import ceil  # for reading time feature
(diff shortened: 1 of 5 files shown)


nikola/post.py
```diff
@@ -1023,6 +1031,28 @@ def _get_metadata_from_file(meta_data):
     if not meta_data[0]:
         meta_data = meta_data[1:]
 
+    # If 1st line is '---', then it's YAML metadata
```
Kwpolska: Perhaps it should be more generic? Also, how will things like reST react to metadata in this format?

ralsina: It works just fine, because we are removing metadata before feeding it to docutils.

It does mean you can't use docutils rst2html with a post anymore.

Issue #1923, opened 2015-08-03
Support reading pelican posts directly


ralsina:
This came up in the podcast I was in along with a pelican guy... why is our post/data/metadata format different?

I know _why_ ours is the way it is, but hey, why not support their version? Most metadata is exactly the same anyway.  This would reduce  vendor lock-in and is The Nice Thing To Do (in general).

- [x] Support markdown metadata
- [x] Support reSt metadata
- [x] Support HTML meta tags

Kwpolska:
:+1: I sort of wanted to create this issue in the past, with a title along the lines of “simplify world domination”.

But please don’t use crazy regexps for that. They use reST `:fields:`, and we can probably talk docutils into giving us that metadata (probably need to look at Pelican code).  Bonus points for implementing that as compiler metadata, and even more bonus points for getting the page title this way (and removing the hacky code).

schettino72:
link to podcast ?

ralsina:
@schettino72 it's not published yet.

ralsina:
@Kwpolska for markdown they use something that looks like ours except not in a comment and without the ".. " (of course).

We'll also need to remove all that stuff from the output to avoid repeating stuff, so ... tricky.

Kwpolska:
Stealing (at least for reST).

ralsina:
@Kwpolska Be my guest, I am not touching this for a few days at least.

Kwpolska:
* need Publisher (and some refactoring around rst2html and friends; cached instance)
* Try to get id for `title` and `docinfo` from `pub.document.first_child_matching_class(docutils.nodes.XXX)`
* **Title**
  1. Get `title` via `pub.document.children[title_id]` (if it is not None)
  2. Use title.astext() to get the title (loses formatting)
* **Metadata**
  1. Get `docinfo` via `pub.document.children[docinfo_id]` (if it is not None)
  2. Iterate over docinfo.children. There are multiple types of field objects. The type of the current field object needs to be checked using `NAME = field.tagname`:
    * authors: `field.child_text_separator = ', '; VALUE = field.astext()`
    * field [reST fallback]: `NAME = field.children[0].astext(); VALUE = field.children[1].astext()`
    * address, author, contact, copyright, date, organization, revision, status, version [or just `else`]: `VALUE = field.children[1].astext()`
* **Post-processing**
  * A mapping needs to be used to translate Pelican/docutils names into Nikola names. Converting names to lowercase might also be useful.
[...]

PR #2805, opened 2017-05-28
Pelican style Markdown metadata (Issue #1923)

ralsina:
This uses the markdown "meta" extension to read metadata in a manner compatible with Pelican (Issue #1923)

It still presents a few problems:

* Performance: file is read and converted twice.
* ~~This doesn't really support translated sources, same problem is present in ipynb plugin, because it uses
  source_path instead of translated_source_path. OTOH, translated_source_path doesn't work yet at this stage :-P~~


nikola/plugins/compile/markdown/__init__.py
```diff
@@ -134,3 +141,17 @@ def create_post(self, path, **kw):
...
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """Read the metadata from a post, and return a metadata dict."""
+        if not self.support_metadata:
+            return {}
+        if Markdown is None:
+            req_missing(['markdown'], 'build this site (compile Markdown)')
+        if lang is None:
+            lang = LocaleBorg().current_lang
+        # FIXME: what about translations????
+        source = post.source_path
```
Kwpolska: `post.translated_source_path(lang)`

Also, note that Pelican uses Title-Cased metadata fields, and has some different names.

ralsina: translated_source_path sort of crashes, I need to debug that. Code in ipynb plugin has the same problem.

I'll probably lower-case the keys then.

ralsina:
From invariance results, it seems this is taking priority over actual explicit metadata (at least for title) which is not good, I think.

nikola/post.py
```diff
@@ -153,15 +153,17 @@ def __init__(
...
+
         default_metadata, self.newstylemeta = get_meta(self, self.config['FILE_METADATA_REGEXP'], self.config['UNSLUGIFY_TITLES'])
 
         self.meta = Functionary(lambda: None, self.default_lang)
         self.meta[self.default_lang] = default_metadata
 
-        # Load internationalized metadata
         for lang in self.translations:
-            if os.path.isfile(get_translation_candidate(self.config, self.source_path, lang)):
-                self.translated_to.add(lang)
```
Kwpolska: Perhaps we should add a fallback to `get_translation_candiate` if `self.translated_to` is empty? There also seems to be a `lang != self.default_lang` case that will return the source path (shouldn’t it be flipped?)

nikola/plugins/compile/rest/__init__.py
```diff
@@ -59,6 +59,38 @@ class CompileRest(PageCompiler):
...
     logger = None
 
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """Read the metadata from a post, and return a metadata dict."""
+        if lang is None:
+            lang = LocaleBorg().current_lang
+        source = post.translated_source_path(lang)
+
+        with io.open(source, 'r', encoding='utf-8') as inf:
+            document = docutils.core.publish_doctree(
```
Kwpolska: I’m not a reST expert, but perhaps there is a way to make this reusable and store it for `compile_string`?

Also, we might need to remove field lists from the output, if metadata is imported.

.gitignore
```diff
@@ -195,5 +195,5 @@ Session.vim
 *~
 
 # Demo for Bagguette
```
Kwpolska: this should be at the top, before the gitignore.io template starts

Kwpolska:
```diff
-<meta property="article:published_time" content="2012-03-30T23:00:00-03:00">
+<meta property="article:published_time" content="2013-12-31T23:59:59Z">
```

Somehow, the `.. date:` from Nikola comment metadata is ignored and replaced by the invariant date ([over here](https://github.com/getnikola/nikola/blob/2623322c6315b0fa30e9dd1b78e64ddf2e3bb14a/nikola/post.py#L200)).

Kwpolska:
Sites using `post-list` crash:

```pytb
  File "/Users/kwpolska/git/nikola/nikola/plugins/compile/rest/post_list.py", line 181, in run
    filename = self.state.document.settings._nikola_source_path
AttributeError: 'Values' object has no attribute '_nikola_source_path'
```

Possible fixes: set that attribute in the metadata thing; disable directives in the metadata thing; make that line failsafe. (@ralsina, you decide)

nikola/plugins/compile/rest/__init__.py
ralsina: Yeah to both things. We can probably keep the doctree somewhere in the Post object and reuse it if it's there.

Same about markdown, actually.

nikola/plugins/compile/rest/__init__.py
```diff
@@ -59,6 +59,38 @@ class CompileRest(PageCompiler):
     demote_headers = True
     logger = None
 
+    def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False, lang=None):
+        """Read the metadata from a post, and return a metadata dict."""
+        if lang is None:
+            lang = LocaleBorg().current_lang
+        source = post.translated_source_path(lang)
+
+        with io.open(source, 'r', encoding='utf-8') as inf:
```
ralsina: Good point.

nikola/post.py
ralsina: Not really sure. I'll try to take a harder look.

.gitignore
ralsina: Ok, changing it.

.gitignore
```diff
@@ -14,6 +17,8 @@ output/
 
 # All of .idea should be ignored
 .idea/
+.vscode/
```
ralsina: And I really like it! except the git "support" which is garbage.

PR #2810, opened 2017-06-02
Introduce core theme.

### LINKED ISSUE #2744: base theme has too much stuff

**Nikola Version:** 7.8.4

### Description:

The `base` theme uses a number of bootstrap specific css classes and html tag conventions that are not necessarily suitable / desirable for use in building out new themes. In my usage, building out a [skeleton](http://getskeleton.com/) based theme, I found that I had to fork the `base` theme and strip a bunch of stuff out, before I could properly derive my own theme.

I'd be happy to unify my efforts with mainline `base` and push out a PR that further strips things down. Would other people be interested in seeing this?

gwax:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Closes #2744 

Introduces new `core` and `core-jinja` themes to serve as lower level starting points for theme developers. These themes remove nearly all style and js based functionality from base. Additionally, `base` and `base-jinja` are redirected to derive from `core` and `core-jinja` respectively.


Kwpolska:
GitHub makes it a bit of a pain to review, so here are comments:

* I’m not sure if `nikola_ipython.css` can be removed from `core`.
* `ipython.min.css` should still be there
* `rst.css` bootstrap-esque styles should still be there, at least in bootstrap3, maybe in base
* `fancydates.js` should be included, or at least available in core/assets. Fancy dates are an advertised feature (in conf.py), so it shouldn’t be limited in non-base themes
* How is `aria-label="${messages('(active)', lang)}"` handled by screen readers?
* Why html5shiv-printshiv?

Kwpolska:
I’ve had some time to think about this. You see, the existence of a `core` and `base` theme side by side can be confusing. If someone just looks at the theme list, they won’t be able to tell the difference between those two. So, perhaps it would be better to swap the roles here? I think that it would work better if `base` was still the root theme in our hierarchy, with all the “opinions” (baguettebox/flowr/whatever). And then there would be a `base-micro` theme, with no styling and such. Now, you may say that this would still copy the unwanted assets into sites — and I think this solvable by making it possible for themes to ignore assets from parent themes. I created #2812 for that (regardless of what happens with the `core` theme).

gwax:
@Kwpolska Those are pretty solid points and, for me, they raise the question of what the purpose of the `base` theme is in the first place. It also calls to question why `base` and `bootstrap3` live in `nikola` while all of the other themes live in `nikola-themes`.

Some of this may be better to discuss in a v8 context where breaking changes might be permitted in themes.

Kwpolska:
`bootstrap3` is our default theme, and it’s a good-enough appearance. Nikola’s philosophy is “install and go” — you do not need to look for a theme and other stuff to create a site (and that is not going to change). We also need `base` to provide some stuff to all themes, including non-bootstrap ones.

Including other themes would increase bloat (and our packages are already quite large) and introduce a burden. You see, some of the themes in the index are not well supported, were contributed by other people, and we don’t have the resources to look after all of them. So we have a repository that everyone can contribute to, and where everyone can find a theme they like.

v8 should probably happen one day, but we’re a bit wary of just making a release to delete stuff… we probably should get it done soon-ish though. That said, I’m not sure I fully agree with your vision for `base`/`core`, whatever the names.

Kwpolska:
I‘m closing this. If you still want to make a `base-micro` theme (what would the difference be?), go ahead.

PR #2813, opened 2017-06-02
Fix #2812 — support ignoring assets in themes

### LINKED ISSUE #2812: Ignoring assets
Use cases:

* no more colorbox-i18n mess
* simpler base themes (cf. #2744, #2810)
* one lightbox is enough

Implementation: `.theme` files, of course!

Kwpolska:
This is #2812. cc @gwax (regarding making base smaller)

This implements the feature, without adding any ignores just yet. (baguettebox might be the first)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,9 @@ New in master
 Features
 --------
 
+* Support ignoring assets via ``ignore_assets`` theme meta field
+  (Issue #2812)
+* Ignore unused Colorbox locales (Issue #2812)
 * Support for reStructured text docinfo metadata with
   USE_REST_DOCINFO_METADATA option, defaulting to False (Issue #1923)
 * New HIDE_REST_DOCINFO option, defaulting to False.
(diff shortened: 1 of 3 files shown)


gwax:
Ignoring individual assets interacts poorly with changes to an underlying theme; say a new asset is added to base. If the person adding the asset forgets to ignore it in a child theme, everyone that uses the child theme will start receiving the asset.

Of course, it's also potentially problematic if they don't. What happens if `base` adds an asset and requires it in a template. What if the child doesn't want the asset but hasn't overridden the template.

gwax:
Fundamentally, we're fighting against a [fragile base class](https://en.wikipedia.org/wiki/Fragile_base_class) problem.

ralsina:
@gwax

If base adds a new asset, and the child theme doesn't need it, it's on the child theme author to fix it, or some random asset gets published and 50kb get wasted.

I suppose we could just add a ** blacklist, where the child theme just says "give me no assets" and everything is manual from then on.

And yes, the other problem also exists, so changing the base theme can break the child in that case. OTOH, if the child theme blocks assets, then it's even more the author's task to handle that.

Honestly, I would block nothing and live with a few random extra assets not linked to anything in the output. Seems to me like a really minor thing to worry about :-)

Issue #2475, opened 2016-08-26
Switch config to YAML


ralsina:
Rationale:

* We currently can't introduce new options in the user's config
* We can't migrate configs when we change options
* We could do those things if the config file was not python

Plan:
- [ ] Write a conf.yaml that matches current conf.py (including comments)
- [ ] Make `nikola init` generate a conf.yaml and no conf.py
- [ ] If there is a conf.py, read it.
- [ ] If there is no conf.py read the conf.yaml

We can even group settings by thematic sections, and just smush them smartly together on reading so we don't need to reflect the sectioning in our code.

**NOTE** for this to be worthwhile, we need to have comments that survive roundtrips. While pyyaml doesn't support that, https://pypi.python.org/pypi/ruamel.yaml/0.6 does

**NOTE** Previously this issue suggested TOML. No, TOML sucks because of it's table syntax.

felixfontein:
If I understand it correctly, currently the only way to define your own filter is to define it in your config as a Python function. That won't work with YAML.

So if I didn't understood this wrongly, we need some way to provide filters by plugins.

ralsina:
IIRC we already have a way to specify filters in metadata which works. It's not perfect, of course.

Also, a way to tweak things via conf.py for advanced usage is still doable.

felixfontein:
In metadata of what? Posts? But I want to specify a custom filter for *every* HTML file, for example, or for JS files (which are not geneated by posts). I don't see how I can do that curently without writing code in the config file.

Kwpolska:
Huge :-1: for YAML. Vim and VSCode both have issues with editing YAML and indenting it right. Also, YAML is plagued by ambiguity. For example, values are strings, unless they aren’t, and sometimes they need quotes. Multi-line strings have two syntaxes. Single characters can drastically change the meaning of values (`|` vs `>`, `-` vs `?`).

YAML may look easy, but to me it’s more complicated than the existing .py file.

ralsina:
@felixfontein which is why I am working on it. I need to see if there is a way to do this without losing interesting functionality. As it is now, there is absolutely no functionality loss. Have some patience :-)

@Kwpolska at least we can change a value in YAML from Nikola and write it back to disk, which we can't in python. Having that would make our whole deprecation strategy much cleaner.

felixfontein:
@ralsina: I'm just trying to bring up things which could be problematic as early as possible :)

richieadler:
[...]

PR #2819, opened 2017-06-04
Allowing to register filters.

felixfontein:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Allows config plugins to register filters. Registers default filters with the names `filters.<name>`.

This implements the first part of the filters discussion in #2475 and is the first step of removing Python code from `conf.py`. This PR should *not* handle the shortcuts which allow to include Python code for filters (and global context) without creating plugins; that's for another PR.


nikola/plugins/misc/default_filters.py
```diff
@@ -0,0 +1,52 @@
...
+    def set_site(self, site):
+        """Set site, which is a Nikola instance."""
+        super(DefaultFilters, self).set_site(site)
+        # Filter names are registered by formatting them by the following string
+        filter_name_format = 'filters.{0}'
+        for filter_name, filter_definition in filters.__dict__.items():
+            # Ignore objects whose name starts with an underscore, or which are not callable
+            if filter_name.startswith('_'):
+                continue
+            if not hasattr(filter_definition, '__call__'):
```
Kwpolska: `if not callable(filter_definition):`

nikola/plugins/misc/default_filters.py
```diff
@@ -0,0 +1,52 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+# PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+# OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"""Registers all default filters."""
```
Kwpolska: It doesn’t need to be a plugin. Put this in nikola.py.

nikola/nikola.py
```diff
@@ -414,6 +414,7 @@ def __init__(self, **config):
         self._template_system = None
         self._THEMES = None
         self._MESSAGES = None
+        self._filters = {}
```
Kwpolska: Underscore not needed.

nikola/nikola.py
```diff
@@ -1183,8 +1173,31 @@ def init_plugins(self, commands_only=False, load_all=False):
...
+        # Check with registered filters
+        for actions in self.config['FILTERS'].values():
+            for i, f in list(enumerate(actions)):
+                if isinstance(f, str):
+                    # Check whether this denotes a registered filter
+                    filter = self._filters.get(f)
+                    if filter is not None:
+                        actions[i] = filter
+
+        # Configure filters
```
Kwpolska: Merge the two loops.

nikola/nikola.py
```diff
@@ -1982,6 +1995,18 @@ def rel_link(self, src, dst):
...
 
+    def register_filter(self, filter_name, filter_definition):
+        """Register a filter.
+
+        filter_name should be a name not confusable with an actual
+        executable. filter_definition should be a callable accepting
+        one argument (the filename).
+        """
+        prev_filter = self._filters.get(filter_name)
+        if prev_filter is not None:
```
Kwpolska:
```py
if filter_name in self.filters:
```

nikola/plugins/misc/default_filters.py
felixfontein: Fixed.

nikola/plugins/misc/default_filters.py
felixfontein: Wasn't the idea in Nikola to move more stuff into plugins to make everything more modular? If this wouldn't need to be backwards compatible, I'd move the filter definitions into the plugin as well.

But if you prefer that, I can remove the plugin.

nikola/nikola.py
felixfontein: Fixed.

nikola/nikola.py
felixfontein: Fixed.

nikola/nikola.py
felixfontein: Fixed.

nikola/plugins/misc/default_filters.py
Kwpolska: Yeah, modularity is good, but we don’t need modularity for something there is no real need/reason for end users to disable. Post scanners are plugins, because you can have other ways of scanning posts. Setting up globals doesn’t need to be a plugin, because everyone needs those. Plus, reading files can be costly (on Raspberry Pi SD cards, for example)

felixfontein: Ok, changed that.

felixfontein:
Thanks for merging!

Issue #2817, opened 2017-06-04
Map Pelican/Hugo metadata to Nikola names


Kwpolska:
In an effort to simplify world domination, we should also allow for mapping rest/markdown/yaml/toml metadata from Pelican/Hugo names to Nikola names (eg. Pelican: summary → description, modified → updated; users would specify their own maps)

PR #2820, opened 2017-06-04
Implement metadata mapping

Kwpolska:
* Support for mapping metadata from other formats to Nikola names using the `METADATA_MAPPING` setting (Issue #2817)
* Accept tag metadata as lists for YAML/TOML (Issue #2801)
* reST docinfo: put 'authors' info in 'author' as well (kinda ugly with multiple authors, but Nikola doesn’t have any specific support for those (yet))

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -2103,7 +2103,8 @@ def rss_writer(rss_obj, output_path):
 def map_metadata(meta, key, config):
     """Map metadata from other platforms to Nikola names.
 
-    This uses the METADATA_MAPPING setting (via ``config``) and modifies the dict in place."""
+    This uses the METADATA_MAPPING setting (via ``config``) and modifies the dict in place.
+    """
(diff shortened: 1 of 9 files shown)


nikola/utils.py
```diff
@@ -2097,3 +2098,12 @@ def rss_writer(rss_obj, output_path):
         if isinstance(data, bytes_str):
             data = data.decode('utf-8')
         rss_file.write(data)
+
+
+def map_metadata(meta, key, config):
+    """Map metadata from other platforms to Nikola names.
+
+    This uses the METADATA_MAPPING setting (via ``config``) and modifies the dict in place."""
+    for foreign, ours in config['METADATA_MAPPING'].get(key, {}).items():
```
felixfontein: `config` can be `None`, when `_get_metadata_from_file` is called without `config` specified.

Kwpolska: Fair point (although it’s very unlikely). Fixed on master.

PR #2823, opened 2017-06-05
Choose metadata format

ralsina:
New METADATA_FORMAT setting, lets you choose how metadata will look on nikola new_post. Part of #2801

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -527,7 +527,7 @@ Markdown metadata
 
 Markdown Metadata only works in Markdown files, and requires the ``markdown.extensions.meta`` extension
 (see `MARKDOWN_EXTENSIONS <#markdown>`__). The exact format is described in
-the `markdown metadata extension docs <https://pythonhosted.org/Markdown/extensions/meta_data.html>`__
+the `markdown metadata extension docs. <https://pythonhosted.org/Markdown/extensions/meta_data.html>`__
 
 .. code:: text
 
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -421,6 +421,14 @@ other static site generators. The currently supported metadata formats are:
 
 You can add arbitrary meta fields in any format.
 
+When you create new posts, by default the metadata will be created as reST style comments.
+If you prefer a different format, you can set the ``METADATA_FORMAT`` to one of these values:
+
+* ``"Nikola"``: reSt comments wrapped in a comment if needed (default)
+* ``"YAML"``: YAML wrapped in "---" 
+* ``"TOML"``: TOML wrapped in "+++"
+* ``"Pelican"``: Native markdown metadata or reSt docinfo fields. Nikola style for other formats.
+
 reST-style comments
 ```````````````````
 
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -962,7 +962,7 @@ class Nikola(object):
             self._GLOBAL_CONTEXT['transition'] = config.get('THEME_REVEAL_CONFIG_TRANSITION', 'cube')
 
         # The pelican metadata format requires a markdown extension
-        if config.get('METADATA_FORMAT', 'nikola') == 'pelican':
+        if config.get('METADATA_FORMAT', 'nikola').lower() == 'pelican':
             if 'markdown.extensions.meta' not in config.get('MARKDOWN_EXTENSIONS', []):
                 utils.LOGGER.warn('To use the pelican metadata format you may need to add "markdown.extensions.meta" to your MARKDOWN_EXTENSIONS setting.')
 
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -962,9 +962,11 @@ class Nikola(object):
             self._GLOBAL_CONTEXT['transition'] = config.get('THEME_REVEAL_CONFIG_TRANSITION', 'cube')
 
         # The pelican metadata format requires a markdown extension
-        if config.get('METADATA_FORMAT', 'nikola').lower() == 'pelican':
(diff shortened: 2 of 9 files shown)


felixfontein:
In the code, you sometimes use lower-case and sometimes title-case variants of the values for `METADATA_FORMAT`.

ralsina:
@felixfontein it all gets converted to lowercase.

nikola/nikola.py
```diff
@@ -960,6 +961,11 @@ def __init__(self, **config):
             self._GLOBAL_CONTEXT['subtheme'] = config.get('THEME_REVEAL_CONFIG_SUBTHEME', 'sky')
             self._GLOBAL_CONTEXT['transition'] = config.get('THEME_REVEAL_CONFIG_TRANSITION', 'cube')
 
+        # The pelican metadata format requires a markdown extension
+        if config.get('METADATA_FORMAT', 'nikola') == 'pelican':
```
felixfontein: Here it isn't converted to lowercase.

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -137,9 +137,13 @@ def create_post(self, path, **kw):
             content += '\n'
         with io.open(path, "w+", encoding="utf8") as fd:
             if onefile:
-                fd.write('<!-- \n')
-                fd.write(write_metadata(metadata))
-                fd.write('-->\n\n')
+                _format = self.site.config.get('METADATA_FORMAT', 'nikola')
```
felixfontein: Here neither.

nikola/plugins/compile/rest/__init__.py
```diff
@@ -177,7 +177,10 @@ def create_post(self, path, **kw):
             content += '\n'
         with io.open(path, "w+", encoding="utf8") as fd:
             if onefile:
-                fd.write(write_metadata(metadata))
+                _format = self.site.config.get('METADATA_FORMAT', 'nikola')
```
felixfontein: Here also not.

nikola/nikola.py
ralsina: Good catch! Fixed.

docs/manual.txt
```diff
@@ -421,6 +421,14 @@ other static site generators. The currently supported metadata formats are:
 
 You can add arbitrary meta fields in any format.
 
+When you create new posts, by default the metadata will be created as reST style comments.
+If you prefer a different format, you can set the ``METADATA_FORMAT`` to one of these values:
+
+* ``"Nikola"``: reSt comments wrapped in a comment if needed (default)
```
Kwpolska: reST* — here, below, in conf.py.in

nikola/nikola.py
```diff
@@ -960,6 +961,11 @@ def __init__(self, **config):
             self._GLOBAL_CONTEXT['subtheme'] = config.get('THEME_REVEAL_CONFIG_SUBTHEME', 'sky')
             self._GLOBAL_CONTEXT['transition'] = config.get('THEME_REVEAL_CONFIG_TRANSITION', 'cube')
 
+        # The pelican metadata format requires a markdown extension
+        if config.get('METADATA_FORMAT', 'nikola').lower() == 'pelican':
```
Kwpolska: No need to do fallback with `'nikola'` here

nikola/nikola.py
```diff
@@ -960,6 +961,11 @@ def __init__(self, **config):
...
+            if 'markdown.extensions.meta' not in config.get('MARKDOWN_EXTENSIONS', []):
```
Kwpolska: if markdown enabled: show warning, without `may`
else: ignore that

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -153,6 +157,12 @@ def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False,
         source = post.translated_source_path(lang)
         with io.open(source, 'r', encoding='utf-8') as inf:
             # Note: markdown meta returns lowercase keys
+            data = inf.read()
+            # If the metadata starts with "---" it's actually YAML and
+            # we should not let markdown parse it, because it will do
+            # bad things like setting empty tags to "''"
+            if data.splitlines()[0] == '---':
```
Kwpolska: What’s the behavior with `+++` ? also, you can just check `if data.startswith('---\n'):` (Windows-proof because of `r` mode)

nikola/plugins/compile/markdown/__init__.py
```diff
@@ -153,6 +157,12 @@ def read_metadata(self, post, file_metadata_regexp=None, unslugify_titles=False,
...
+                return {}
             _, meta = self.converter.convert(inf.read())
```
Kwpolska: pass `data` here!

docs/manual.txt
ralsina: Well, we also use it for a dozen other formats, and noone else uses it, so I say it's ours :-)

nikola/plugins/compile/markdown/__init__.py
ralsina: +++ means TOML (foo = bar) and that means it's not valid markdown metadata, so nothing bad happens. The problem was that YAML sometimes looks like valid markdown metadata :-P

docs/manual.txt
felixfontein: I think Chris means that you typed `reSt` instead of `reST` (note the last `T`), not that you should rename the format :)

nikola/nikola.py
ralsina: Done.

nikola/plugins/compile/markdown/__init__.py
ralsina: Doh

PR #2824, opened 2017-06-06
Add travis testing for OSX and fix locale tests.

gwax:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Fix locale tests on OSX and add OSX build to Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 before_install:
 - pip install --upgrade pip wheel
 language: python
-cache: apt
+cache:
+- apt
+- pip
 sudo: false
 addons:
   apt:
@@ -16,6 +18,15 @@ python:
 - '3.6'
 matrix:
   include:
+  - os: osx
+    language: generic
+    env: NMODE=nikola
+    python: '3.6'
+    before_install:
+    - brew upgrade pyenv
+    - eval "$(pyenv init -)"
+    - pyenv install 3.6.1
+    - pyenv global 3.6.1
   - python: '3.6'
     env: NMODE=flake8
 env:
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,11 @@ Features
   (Nikola/YAML/TOML/Pelican) (Part of Issue #2801)
 * Extract metadata from HTML meta and title tags like Pelican (Issue #1923)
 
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -16,6 +18,15 @@ python:
 - '3.6'
 matrix:
   include:
+  - os: osx
+    language: generic
+    env: NMODE=nikola
+    python: '3.6'
+    before_install:
+    - brew upgrade pyenv
```
Kwpolska: Why not `brew install python3; brew link python3`? I’m not a fan of pyenv myself, and the Homebrew python is pretty good. Plus, it doesn’t need to be updated to track Python patch versions (and minor versions, too)

CHANGES.txt
```diff
@@ -47,6 +47,7 @@ Bugfixes
 
 * Use inheritance tree when looking for theme engine (Issue #2816)
 * Fix unbound variable error in locale guessing (Issue #2806)
+* Tests run on OSX.
```
Kwpolska: This should go into “New in master”, I’ll fix it when merging.

(PS. macOS*)

gwax: Fixed.

.travis.yml
gwax: I considered that but it doesn't provide any control over which specific version of python is running and it means that changes in Homebrew could unexpectedly break builds.

The crux of it is that Travis support for Python builds on macOS is a little lacking and this was the simplest, reliable, reproducible way that I could come up with to get things to build (there are a lot of failed attempts squashed into this commit).

Kwpolska: You could argue the same about Travis CI’s images (in fact, I had to take down a C project of mine because Travis introduced random crashes I can’t debug in any way).

I’m going to merge this and experiment using Python from Homebrew. (and if it breaks, I’ll be the first to know as that’s how I’m working on my Mac).

gwax: As an aside, I was also not a fan of pyenv for quite a while but pyenv-virtualenv combined with `pyenv local` has really brought me around.

felixfontein:
There's one disadvantage for Travis macOS builds: they take forever to start. Take a look at the OSX Build Backlog here: https://www.traviscistatus.com/

PR #2825, opened 2017-06-06
Html metadata

ralsina:
Extract metadata from HTML tags.

Last part of #1923

### DIFF
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -418,6 +418,7 @@ other static site generators. The currently supported metadata formats are:
 * TOML, between ``+++`` (Hugo)
 * reST docinfo (Pelican)
 * Markdown metadata extension (Pelican)
+* HTML meta tags (Pelican)
 
 You can add arbitrary meta fields in any format.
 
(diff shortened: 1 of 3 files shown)


docs/manual.txt
```diff
@@ -529,6 +529,30 @@ the `markdown metadata extension docs <https://pythonhosted.org/Markdown/extensi
 
 Note that keys are converted to lowercase automatically.
 
+HTML metadata
```
Kwpolska: add it after line 420; I think *HTML meta tags* would be a more descriptive name.

docs/manual.txt
```diff
@@ -529,6 +529,30 @@ the `markdown metadata extension docs <https://pythonhosted.org/Markdown/extensi
...
+.. code:: html
+
+    <html>
+        <head>
+            <title>My super title</title>
+            <meta name="tags" content="thats, awesome" />
+            <meta name="date" content="2012-07-09 22:28" />
+            <meta name="modified" content="2012-07-10 20:14" />
+            <meta name="category" content="yeah" />
+            <meta name="authors" content="Alexis Métaireau, Conan Doyle" />
```
Kwpolska: I’m not 100% sure if the Pelican maintainer would like his name in our docs.

docs/manual.txt
ralsina: Ack, added.

PR #2827, opened 2017-06-06
Support blank lines in YAML/TOML metadata (part of issue #2801)

ralsina:
Split on the right markers for YAML/TOML metadata.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -338,7 +338,12 @@ class PageCompiler(BasePlugin):
         This splits in the first empty line that is NOT at the beginning
         of the document.
         """
-        split_result = re.split('(\n\n|\r\n\r\n)', data.lstrip(), maxsplit=1)
+        if data.startswith('---'):  # YAML metadata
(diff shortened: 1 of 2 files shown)


nikola/plugin_categories.py
```diff
@@ -338,7 +338,12 @@ def split_metadata(self, data):
         This splits in the first empty line that is NOT at the beginning
         of the document.
         """
-        split_result = re.split('(\n\n|\r\n\r\n)', data.lstrip(), maxsplit=1)
+        if data.startswith('---'):  # YAML metadata
+            split_result = re.split('(---\n\n|----\r\n\r\n)', data.lstrip(), maxsplit=1)
```
felixfontein: You have four dashes before the Windows newlines. I guess that's one too many.

felixfontein: Also: isn't it better to split along `\n\n---\n\n` (for Unix-style newlines) instead of `---\n\n`? Otherwise, a line ending with `---` will also suffice for a split (which is probably not what you want), as opposed to a line containing only `---`.

ralsina: Good points.

PR #2828, opened 2017-06-07
Use Python 3 from Homebrew for macOS Travis builds

Kwpolska:
cc @gwax, @felixfontein — I think this will be better, but feel free to disagree

I’ll merge this if nobody complains by Friday morning (UTC).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,14 @@ matrix:
   include:
   - os: osx
     language: generic
-    env: NMODE=nikola
+    env: NMODE=nikola OS=macOS
     python: '3.6'
     before_install:
-    - brew upgrade pyenv
-    - eval "$(pyenv init -)"
-    - pyenv install 3.6.1
-    - pyenv global 3.6.1
+    - brew install python3
+    - pip3 install virtualenv
+    - python3 -m virtualenv nikola-tests-venv
+    - source nikola-tests-venv/bin/activate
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,13 +20,14 @@ matrix:
   include:
   - os: osx
     language: generic
-    env: NMODE=nikola
+    env: NMODE=nikola OS=macOS
```
gwax: The apple vs penguin icons aren't enough?

.travis.yml
```diff
@@ -20,13 +20,14 @@ matrix:
...
     before_install:
-    - brew upgrade pyenv
-    - eval "$(pyenv init -)"
-    - pyenv install 3.6.1
-    - pyenv global 3.6.1
+    - brew install python3
+    - pip3 install virtualenv
+    - python3 -m virtualenv nikola-tests-venv
+    - source nikola-tests-venv/bin/activate
+    - pip install --upgrade pip
```
Kwpolska: Good point, fixed.

.travis.yml
Kwpolska: They’re hard to spot, especially if you’ve never looked there before (and the icons are barely-visible grey)

gwax: I asked the question expecting (and largely agreeing with) that answer.

Issue #2777, opened 2017-05-19
Use a lightbox that doesn't require JQuery


ralsina:
Now the base theme is using BaggetteBox.js

Still TBD:

- [x] Use captions
- [ ] Use it in bootstrap3
- [ ] Remove colorbox?

Kwpolska:
(If we can find a flowr alternative that does not use jQuery, that would be great)

ralsina:
PR #2929 does the rest of this issue.

ralsina:
This is done.

PR #2829, opened 2017-06-07
image captions (Part of #2777)

ralsina:
* Use alt attribute for lightbox captions
* Set alt attribute to image title in base theme galleries.


ralsina:
This seems to fail on windows because of some missing symlinks? Any ideas @Kwpolska ?

Kwpolska:
Fixed (I deleted some useless unminified files and hadn’t updated the symlinked list when you branched out)

PR #2832, opened 2017-06-11
Fix two grave and one interop bug in TOML/YAML meta handling

Kwpolska:
1. toml.load wants a file handle, toml.loads wants a string
2. `+` has special meaning in regular expressions
3. Require one linebreak after TOML/YAML (as Jekyll does)

Releasing v7.8.8 after this merges.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -336,7 +336,7 @@ class PageCompiler(BasePlugin):
         """Split data from metadata in the raw post content.
 
         This splits in the first empty line that is NOT at the beginning
-        of the document.
+        of the document, or after YAML/TOML metadata without an empty line.
         """
         if data.startswith('---'):  # YAML metadata
             split_result = re.split('(\n---\n|\r\n---\r\n)', data.lstrip(), maxsplit=1)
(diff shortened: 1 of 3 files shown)


nikola/plugin_categories.py
```diff
@@ -336,12 +336,12 @@ def split_metadata(self, data):
         """Split data from metadata in the raw post content.
 
         This splits in the first empty line that is NOT at the beginning
-        of the document.
+        of the document, or after YAML/TOML metadata without an empty line.
         """
         if data.startswith('---'):  # YAML metadata
-            split_result = re.split('(\n---\n\n|\r\n---\r\n\r\n)', data.lstrip(), maxsplit=1)
```
felixfontein: So we don't want to expect a double newline after the end of metadata?

felixfontein:
For anyone else interested, the toml library documentation: https://github.com/uiri/toml/blob/master/README.rst#api-reference

nikola/plugin_categories.py
Kwpolska: It’s not necessary if we have a `---` / `+++` separator already. This also mimics Jekyll behavior.

PR #2838, opened 2017-06-13
v8: cleanup and a few features

Kwpolska:
This is the groundwork for Nikola v8. Opinions, nitpicks, etc. always welcome.

**When this merges, the `master` branch is going to be v8, while `v7-maintenance` is where bugfixes for v7 can happen.**


nikola/__init__.py
```diff
@@ -26,11 +26,14 @@
 
 """Nikola -- a modular, fast, simple, static website generator."""
 
-from __future__ import absolute_import
 import os
+import sys
 
 __version__ = '7.8.8'
```
felixfontein: Shouldn't this be changed to '8.0.0'? Or will that happen later?

Kwpolska: We generally change version numbers right before release. I’ll make it `8.0.0.dev0`.

Kwpolska:
If no-one objects, I’m merging this on Friday.

nikola/plugin_categories.py
```diff
@@ -315,11 +314,6 @@ def compile_string(self, data, source_path=None, is_two_file=True, post=None, la
         # This function used to have some different APIs in different places.
         raise NotImplementedError()
 
-    # TODO remove in v8
-    def compile_html(self, source, dest, is_two_file=True):
```
felixfontein: The `self.compile_html(...)` call a few lines up should be replaced by `raise NotImplementedError()`.

nikola/plugins/compile/html.py
```diff
@@ -49,7 +48,7 @@ def compile_string(self, data, source_path=None, is_two_file=True, post=None, la
         if not is_two_file:
             _, data = self.split_metadata(data)
         new_data, shortcodes = sc.extract_shortcodes(data)
-        return self.site.apply_shortcodes_uuid(new_data, shortcodes, filename=source_path, with_dependencies=True, extra_context=dict(post=post))
+        return self.site.apply_shortcodes_uuid(new_data, shortcodes, filename=source_path, extra_context={'post': post})
```
felixfontein: Why are you removing `with_dependencies=True` here and at other places?

felixfontein: Ah, I didn't notice that you changed the API in `nikola.py` since apparently github was hiding its diff and so I couldn't find it. I now saw it when checking out the branch and looking in there...

felixfontein: BTW, the `compile_string` in `php.py` was bad; I added a commit to fix it.

PR #2846, opened 2017-06-16
Remove all leftover loghandlers support

Kwpolska:
Since we don’t support the entire log handlers thing, I dropped support for it, which also cleaned up all the imports and calls to `get_logger`.


Kwpolska:
Thanks!

PR #2848, opened 2017-06-19
Fix #2847 – non-fatal rev-parse failures in github_deploy

### LINKED ISSUE #2847: github_deploy shouldn’t fail if it can’t determine the source commit
[github_deploy](https://github.com/getnikola/nikola/blob/34e4767194c72c473ff4a31b9a8738f8cc6d662e/nikola/plugins/command/github_deploy.py#L128) tries to deploy from the branch specified by `GITHUB_SOURCE_BRANCH`, which [defaults](https://github.com/getnikola/nikola/blob/34e4767194c72c473ff4a31b9a8738f8cc6d662e/nikola/nikola.py#L620) to `master`. 

This doesn't make sense to me. The default behavior that I would expect would be to just deploy the current contents of the repo.

Kwpolska:
This is #2847. cc @asmeurer

*(note to future self: this goes to v7-maintenance as well)*

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Features
 Bugfixes
 --------
 
+* Make failures to get source commit hash non-fatal in
+  ``github_deploy`` (Issue #2847)
 * Less cryptic error when guessing format from extension in ``new_post``
   fails
 * Use Jupyter name more consistently in docs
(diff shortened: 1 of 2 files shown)


asmeurer:
Does it not try to deploy from `master` even if `master` exists when `GITHUB_COMMIT_SOURCE = False`?

Kwpolska:
It never did that. It only took the commit hash from there.

asmeurer:
That wasn't what I saw when I was testing this on Travis. From my PR build it was apparently trying to build from master (the build failed because of a change that I had only made on my branch).

Kwpolska:
Merged, it shouldn’t break much.

PR #2854, opened 2017-06-26
Killed sllides

ralsina:
Will add them as a plugin in the plugins repo in a separate PR.


docs/manual.txt
```diff
@@ -80,7 +80,7 @@ Nikola can't do:
 * An Issue tracker
 * Anything with forms, really (except for `comments`_!)
 
-Keep in mind that "static" doesn't mean **boring**. You can have animations, slides
+Keep in mind that "static" doesn't mean **boring**. You can have animations
```
Kwpolska: I don’t think anyone would have parsed this as *the slides directive*.

felixfontein:
I think you forgot to remove a slides specific test :)

Issue #2830, opened 2017-06-08
[RFC] Metadata extraction plugins


Kwpolska:
`get_meta` is complicated and hard to extend (eg. plugins.getnikola.com needs a special compiler to read meta). So, perhaps we could split this out into special plugins?

Some preliminary ideas:

* plugins can specify supported file formats (extensions)/compilers
* plugins can specify whether they should be used as fallback or not (eg. regex title/filename hacks)
* plugins can specify other characteristics (eg. first line of file)
* built-in stuff *should not* be real yapsy plugins for performance reasons (that would require reading way too many files)

felixfontein:
Me too. I guess it is also related to #1278.

Issue #2380, opened 2016-06-20
new_post -2 -i could (should?) copy a file instead of read it.


tritium21:
I write my posts in MS Word, and have recently started trying to use the .docx files directly with Nikola.  Nikola supports compiling these documents out of the box, which is nice.  However, I would like to streamline the process further by using Nikola's built in import file functionality.

Currently, it attempts to read the file contents and pass it to the compiler's create post function.  While this works for most file types, it does not work for docx's binary file type (zipped xml) as it attempts to read the file as utf-8.

Because file types like docx have to use two file mode anyways, it would be convenient to copy the file instead of trying to read it when -2 is passed to new_post.

ralsina:
Sounds good.

Kwpolska:
:+1: from me. I’ll work on this later.

tritium21:
The PR works great on my end.

PR #2856, opened 2017-07-01
Generalize metadata functions and moving them to utils.py

felixfontein:
This makes implementing getnikola/plugins#232 much easier, and separates the metadata extraction from the higher level parts (metadata transformation).

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -28,13 +28,12 @@
 
 import sys
 import os
-import re
 import io
 
 from yapsy.IPlugin import IPlugin
 from doit.cmd_base import Command as DoitCommand
 
-from .utils import LOGGER, first_line
+from .utils import LOGGER, first_line, split_metadata
 
 __all__ = (
     'Command',
@@ -329,16 +328,8 @@ class PageCompiler(BasePlugin):
         This splits in the first empty line that is NOT at the beginning
         of the document, or after YAML/TOML metadata without an empty line.
         """
-        if data.startswith('---'):  # YAML metadata
-            split_result = re.split('(\n---\n|\r\n---\r\n)', data.lstrip(), maxsplit=1)
-        elif data.startswith('+++'):  # TOML metadata
-            split_result = re.split('(\n\\+\\+\\+\n|\r\n\\+\\+\\+\r\n)', data.lstrip(), maxsplit=1)
-        else:
-            split_result = re.split('(\n\n|\r\n\r\n)', data.lstrip(), maxsplit=1)
-        if len(split_result) == 1:
-            return '', split_result[0]
-        # ['metadata', '\n\n', 'post content']
-        return split_result[0], split_result[-1]
+        meta, content, _ = split_metadata(data)
+        return meta, content
 
     def get_compiler_extensions(self):
         """Activate all the compiler extension plugins for a given compiler and return them."""
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -937,19 +937,8 @@ class Post(object):
 # Code that fetches metadata from different places
 
 
-def re_meta(line, match=None):
-    """Find metadata using regular expressions."""
-    if match:
-        reStr = re.compile('^\.\. {0}: (.*)'.format(re.escape(match)))
-    else:
-        reStr = re.compile('^\.\. (.*?): (.*)')
-    result = reStr.findall(line.strip())
-    if match and result:
-        return (match, result[0])
-    elif not match and result:
-        return (result[0][0], result[0][1].strip())
-    else:
-        return (None,)
+# For backwards compatibility
+re_meta = utils.re_meta
 
 
 def _get_metadata_from_filename_by_regex(filename, metadata_regexp, unslugify_titles, lang):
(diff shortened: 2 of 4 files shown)


nikola/utils.py
```diff
@@ -2028,3 +2028,106 @@ def read_from_config(self, site, basename, posts_per_classification_per_language
...
+        reStr = re.compile('^\.\. {0}: (.*)'.format(re.escape(match)))
+    else:
+        reStr = re.compile('^\.\. (.*?): (.*)')
+    result = reStr.findall(line.strip())
+    if match and result:
+        return (match, result[0])
+    elif not match and result:
+        return (result[0][0], result[0][1].strip())
+    else:
+        return (None,)
```
Kwpolska: `return None, None` — and no parentheses above

nikola/utils.py
```diff
@@ -2028,3 +2028,111 @@ def read_from_config(self, site, basename, posts_per_classification_per_language
...
+    format.
+
+    Valid values for ``metadata_type`` are:
+    * ``'none'``: no metadata was found (file was empty)
+    * ``'yaml'``: metadata in YAML format
+    * ``'toml'``: metadata in TOML format
+    * ``'rest'``: metadata in reST format (the standard Nikola
+                  reST-like metadata format)
+    """
+    if data.startswith('---'):  # YAML metadata
```
Kwpolska: There should be something to split metadata based on the post object as well, and the post object should know what metadata type it was.

felixfontein:
I think you mean #2830, not #2380.

I actually forgot about #2380; I just wanted to convert the existing functionality to something which can be easier reused.

nikola/utils.py
felixfontein: Fine for me. I just copy'n'pasted this code from `post.py` and didn't modify anything in this function.

nikola/utils.py
felixfontein: That would be for #2830, right?

Kwpolska: Yes, I’ll handle it then.

Kwpolska:
Yeah, that was a typo: #2830. I’ll go :+1: for merging it, but don’t expect this API to be stable.

nikola/plugin_categories.py
```diff
@@ -329,16 +328,8 @@ def split_metadata(self, data):
...
-            split_result = re.split('(\n---\n|\r\n---\r\n)', data.lstrip(), maxsplit=1)
-        elif data.startswith('+++'):  # TOML metadata
-            split_result = re.split('(\n\\+\\+\\+\n|\r\n\\+\\+\\+\r\n)', data.lstrip(), maxsplit=1)
-        else:
-            split_result = re.split('(\n\n|\r\n\r\n)', data.lstrip(), maxsplit=1)
-        if len(split_result) == 1:
-            return '', split_result[0]
-        # ['metadata', '\n\n', 'post content']
-        return split_result[0], split_result[-1]
+        meta, content, _ = split_metadata(data)
```
felixfontein: I don't understand this one. Is codacy maybe confusing the imported `utils.split_metadata` with `PageCompiler.split_metadata`?

Kwpolska: It seems to be. Ignore that warning.

felixfontein:
Sure. I assume nothing in `master` is stable until the first v8 version (or release candidate) is released :)

I'm not sure whether I'll use this API though, now that you mentioned #2830...

nikola/post.py
```diff
@@ -999,52 +988,15 @@ def get_metadata_from_file(source_path, config=None, lang=None):
...
-            utils.req_missing('pyyaml', 'use YAML metadata', optional=True)
-            raise ValueError('Error parsing metadata')
-        idx = meta_data.index('---', 1)
-        meta = yaml.safe_load('\n'.join(meta_data[1:idx]))
-        # We expect empty metadata to be '', not None
-        for k in meta:
-            if meta[k] is None:
-                meta[k] = ''
+    meta, metadata_type = utils.extract_metadata(file_lines)
+    if metadata_type == 'yaml':
```
Kwpolska: For both blocks:

```py
      if metadata_type in ('yaml', 'toml'):
          # Map metadata from other platforms to names Nikola expects (Issue #2817)
          map_metadata(meta, metadata_type, config)
```

felixfontein: Fixed.

nikola/utils.py
```diff
@@ -2028,3 +2028,111 @@ def read_from_config(self, site, basename, posts_per_classification_per_language
         args = {'translation_manager': self, 'site': site,
                 'posts_per_classification_per_language': posts_per_classification_per_language}
         signal('{}_translations_config'.format(basename.lower())).send(args)
+
+
+# Moved to global variable to avoid recompilation
+# of regex every time re_meta() is called.
```
Kwpolska: Regex fun fact: there’s a cache, for 512 entries if memory serves. It probably doesn’t get recompiled. (Don’t change this back)

felixfontein: Ah, I didn't knew about that cache. I don't like the idea of depending on it in such cases where it is obvious that the regex will be used over and over again, though, so yes, I won't change this back :)

felixfontein:
Thanks for reviewing!

PR #2858, opened 2017-07-03
Add NIKOLA_SHOW_TRACEBACKS environment variable

Kwpolska:
Not showing tracebacks to end-users is generally fine, but as a developer, it’s pretty annoying. On the other hand, ``NIKOLA_DEBUG=1`` floods the console with useless Yapsy debug messages.

So, the fix: add ``NIKOLA_SHOW_TRACEBACKS`` that shows tracebacks but no debug stuff (as a middle ground).

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,8 @@ New in master
 Features
 --------
 
+* Add ``NIKOLA_SHOW_TRACEBACKS`` environment variable that shows
+  full tracebacks instead of one-line summaries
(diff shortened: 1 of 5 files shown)


Kwpolska:
(FYI, I added it to my `.zshrc` to make stuff easier.)

tritium21:
If this wasn't already merged, I would have suggested making NIKOLA_DEBUG a non-boolean option.  Or, at least except other values for levels of verbosity (with '1', 'true', 'yes' defaulting current behavior.)

PR #2861, opened 2017-07-05
Extendable metadata extractors (#2830)

Kwpolska:
**This is #2830 — extendable metadata extractors.**

Every extractor plugin specifies the source of information (currently supported are text and filename) and its priority (based on how well it knows it’s needed — YAML/TOML use markers, so they’re `specialized`; Nikola is `normal`, and FileMetadataRegexp is `fallback`). This tries to be extendable, although I probably missed a few things regarding openness.

**Highlights:**

 * YAML and TOML support in 2-file posts
 * FILE_METADATA_REGEXP still supported
 * Nikola-site and a lot of test stuff works properly, without changes
 * Posts know what metadata extractor was used, if they are 1-file posts
 * Compilers must advertise support for metadata, can use conditions, and have the highest priority
 * Metadata is split for parsing (and then again when compiling)
 * I’m doing modern Python 3: using enums and (basic) type annotations. This adds `enum34` for Python 3.3 (do we need that version though?)
 * I had to effectively undo #2856 as part of the implementation, since the new extractors have different requirements
 * Some compiler APIs were changed as well, so old compilers might not work
 * If someone needs direct access to the Nikola extractor, it’s metadata_extractors.DEFAULT_EXTRACTOR — but most of the time, you should go through `metadata_extractors_by` or other convenience methods.
 * As I promised, the built-in extractors are not loaded by yapsy. This sort-of optimization could also be useful for a few other mandatory plugins, such as PostScanner or render_posts/render_pages

**Questions:**

 * What was `match` used for in `re_meta`? I didn’t see any uses, so I nuked it.
 * I removed support for `\r\n` when splitting metadata. The reason? This happens *only* when dealing with Windows-style newlines on Linux. (unless I’m wrong. We can restore it just in case.)
 * Right now, metadata_extractors_by is on the site object, but I’m not sure about it — perhaps it would work better in metadata_extractors.py as a global?
 * I kinda want to change `create_post` compiler API to `def create_post(self, metadata, options):` — opinions?

**Future and caveats:**

 * Compiler plugins may need changes for compatibility, I didn’t do that.
* `pkgindex` needs a rewrite. I didn’t do that either.
 * I plan to extend annotations over to all plugin categories. PyCharm is oh-so-nice when you add type annotations.
 * We still parse reST/Markdown posts twice if using their respective meta formats. Memory usage would skyrocket if we tried to avoid that.

Comments and reviews welcome.


nikola/nikola.py
```diff
@@ -728,6 +732,11 @@ def __init__(self, **config):
         if self.config['PRESERVE_EXIF_DATA'] and not self.config['EXIF_WHITELIST']:
             utils.LOGGER.warn('You are setting PRESERVE_EXIF_DATA and not EXIF_WHITELIST so EXIF data is not really kept.')
 
+        # TODO: should we keep backwards compat here?
```
felixfontein: It's nice to have a warning for some removed/renamed features for at least one or two releases. If someone is still using this feature with an up-to-date Nikola, he at least will notice that something changed :)

nikola/plugin_categories.py
```diff
@@ -372,6 +379,71 @@ class MarkdownExtension(CompilerExtension):
...
+    supports_write = False
+
+    def _extract_metadata_from_text(self, source_text: str) -> dict:
+        """Extract metadata from text."""
+        raise NotImplementedError()
+
+    def split_metadata_from_text(self, source_text: str) -> (str, str):
+        """Split text into metadata and content (both strings)."""
+        if self.split_metadata_re is None:
+            return source_text
```
felixfontein: Shouldn't this be `return '', source_text`?

nikola/plugin_categories.py
```diff
@@ -372,6 +379,71 @@ class MarkdownExtension(CompilerExtension):
...
+        else:
+            split_result = self.split_metadata_re.split(source_text.lstrip(), maxsplit=1)
+            if len(split_result) == 1:
+                return split_result[0], split_result[0]
```
felixfontein: Shouldn't this be `return '', split_result[0]`?

nikola/plugin_categories.py
```diff
@@ -372,6 +379,71 @@ class MarkdownExtension(CompilerExtension):
...
+            else:
+                # Necessary?
+                return split_result[0], split_result[-1]
+
+    def extract_text(self, source_text: str) -> dict:
+        """Split file, return metadata and the content."""
```
felixfontein: The function is doing something else. It returns the parsed metadata, but *not* the content.

nikola/plugins/basic_import.py
```diff
@@ -166,7 +166,7 @@ def write_metadata(filename, title, slug, post_date, description, tags, **kwargs
         with io.open(filename, "w+", encoding="utf8") as fd:
             data = {'title': title, 'slug': slug, 'date': post_date, 'tags': ','.join(tags), 'description': description}
             data.update(kwargs)
-            fd.write(utils.write_metadata(data))
+            fd.write(utils.write_metadata(data, metadata_format='nikola', comment_wrap=False))
```
felixfontein: Even though it's technically not needed (because `nikola` is always provided by Nikola core), why not pass `site` as well?

nikola/post.py
```diff
@@ -934,129 +942,103 @@ def source_ext(self, prefix=False):
...
+    """Get post meta from compiler or source file."""
     meta = defaultdict(lambda: '')
+    used_extractor = None
 
-    try:
-        config = post.config
-    except AttributeError:
-        config = None
+    config = getattr(post, 'config', None)
+    metadata_extractors_by = getattr(post, 'metadata_extractors_by', metadata_extractors.default_metadata_extractors_by())
```
felixfontein: This always calls `metadata_extractors.default_metadata_extractors_by()`, even though it should never be used. Wouldn't it be better to add two more lines (`if metadata_extractors_by is None:` / `metadata_extractors_by = metadata_extractors.default_metadata_extractors_by()`) to avoid this call every time?

nikola/utils.py
```diff
@@ -1437,23 +1436,44 @@ def get_translation_candidate(config, path, lang):
...
 
-    if _format == 'yaml':
-        if yaml is None:
-            req_missing('pyyaml', 'use YAML metadata', optional=False)
-        return '\n'.join(('---', yaml.safe_dump(data, default_flow_style=False).strip(), '---', '', ''))
+    Recommended usage: pass `site`, `comment_wrap`, and optionally `compiler`. Other options are for backwards compatibility.
+    """
+    # API compatibility
+    if metadata_format is None:
+        metadata_format = site.config.get('METADATA_FORMAT', 'nikola').lower()
```
felixfontein: In case `site` is done, this dies.

nikola/utils.py
```diff
@@ -1437,23 +1436,44 @@ def get_translation_candidate(config, path, lang):
...
+        else:
+            # Quiet fallback.
+            metadata_format = 'nikola'
+
+    extractor = metadata_extractors_by['name'].get(metadata_format)
+    if extractor and extractor.supports_write:
+        extractor.check_requirements()
+        return extractor.write_metadata(data, comment_wrap)
+    elif metadata_format is None:
+        pass  # Quiet fallback to Nikola
```
felixfontein: This won't be quiet, because the new `if` below will print out a warning.

felixfontein:
It's nice to see that Python also got type annotations. Also, I'm for enums; whether it will be `enum34` or Python 3.4+ only, I don't mind.

tritium21:
What platforms ship 3.3 as the default python3?  Are those platforms still supported?  I'm pretty sure ubuntu 12.04 was 3.2, but 14.04 was 3.4, but i could be wrong.

felixfontein:
Debian 8 (jessie == oldstable) ships with Python 3.4, and Debian 9 probably with something newer.

Python 3.4 was released in March 2014, so Ubuntu 14.04 supports either 3.3 or 3.4, with 3.3 being more likely. I have no Ubuntu 14.04 boxes left to check, though...

Kwpolska:
You don’t need boxen to check: [14.04 trusty ships 3.4.0](https://packages.ubuntu.com/trusty/python3), [9 stretch ships 3.5.3](https://packages.debian.org/stretch/python3)

nikola/metadata_extractors.py
```diff
@@ -0,0 +1,254 @@
...
+
+    text = 1
+    filename = 2
+
+
+def check_conditions(post, filename: str, conditions: list, config: dict, source_text: str) -> bool:
+    """Check the conditions for a metadata extractor."""
+    for ct, arg in conditions:
+        if any((
+            ct == MetaCondition.config_bool and (arg not in config or (arg in config and not config[arg])),
```
Kwpolska: Brain fart. Fixed.

nikola/metadata_extractors.py
```diff
@@ -0,0 +1,254 @@
...
+@_register_default
+class NikolaMetadata(MetadataExtractor):
+    """Extractor for Nikola-style metadata."""
+
+    name = 'nikola'
+    source = MetaSource.text
+    priority = MetaPriority.normal
+    supports_write = True
+    split_metadata_re = re.compile('\n\n')
+    nikola_re = re.compile('^\s*\.\. (.*?): (.*)')
```
Kwpolska: In Nikola, you’re right. In TOML, this is intended:

```py
split_metadata_re = re.compile('\n\\+\\+\\+\n')
```
We need `\n` for a real newline. So the double backslash becomes a single backslash, and it ends up escaping the `+` from regexp’s point of view.

nikola/plugin_categories.py
Kwpolska: Not quite, see below.

nikola/plugin_categories.py
Kwpolska: This is actually intended, and handles two special cases: 2-file posts, and posts which do not have any `\n\n` in them (which is legal, although rare). Since `extract_text` splits and then passes *only* metadata to the parser, we need to return the same thing as both text and metadata.

nikola/plugin_categories.py
Kwpolska: Leftover from a previous design. Fixed.

nikola/plugin_categories.py
```diff
@@ -372,6 +379,71 @@ class MarkdownExtension(CompilerExtension):
...
+        split = self.split_metadata_from_text(source_text)
+        meta = self._extract_metadata_from_text(split[0])
+        return meta
+
+    def extract_filename(self, filename: str, lang: str) -> dict:
+        """Extract metadata from filename."""
+        return {}
+
+    def write_metadata(self, metadata: dict, comment_wrap=False) -> str:
```
Kwpolska: Better solution, change the behavior of Nikola’s extractor. A copy of a small dict shouldn’t hurt.

nikola/plugins/basic_import.py
Kwpolska: There is no site in the importer plugins.

nikola/post.py
Kwpolska: Fixed.

nikola/utils.py
Kwpolska: Fixed.

nikola/utils.py
Kwpolska: I made sure `metadata_format` will never be None.

nikola/plugins/basic_import.py
felixfontein: But you still have a `site` object (with all default options)?

nikola/plugin_categories.py
felixfontein: That's even better :)

nikola/metadata_extractors.py
felixfontein: I know; I meant that `'\+' == '\\+'`, which is true for the same reason than `'\s' == '\\s'` (because `\s` is no valid escape sequence).

Kwpolska: [That’s deprecated behavior.](https://bugs.python.org/issue27364)

nikola/plugins/basic_import.py
Kwpolska: Do we really? I can see references to that in the WordPress importer, but I have no idea where it comes from…

nikola/plugin_categories.py
felixfontein: I still don't understand this. I would *never* expect a function called `split_metadata_from_text` with docstring `"Split text into metadata and content (both strings)."` to do something like this. I also don't see why `extract_text` would require this behavior.

nikola/plugins/basic_import.py
felixfontein: Every plugin has a `set_site` call which is called on plugin initialization. There is no documentation saying that this isn't the case for `Command` plugins, and in particular importer plugins. So I would assume it is a bug if the `site` object isn't there.

Kwpolska: Ah, right. I misremembered `needs_config` — turns out a site always exists, but isn’t always required to be configured. I’ll fix that.

nikola/metadata_extractors.py
felixfontein: Makes sense; I was surprised that it works. I was just confused because here you seemed to use the (deprecated) behavior, while further down you were escaping, and I didn't know if that was intentional or not. But now that's resolved :)

nikola/plugin_categories.py
Kwpolska: You can’t know what is metadata and what is text if splitting fails.

The old implementation did it differently: it always split text into lines, and kept looking for data until it found an empty line. YAML/TOML was hacked onto it — if it found the first line, it would have to re-join text after finding the second delimiter. There were two different splitting algorithms, one based on `str.split('\n')` and the other on regexes.

This implementation unifies splitting with the regex-based solution, and `str.split('\n')` happens only in the Nikola extractor. It also allows YAML/TOML in 2-file posts. But there’s a side effect of that: since we always call `split_metadata_from_text`, even for 2-file posts, it needs to return the same string twice if it doesn’t find the delimiter.

I consider this to be good-enough, because we always use only one of this function’s return values.

felixfontein: Ok. But:
* Why not document the behavior in the docstring?
* Why not add an additional argumen to `extract_text` which says that this is a metadata-only file (in which case it would not call `split_metadata_from_text` first), and then set this argument to `True` when parsing metadata files?

Kwpolska: 1. Done.
2. This complicates stuff. Currently, the YAML/TOML extractors *depend* on a split happening in this case, since they don’t even try looking for the trailing `---`/`+++`.

felixfontein: Ok. I can live with that :)

Kwpolska:
@felixfontein: thanks!

@ralsina: It’d be nice if you reviewed, too.

ralsina:
@Kwpolska I'll give it a look this weekend.

PR #2872, opened 2017-07-17
Fixed cleanup of 'terminado' and 'tornado'

uli-heller:
After creating the snap, these are directories containing files. So you
cannot delete them via 'rm'

### Pull Request Checklist

- [x] I tested my changes.

### Description

Without this change, you'll get an error message towards the end of the
creation of the snap. Unfortunately, I don't have a copy of it available at the moment.
I can recreate it when required.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -53,5 +53,5 @@ parts:
         build-packages: [libjpeg-dev]
         install: |
             find $SNAPCRAFT_PART_INSTALL -name '*.a' -exec rm {} \;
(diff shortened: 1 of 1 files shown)


Kwpolska:
This *is* a minor change, not worthy of a changelog entry. Most users don’t care about building snaps.

uli-heller:
OK, thanks. So I'll leave the changelog unmodified for future 
modifications of the snapcraft.yaml file.

uli-heller:
I've updated the PR, no more changes to CHANGES.txt

Kwpolska:
Using `xargs` like that can be unsafe (if you have spaces in directory names), I fixed it in 37eb408a.

PR #2873, opened 2017-07-17
snap: Use the current checkout as source of the snap

uli-heller:
### Pull Request Checklist

- [x] I tested my changes.

### Description

With this PR, the currently active checkout is used when creating the snap package
instead of the head of the github master.

I'm not sure if this PR is actually a good idea. I just needed a similar approach to build a
snap package of nikola-7.8.9.


ralsina:
Yes, it's perfectly fine.

PR #2878, opened 2017-07-24
Allow post and config arguments of metadata_extractors.check_conditions to be None

felixfontein:
Makes the function more flexible. For getnikola/plugins#241 (at least `post` being `None`).


felixfontein:
Hmm, maybe allowing `config` to be `None` is a bit too much. What do you think?

Kwpolska:
Yeah, it’s too much.

felixfontein:
I changed it. I decided to force-push this so that this minimal change won't have two commits.

felixfontein:
Thanks for idea, review and merging! :)

PR #2879, opened 2017-07-24
Absolute paths in path handlers

### LINKED ISSUE #2876: Plugin with a path handler generating external links
### Description:
I wanted to write a plugin allowing me register a path handler which would give me a possibility to share a bibliography of useful externals links between posts.

For example:
```rst
__ link://biblio/linkname ---> https://some.random.site.com/path/to/resource
```
Turns out, judging by the Nikola code, path handlers are only able to generate internal links, which makes the whole link handler mechanism useless for the case.

Is there a way I can help with developing the feature? Or is there an alternative approach to the problem?

vkazanov:
The branch does the following changes to Nikola.path:

1. Makes try..except more granular, i.e. the exception only wraps the code that is expected to throw KeyError here.

2. If path handler returns a string the it is considered to be a complete absolute URL not requiring any processing.

Fixes #2876

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1750,23 +1750,23 @@ class Nikola(object):
 
         try:
             path = self.path_handlers[kind](name, lang, **kwargs)
-            if path is None:
-                path = "#"
-            else:
-                path = [os.path.normpath(p) for p in path if p != '.']  # Fix Issue #1028
-            if is_link:
-                link = '/' + ('/'.join(path))
-                index_len = len(self.config['INDEX_FILE'])
-                if self.config['STRIP_INDEXES'] and \
-                        link[-(1 + index_len):] == '/' + self.config['INDEX_FILE']:
-                    return link[:-index_len]
-                else:
-                    return link
-            else:
-                return os.path.join(*path)
         except KeyError:
             utils.LOGGER.warn("Unknown path request of kind: {0}".format(kind))
             return ""
+        if path is None:
+            path = "#"
+        else:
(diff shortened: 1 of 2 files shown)


nikola/nikola.py
```diff
@@ -1736,38 +1736,47 @@ def path(self, kind, name, lang=None, is_link=False, **kwargs):
         * slug (name is the slug of a post or page)
         * filename (name is the source filename of a post/page, in DEFAULT_LANG, relative to conf.py)
 
-        The returned value is always a path relative to output, like
-        "categories/whatever.html"
+        The returned value is either a path relative to output, like "categories/whatever.html", or
+        an absolute URL ("https://en.wikipedia.org/wiki/Nikola"), if path handler returns a string.
```
Kwpolska: (A better sample URL would be `http://getnikola.com/`)

Kwpolska: (A better sample URL would be `https://getnikola.com/`)

vkazanov: True, will fix in a sec

Kwpolska:
Thanks for contributing to Nikola! :tada:

vkazanov:
Great! Thank you for handling the issue in the fastest way possible. 

BTW, what are the plans for the 8.0.0 milestone? Currently I use a patched 7.8.9, which means I can't publish the plugin (external link bibliography) officially.

Kwpolska:
There is currently no planned release date; however it will take a few more months to complete our plans. You can release your plugin to the repo as v8-only, telling interested users that they need either a manual patch, or the development version from GitHub.

vkazanov:
Ok, thanks for clarifying.

Issue #2868, opened 2017-07-17
shortcode for equivalent of ReST listing directive


RobinD42:
Feature request:  It would be very nice to be able to have a shortcode that would enable the equivalent of the `.. listing::` ReST directive for the other compilers. Most of my existing docs are in Markdown and converting one to ReST in order to add a listing seems like it shouldn't be necessary.

ralsina:
https://github.com/getnikola/nikola/pull/2880/

Kwpolska:
Fixed in #2880.

PR #2880, opened 2017-07-24
Listing shortcode

ralsina:
Implement #2868


nikola/plugins/shortcode/listing.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+# This file is public domain according to its author, Brian Hsu
```
Kwpolska: ???

ralsina: Whoa, flashback. I don't think any of that code is still there. OTOH, I suspect this is because I copied gist.py :-)

Kwpolska:
Could we avoid duplicating rest/listings.py code?

ralsina:
@Kwpolska there is not that much repeated code:

* The code to find the original file
* The code to generate the link to the sources

It's about 10 LOC and only used in these two, so can't go to utils/ ... any suggestions where I could put it?

PR #2882, opened 2017-07-30
Fix #2850 — rewrite `nikola auto` with asyncio and aiohttp

### LINKED ISSUE #2850: Rewrite `nikola auto` using asyncio and aiohttp
Will make it much more reliable. We might also drop `nikola serve` or merge it with `nikola auto`.

asyncio supports 3.3+ ([3.3 backport](https://pypi.python.org/pypi/asyncio)/3.4 builtin). Some syntax niceties are 3.5+, but those are optional.

Resources: ~~[autobahn on GitHub](https://github.com/crossbario/autobahn-python) — [asyncio 3.6 docs](https://docs.python.org/3/library/asyncio.html) — [soundtrack](https://www.youtube.com/watch?v=e11h73WhqK4)~~ switched to `aiohttp`

Kwpolska:
This is #2850, a more modern (and less hacky) version of `nikola auto`.

~~Work in progress. To do:~~

- [x] add a rebuild queue (currently, rebuilds may happen concurrently)
- [x] restore `no-server` option (do we need it?)
- [x] figure out what the `liveCSS` option means in LiveReload *(it reloads CSS without reloading the entire page, let’s leave it enabled)*
- [x] code style
- [x] test on Linux and Windows
- [x] test on Python ~~3.3~~ (**dropped support**), 3.4, 3.5, 3.6
- [x] stop depending on virtualenvs being active/`nikola` in `$PATH`

I don’t think we can merge it with `nikola serve`, because it requires extra dependencies.

---

Tested on:

* Ubuntu 14.04, Python 3.4, using a remote server (it works!)
* macOS, Python 3.5 (thanks @offby1)
* macOS Sierra, Python 3.6
* Arch Linux, Python 3.6
* Windows 7, Python 3.6
* Kubuntu 16.04, Python 3.5 (thanks @altendky)

---

If anyone wants to help with testing: clone and install with extra requirements; create a demo site (`nikola init -qd n`), run `nikola auto -b` in the site, verify everything works (index; `/pages/` path; any subpage), try editing a post and see if it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,6 @@ env:
   global:
     secure: GqUDW0j4Ay8pS9tS/g5JBYGtgBG0g0oD2g4X9JXgVchiFj3GIRL+dwiXpgx3LELy7nCWF3EzClFxxMdVqqxizQ/I8xqiA0XL/rf5z/y+9d9nmKOxV8SV73n0eCgsHWkbdSLJ6MKs6s0Trlz9jLu/P+Bhfu2ttqkOH8WttfEf3VE=
 install:
-- if [[ $NMODE == 'nikola' ]]; then scripts/getwheelhouse.sh $(scripts/getpyver.py short); fi
 - if [[ $NMODE == 'nikola' ]]; then pip install -r requirements-tests.txt; if [[ "$?" == '1' ]]; then cat /home/travis/.pip/pip.log; false; fi; fi
 - if [[ $NMODE == 'nikola' ]]; then pip install .; fi
 - if [[ $NMODE == 'flake8' ]]; then pip install flake8 pydocstyle; fi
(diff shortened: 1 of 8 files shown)


Kwpolska:
@ralsina, @felixfontein, this is finished — reviews welcome.

felixfontein:
I don't use `nikola auto` so I cannot comment much on this, but I went through the change and it looks good to me.

Just a minor thing: in case you change `conf.py` (like changing the name of the posts folders), `nikola auto` will not start watching the new folders, right? I guess implementing that would require a lot of extra work (and the previous version didn't do that either) and is a very infrequent usecase, but I thought I'd still ask :)

Kwpolska:
@felixfontein Yes, that’s a edge case that would be pretty hard to support.

Issue #2809, opened 2017-06-01
Make markdown a 1st class citizen


ralsina:
More and more people use markdown, while reStructuredText is stagnant.

As much as I prefer reSt, it's clear where this is headed. So, we need to support markdown as much as possible.

- [x] Enable markdown posts/pages by default
- [x] Don't make it optional
- [x] Ensure all important features work with markdown as well as reSt (good thing we have shortcodes)
- [ ] Make gallery directive into a shortcode (user request!)
- [x] Ensure no functionality is lost if reSt is disabled (right now a few things like charts go away?)

mgaitan:
Roberto, what about to use this ? https://github.com/rtfd/recommonmark

ralsina:
Sure, why not. Docs make it look sorta sphinx-specific, tho.

stefanos82:
@ralsina so if I understood you correctly, does it mean that existing reStructuredText projects will have to migrate to Markdown format or will there be an automating process with future versions of Nikola?

To be honest with you, I chose Nikola **exactly for this particular reason**, to use reStructuredText that is.

This is a very sad day for me...

ralsina:
@stefanos82 not at all. Just that markdown will be as functional as reSt is now. Nothing changes for reSt users.

stefanos82:
Awesome! Now I feel relieved :+1:

Hawker65:
Are titles using the following format supported?

    # H1
    ## H2
    ### H3
    #### H4
    ##### H5
    ###### H6

I can only get to work this kind of title :

    Alt-H1
    ======
    
    Alt-H2
    ------

Kwpolska:
@Hawker65 Works for me. Are you sure you’re working with Markdown?

(Also, in the future, please create new issues instead of responding to old, unrelated ones.)

Hawker65:
@Kwpolska My bad, wrong file type. For usage questions, should I use https://groups.google.com/forum/#!forum/nikola-discuss?

Kwpolska:
@Hawker65 Yes, we encourage using the mailing list. Creating issues is also fine if you’re dealing with something that looks like a bug.

Hawker65:
Understood, sorry for the inconvenience.

Kwpolska:
I just added a `gallery_shortcode` plugin. This is done.

PR #2904, opened 2017-09-19
Thumbnail shortcode, take 2

Kwpolska:
Adds a less basic thumbnail shortcode with ``alt`` and ``align`` options, as well as caption support.

Supersedes / Closes #2903. Via Issue #2809.


Kwpolska:
Thanks for reviewing this!

PR #2926, opened 2017-11-29
Update PT and PT_BR translation

FabioRosado:
### Pull Request Checklist

- [ ] I tested my changes.

### Description
I've noticed that both the Portuguese and Portuguese from Brazil translations were missing some of the fields so I've added them.  I haven't added my name to the translators header because I was unsure if this is the right thing to do since its such a quick modification.

In the Portuguese translation (pt.po) I've also changed some of the translated words:
- line 34: Replace tags to its plural form (Etiquetas)
- line 69: Both versions could be used. Swapped "from year"(do ano) to "on year" (no ano) 

Please let me know if this is okay or if you would like me to revert some or all of the changes 👍

### DIFF
--- a/translations/nikola.messages/pt.po
+++ b/translations/nikola.messages/pt.po
@@ -31,7 +31,7 @@ msgid "LANGUAGE"
 msgstr "Português"
 
 msgid "Tags"
-msgstr "Etiqueta"
+msgstr "Etiquetas"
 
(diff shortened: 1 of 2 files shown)


Kwpolska:
Translations are handled [on Transifex](https://www.transifex.com/ralsina/nikola/). Please submit your changes there.

PR #2928, opened 2017-12-03
Don't refresh auto builds on files inside hidden folders

ralsina:


nikola/plugins/command/auto/__init__.py
```diff
@@ -274,11 +275,11 @@ def run_nikola_build(self, event):
         # move on larger save operations for write protection
         event_path = event.dest_path if hasattr(event, 'dest_path') else event.src_path
         fname = os.path.basename(event_path)
-        if (fname.endswith('~') or
-                fname.startswith('.') or
+        is_hidden = os.stat('filename').st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN
```
Kwpolska: `'filename'` should not be quoted.

ralsina: and should probably be fname or event_path

PR #2929, opened 2017-12-04
Baguette in bootstrap

ralsina:
* Use baguetteJS (nicer, smaller, faster than colorbox) in bootstrap theme.
* Remove colorbox

This change is backwards-incompatible since it's bound to break people's themes.


nikola/data/themes/bootstrap3/templates/base.tmpl
```diff
@@ -75,14 +75,22 @@ ${template_hooks['extra_head']()}
...
+        <script src="/assets/js/baguetteBox.min.js"></script>
+    % endif
     <!-- fancy dates -->
     <script>
     moment.locale("${momentjs_locales[lang]}");
     fancydates(${date_fanciness}, ${js_date_format});
     </script>
     <!-- end fancy dates -->
     <%block name="extra_js"></%block>
+    <script>baguetteBox.run('a.reference', {
```
Kwpolska: Why did we change the classes baguetteBox works with?

ralsina: That's copied from base theme. Probably both should be what was in bootstrap instead.

I'll take a look at it tomorrow.

Kwpolska: Ping?

ralsina:
Sorry am sorta swamped. I'll take a look this week-ish.

nikola/data/themes/bootstrap3/templates/base.tmpl
ralsina: This seems to work in the demo cases. Since baguette is not JQuery based, I don't see a simple way to use that complex selector :-(

Kwpolska:
Please backport the `.islink` thing back to `base` and run `jinjify` again.

ralsina:
Thanks for merging and reviewing @Kwpolska

PR #2943, opened 2018-01-07
Taxonomy improvements

felixfontein:
Simplified code a bit, and avoided collision check for non-generated classifications.


felixfontein:
Thanks!

I tested it with all of my blogs, seems to work fine.
I improved the collision check since I actually got such a collision; after adding this check, it worked as it should (i.e. only test "real" collisions).

So I guess it should be good enough :)

PR #2944, opened 2018-01-07
Avoid broken links when links are not as expected.

felixfontein:
When inserting HTML files with the `copy_files` task, they might use different link schemes. When `URL_TYPE` is not `rel_path`, the current code will mangle and destroy links which aren't as it expects.

The PR improves the logic to not mangle such links, but handle them correctly.

### DIFF
--- a/nikola/plugins/command/check.py
+++ b/nikola/plugins/command/check.py
@@ -328,16 +328,32 @@ class CommandCheck(Command):
                             os.path.join(os.path.dirname(filename).encode('utf-8'), unquoted_target))
 
                 elif url_type in ('full_path', 'absolute'):
+                    relative = False
                     if url_type == 'absolute':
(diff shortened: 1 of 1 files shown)


Kwpolska:
Could you show an example of a situation that would be fixed by this PR?

felixfontein:
On a site with `URL_TYPE = "full_path"`, I put the following HTML in `output/test/file.html`:
``` .html
<!DOCTYPE html>
<html lang="en">
<head>
  <link href="style.css" rel="stylesheet" type="text/css">
[...]
```
(There's a `output/test/style.css` there.)

The link checker then complains that `output/tyle.css` is missing.

Kwpolska:
Makes sense, merging.

felixfontein:
Thanks!

PR #2957, opened 2018-01-27
Preserve image ICC profiles

mchapman87501:
This improves appearance of images on systems with wide gamut displays, e.g., Retina iMacs.

### Pull Request Checklist

- [x] I tested my changes.

### Description

Images served by 'nikola serve -b' looked flat on my iMac with retina display in comparison to the corresponding original images.  Interactive testing with Pillow's Image.open and Image.save revealed that it did not automatically copy 'icc_profile' image info.

If it would be better for me to open an issue in the python-pillow/Pillow project, please let me know.  Thanks!

See the mention of available options for [https://pillow.readthedocs.io/en/latest/reference/Image.html#PIL.Image.Image.save](Image.save).

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -75,6 +75,7 @@
 * `Michael McNeil Forbes <https://github.com/mforbes>`_
 * `Michal Petrucha <https://github.com/koniiiik>`_
 * `Miguel Ángel García <https://github.com/magmax>`_
+* `Mitch Chapman <https://github.com/mchapman87501>`_
 * `mrabbitt <https://github.com/mrabbitt>`_
 * `Neil MartinsenBurrell <https://github.com/neilmb>`_
 * `Niels Böhm <https://github.com/blubberdiblub>`_
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,7 +3,7 @@ New in master
 
 Features
 --------
-
+* New PRESERVE_ICC_PROFILES option to control whether ICC profiles are preserved when copying images.
 * Use baguetteJS in bootstrap theme (part of Issue #2777)
 * New default-config command to generate a clean configuration.
 * New ``thumbnail`` shortcode similar to the reStructuredText
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -1995,6 +1995,22 @@ listed there.
 There is a huge number of EXIF tags, described in `the standard <http://www.cipa.jp/std/documents/e/DC-008-2012_E.pdf>`__
 
 
+Handling ICC Profiles
+---------------------
+
+Your images may contain `ICC profiles. <https://en.wikipedia.org/wiki/ICC_profile>`__  These describe the color space in which the images were created or captured.
+
+Most desktop web browsers can use embedded ICC profiles to display images accurately.  As of early 2018 few mobile browsers consider ICC profiles when displaying images.  A notable exception is Safari on iOS.
+
+By default Nikola strips out ICC profiles when preparing images for your posts and galleries.  If you want Nikola to preserve ICC profiles, add this in your ``conf.py``:
+
+.. code:: python
+
+  PRESERVE_ICC_PROFILES = True
+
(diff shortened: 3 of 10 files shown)


Kwpolska:
Those images may just happen to look better on your system. But ICC color
profiles are often broken on the web. Many browsers render those images
incorrectly. PNG optimizers also like to remove them. (And getting my Mac
to display reasonable colors on an external monitor was fun.) My opinion is
👎 for this PR.

-- 

mchapman87501:
Current recommended practice seems to be to embed color profiles when exporting images for the web.  Most desktop web browsers now support ICC profile management.  See the links below.

The images of concern to me are JPEG photos that embed either sRGB (e.g., "sRGB IEC61966-2.1") or wide-gamut ("Display P3") color profiles.  I've done follow-up tests to better understand cases in which such images render poorly.  In case it's of use to anyone, here's what I've found.

The tests encompassed:

- A single RAW photo exported from macOS Photos as JPEGs, embedding sRGB and Display P3 profiles
- "No profile" copies of the images above, created using PIL.open() followed by PIL.save()
- iMac Retina display and laptop displays (presumably covering the sRGB gamut)
- Safari on macOS; and Firefox and Chrome, on linux and macOS

In all cases, one image looked flat in comparison to the others:  the "no profile" image created from the "Display P3" image.

So, at least for my use case, a workaround for the current release of Nikola is to use only JPEG images exported with an embedded sRGB profile.

# Links
- https://photo.stackexchange.com/a/84558
- https://www.color-management-guide.com/web-browser-color-management.html  (starting from "Now (January 2018)!")

Kwpolska:
This sRGB PNG bug is still present in Firefox, and there is also an unrelated example of a JPEG that doesn’t work properly in Firefox at the end of this thread: https://bugzilla.mozilla.org/show_bug.cgi?id=621474

Removing the color profile fixes the sample PNGs shown in the thread. (They work fine in Chrome and Safari. Compared with Digital Color Meter.app.) While this might work better for JPEG and wide color gamuts, it might also hinder standard gamut PNG support in some cases.

mchapman87501:
Here is a bit more detail about those test cases:

The Firefox JPEG image cited (https://drive.google.com/file/d/0B_yEbP1sy9ScX09CYS1UaGxyWWs/view?usp=sharing) contains a CanonEOS5DMk2-Generic color profile.

On macOS 10.13.3, Digital Color Meter.app 5.11 shows the PNG image data (https://bugzilla.mozilla.org/show_bug.cgi?id=621474) are presented in the same way by Safari 11.0.3, Chrome 64.0.3282.119 and Firefox 58.0, with one exception: in Firefox, the sample values for the "none" (i.e., untagged) PNG image data exactly match the original hex color values - Firefox renders them without applying the system display profile.

ralsina:
Hmmm ... so, executive summary?

This is a good idea for jpeg but not for png?

Kwpolska:
Some browsers (Firefox) still have issues with color profile support. (That didn’t change since about 2011.) At the same time, a color profile can improve the appearance of images on wide color gamut displays (such as the ones found in newer Apple hardware).

mchapman87501:
Your executive summary and Chris's follow-up look good to me.

Kwpolska:
Perhaps this could be an optional setting? @mchapman87501, can you make it so ICC profiles are preserved only when the user asks for it via settings?

ralsina:
Making it an option should not be terribly hard. If @mchapman87501 can't/won't I can hack it

mchapman87501:
I'll be happy to make the changes.  It may be a day or two before I have time, if that's ok.

ralsina:
Take your time!

Kwpolska:
Thanks for contributing, and especially for improving test coverage! :tada:

You may want to add the e-mail used for the original commit (***-8.com) [to your GitHub profile](https://github.com/settings/emails) so it will appear as yours.

mchapman87501:
Thanks!  I wish I had checked my git config before that first commit - the fig8 email account was retired just at the beginning of the year.

PR #2977, opened 2018-03-08
Add a reasonable self.logger to all plugins

ralsina:
We were using self.logger in some plugins and it was never actually initialized.

This fixes that and removes some code it makes redundant in the rest compiler.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -26,14 +26,15 @@
 
 """Nikola plugin categories."""
 
-import sys
-import os
 import io
+import os
+import sys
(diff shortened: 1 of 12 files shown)


Kwpolska:
In which plugins? Are there no plugins that would create a logger twice now?

ralsina:
@Kwpolska the ones I see are:

* pandoc
* html
* markdown

These were now doing this twice (removed it from the plugin):

* ipynb
* bundles
* galleries
* scale_images
* deploy
* serve
* check

Pushing now fixes for all of them.

PR #2988, opened 2018-03-17
Remove Colorbox leftovers

Kwpolska:
There were still some Colorbox references that were not removed. This gets rid of them.


Kwpolska:
Thanks for reviewing!

Issue #2970, opened 2018-02-25
Settings (per language) for Markdown extensions


saadatm:
Some Markdown extensions allow the configuration of their output via options or settings (see, e.g., the settings for [SmartyPants](https://python-markdown.github.io/extensions/smarty/#usage) or [Footnotes](https://python-markdown.github.io/extensions/footnotes/#usage)).

Furthermore, in a multilingual website, users may want to override an extension's defaults for only some of the languages. As an example, I have a bilingual blog in English and Urdu, and when using SmartyPants, I would want to override the smart quote substitutions only for my Urdu posts, and keep using the default ones when writing in English.

So could it be possible to configure the settings for Markdown extensions on a per language basis?

ralsina:
@saadatm if you want to test the proposed PR, I could use the help, since I am not really a markdown person.

saadatm:
@ralsina Just tested the PR -- two languages, and each language having different options for Smarty Pants and Footnotes extensions -- and it's working. Thanks! 👍

ralsina:
@saadatm neat, thanks for the test

ralsina:
Fixed.

saadatm:
I upgraded to v8.0.0b1, and  `MARKDOWN_EXTENSION_CONFIGS` does not seem to be working. I have created a [test Nikola site](https://github.com/saadatm/nikola-v8-test) that demonstrates the issue. Here are the relevant bits from conf.py:

```python
MARKDOWN_EXTENSIONS = [
    'markdown.extensions.abbr',
    'markdown.extensions.footnotes',
    'markdown.extensions.smarty',
]

MARKDOWN_EXTENSION_CONFIGS = {
    DEFAULT_LANG: {
        'markdown.extensions.footnotes': {
            'BACKLINK_TEXT': '[Back]'
        },
        'markdown.extensions.smarty': {
            'substitutions': {
                'left-single-quote': '&lsaquo;',
                'right-single-quote': '&rsaquo;',
                'left-double-quote': '&laquo;',
                'right-double-quote': '&raquo;'
            }
        }
    },
    'ur': {
        'markdown.extensions.footnotes': {
            'BACKLINK_TEXT': '[واپس]'
        },
        'markdown.extensions.smarty': {
            'substitutions': {
                'left-single-quote': '&rsquo;',
                'right-single-quote': '&lsquo;',
                'left-double-quote': '&rdquo;',
                'right-double-quote': '&ldquo;'
            }
        }
    }
}
[...]

PR #2990, opened 2018-03-17
[WIP] Initial implementation of #2970, untested

ralsina:
Adds a new MARKDOWN_EXTENSION_CONFIGS translatable option, which is passed as-is to python markdown

Totally untested yet.


nikola/plugins/compile/markdown/__init__.py
```diff
@@ -89,8 +90,12 @@ def set_site(self, site):
         site_extensions = self.site.config.get("MARKDOWN_EXTENSIONS")
         self.config_dependencies.append(str(sorted(site_extensions)))
         extensions.extend(site_extensions)
+
+        site_extension_configs = self.site.config.get("MARKDOWN_EXTENSION_CONFIGS", {})
+        self.config_dependencies.append(json.dumps(site_extension_configs, sort_keys=True))
```
Kwpolska: Does it actually work like this, without actually calling the translatable setting?

ralsina: Forget it, this approach is totally broken. Looks like we need one markdown converter per-language.

Issue #2964, opened 2018-02-15
Make Bootstrap 4 the default theme


Kwpolska:
For Nikola v8.

- [ ] Move bootstrap4 to Nikola core, and bootstrap3 to the Themes repo
- [ ] Make the bootstrap4 theme independent of bootstrap3
- [ ] Get away from bower as it does not support bs4 (#2790)
- [ ] Create bootblog4 theme based on [the example](https://getbootstrap.com/docs/4.0/examples/blog/), consider putting it in core

ralsina:
+1 to making a bootblog4 and even to making it default

Kwpolska:
Work is happening on the `bootstrap4` branch.

Bootblog4 now has a regression against other themes: source links are not not visible on mobile. If anyone knows how to implement them cleanly and neatly, patches welcome.

Also welcome: ideas on how to pick featured posts. Special tags? Meta keys? Or just show the latest 3 posts there?

ralsina:
I would add a "featured" tag and just pick the N newest

PR #2991, opened 2018-03-18
Initial Bootstrap 4 implementation

Kwpolska:
This implements parts of #2964:

* implements `bootstrap4`, which looks similarly to the classic Bootstrap themes (`bootstrap3` and `bootstrap2`)
* replaces `flowr.js` with `justified-layout.js`, which does not require jQuery, is actively maintained, and comes directly from Flickr (yay for open source!)
* implements parts of the `bootblog4` theme (the base layout is mostly there)

Not done yet, in bootblog4:

* featured posts, which I’d prefer to handle as a core feature available to all willing themes
* the sidebar might not make the cut, or become a configurable optional feature
* the footer uses our standard styles right now
* bootblog4-jinja

I feel this is mergeable, and other parts can be done as future PRs.


ralsina:
I am all for it. We will never get the v8 omelette done unless we start breaking some v8 eggs.

PR #2994, opened 2018-03-21
Fix itemprop errors from w3c validator in jinja themes

edwinsteele:
https://validator.w3c.org flags itemprop attributes as an error
on any element that specifies them unless the itemtype attribute
in a parent element references a valid schema. The validation
error is:

"Error: the itemprop attribute was specified but the element is
 not a property of any item"

This change fixes the validation error.

### Pull Request Checklist

- [/] I’ve read the [guidelines for contributing](https://github.com/getnikola/nikola/blob/master/CONTRIBUTING.rst).
- [/] I updated AUTHORS.txt and CHANGES.txt (if the change is non-trivial) and documentation (if applicable).
- [/] I tested my changes.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -38,7 +38,7 @@ Features
 Bugfixes
 --------
 
-* Fix w3c validation errors for itemscope entries in jinja themes
+* Fix w3c validation errors for itemscope entries in default themes
 * Hide “Incomplete language” message for overrides of complete
   languages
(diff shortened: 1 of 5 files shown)


ralsina:
The -jinja themes are autogenerated using scripts/jinjify from their mako siblings.

Could you make the same change in base instead of base-jinja and then run jinjify?

If not, I can do it.

Kwpolska:
I can see the change in both versions.

Issue #2971, opened 2018-03-01
ATOM_PATH


davidak:
We have `RSS_PATH` but no `ATOM_PATH`.

I would like to have consistent URLs for RSS and ATOM Feeds.

Current:
RSS: https://davidak.de/rss.xml
ATOM: https://davidak.de/blog/index.atom

What i want:
RSS: https://davidak.de/rss.xml
ATOM: https://davidak.de/index.atom

(I'm not sure how important this is and if anyone uses ATOM Feeds nowadays)

ralsina:
Probably not very important, but also probably easy to do :-)

PR #3015, opened 2018-04-09
Add ATOM_PATH setting

davidak:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Add ATOM_PATH setting, similar to RSS_PATH.

Reference #2971


davidak:
I started working on this feature.

Currently i struggle to test it. Maybe Pythons [“Development Mode”](https://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode) can help.

The main functionality is still missing. Do i have to add similar lines for "atom"? https://github.com/getnikola/nikola/blob/master/nikola/plugins/task/indexes.py#L93-L94

davidak:
Now i have a shell with the dependencies installed and nikola installed with `setup.py develop`.

[Nix](https://nixos.org/nix/) is a great help as it has this integrated: https://nixos.org/nixpkgs/manual/#development-mode

Just put `default.nix` in project folder and run `nix-shell`.

```
with import <nixpkgs> {};
with pkgs;
with python36Packages;

buildPythonPackage rec {
  name = "nikola";
  src = ./.;
  propagatedBuildInputs = [
    pygments pillow dateutil docutils Mako unidecode lxml Yapsy PyRSS2Gen
    Logbook blinker setuptools natsort requests piexif markdown phpserialize
    jinja2 doit lessc sass 
  ];
}
```
This works at least on Linux, on macOS is a dependency missing i think. Would you like a PR for that?

Now i'm able to build my website, no change in atom path as expected.

nikola/plugins/task/indexes.py
```diff
@@ -92,7 +92,8 @@ def get_path(self, classification, lang, dest_type='page'):
         """Return a path for the given classification."""
         if dest_type == 'rss':
             return [self.site.config['RSS_PATH'](lang)], True
```
felixfontein: Please change the `True` to `'auto'` (I apparently forgot to change this a long time ago).

nikola/plugins/task/indexes.py
```diff
@@ -92,7 +92,8 @@ def get_path(self, classification, lang, dest_type='page'):
...
-        # 'page' (index) or 'feed' (Atom)
+        if dest_type == 'feed':
+            return [self.site.config['ATOM_PATH'](lang)], True
```
felixfontein: Please also change the `True` to `"auto"` here.

felixfontein:
Apropos, I'm ususally testing with `pip install -e /path/to/nikola/git/checkout`. Not sure how that works with Nix, though.

davidak:
@felixfontein Nix does that, this way pip installs every dependency that is not installed by Nix into /tmp/.... This way you get a temporary development environment without messing up your python install.

When i set

```
RSS_PATH = INDEX_PATH
ATOM_PATH = INDEX_PATH
```

the links in the `index.html` are

```
<link rel="alternate" type="application/rss+xml" title="RSS" href="blog.xml">
--
  | <link rel="alternate" type="application/atom+xml" title="Atom" href="blog.atom">
```
Now the RSS and Atom link are similar, but have strange file names. Still breaking backwards compatibility.

With this setting
```
RSS_PATH = INDEX_PATH + "/rss"
ATOM_PATH = INDEX_PATH + "/index"
```
the links are

```
<link rel="alternate" type="application/rss+xml" title="RSS" href="blog/rss.xml">
--
  | <link rel="alternate" type="application/atom+xml" title="Atom" href="blog/index.atom">
```
I can set this for Atom and leave RSS as is, then it's backwards compatible. But we should change this in 8.0 to have consistent links among RSS and Atom by default.

nikola/plugins/task/indexes.py
davidak: Is `'auto'` ok or exactly `"auto"`? I think it makes no difference and is better in this case.

felixfontein: Ah sorry, I wanted to write `'auto'` here as well :)

Kwpolska:
As you noticed in #3016, Atom feeds work differently to RSS. We generate them with the same data as `index.html`, only in a different storage format. With that, I’m not sure if this PR/feature request will work out.

> ```py
> 'ATOM_PATH': INDEX_PATH + '/index', # TODO: '' in 8 to be consistent with RSS
>```

This feature will be v8-only.

To do this properly, you’d do it after config is loaded from `conf.py`, and then use something like:

```py
config['ATOM_PATH'] = config['ATOM_PATH'] if config['ATOM_PATH'] else config['INDEX_PATH']
```

> [Nix stuff] works at least on Linux, on macOS is a dependency missing i think. Would you like a PR for that?

I don’t think we need that. The more typical route (virtualenv + `pip install -e .`) is enough for most people, and we really don’t want yet another dependency list to update.

davidak:
>As you noticed in #3016, Atom feeds work differently to RSS. We generate them with the same data as index.html, only in a different storage format.

@Kwpolska What is the reason for that? It seems not to be useful to subscribe to the `index-1.atom`, since it will never change. Or is there a different use case for Atom feeds?

With your suggested change, i get `href="blog.atom">` when `ATOM_PATH = ''` and `INDEX_PATH = "blog"` in config.py. Is that what you want or did you want the past behavior?

That can be achived with this lines added in nikola.py at line 616:

```
        # Use ATOM_PATH when available
        self.config['ATOM_PATH'] = self.config['ATOM_PATH'] or self.config['INDEX_PATH'] + '/index'
```
Resulting in `href="blog/index.atom">`.

I suggest to have the same behavior with RSS and Atom path, at least in 8 since it breaks backwards compatibility.

Kwpolska:
That’s the way this feature was written. I suppose the use-case was to have pagination between atom feeds. But perhaps we could clean it up and drop that?

My code snippet was just an untested example. That said, if you’re letting people customize the path, you should also ask for the file extension (some people might want `atom.xml` or whatever)

davidak:
>But perhaps we could clean it up and drop that?

If there is no good reason for this behavior, sure. That's a topic for the other issue.

>My code snippet was just an untested example. That said, if you’re letting people customize the path, you should also ask for the file extension (some people might want atom.xml or whatever)

My first idea was to have `/rss.xml` and `/atom.xml` for my blog, but after a little research i found out that the `.atom` file extension is recommended in [RFC 4287](https://www.ietf.org/rfc/rfc4287.txt). So i will leave it `index.atom`. Should we still let people change it?

What default behavior do you want for this feature?

1. `INDEX_PATH + '/index'` (backwards compatible)
2. `index` (consistent with RSS)
3. something else

Kwpolska:
I vote for option 1.

Kwpolska:
Are you sure it’s broken? It’s percent-encoding, and looks like UTF-8, so it will work in most sane environments. (Will work on modern Linux. May crash on macOS and Windows if web server is not careful enough.)

davidak:
It works, but look strange. So, no problem.

felixfontein:
I've tried running your branch with one of my blogs, and I got:
```
TaskError - taskid:render_taxonomies:/index.atom
PythonAction Error
  File "/usr/lib/python3.6/site-packages/doit/action.py", line 424, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/home/.../nikola/nikola.py", line 2401, in atom_feed_renderer
    with io.open(output_path, "w+", encoding="utf-8") as atom_file:
PermissionError: [Errno 13] Permission denied: '/index.atom'
```
I didn't set `ATOM_PATH`, and `INDEX_PATH` is `''` (default value).

After replacing `self.config['INDEX_PATH'] + '/index'` with `os.path.join(self.config['INDEX_PATH'], 'index')` (probably doesn't work under Windows, but for a quick test that's enough). After this change, everything worked (and produced the same output as before). So I guess you need to catch the case where `self.config['INDEX_PATH']` is empty or ends with a slash. (Too bad `os.path.join` cannot be used under Windows...)

Kwpolska:
@tritium21 you’re missing context. `os.path.join` works fine on Windows to produce paths that use backslashes. Which can’t be used in URLs.

tritium21:
@Kwpolska if you're building URLs (not for paths for files on disk, which i assumed the context was), then, yes, the os.path alias is not what to use.  import posixpath directly (yes, it is always available, even on windows)

felixfontein:
@davidak: you'll need to rebase this. The behavior of `RSS_PATH` changed (see #3036), so your implementation of `ATOM_PATH` probably needs to be adjusted accordingly. Actually, this should make the `+ '/index'` part superfluous.

felixfontein:
It now works with my blogs (using the default `ATOM_PATH` and `INDEX_PATH` being unset).

@davidak: does the current version also works with your blog?

felixfontein:
Another thing to look at: https://github.com/getnikola/nikola/commit/987f18610fb76545d0ca4d705e6f6dd976445949

felixfontein:
If this is merged after #3043, `ATOM_PATH` needs to be adjusted as in that PR. Otherwise, if #3043 is merged after this one, it has to be adjusted. Whatever comes last :-)

Kwpolska:
#3043 was just merged.

tests/data/translated_titles/conf.py
```diff
@@ -225,6 +225,10 @@
 # output / TRANSLATION[lang] / RSS_PATH / rss.xml
 # RSS_PATH = ""
 
+# Final location for the blog main Atom feed is:
+# output / TRANSLATION[lang] / ATOM_PATH / index.atom
+# ATOM_PATH = ""
```
Kwpolska: (FYI, this file does not need to be updated.)

nikola/plugins/task/indexes.py
```diff
@@ -95,7 +95,8 @@ def get_path(self, classification, lang, dest_type='page'):
                 self.site.config['RSS_PATH'](lang),
                 self.site.config['RSS_FILENAME_BASE'](lang)
             ], 'auto'
-        # 'page' (index) or 'feed' (Atom)
+        if dest_type == 'feed':
+            return [self.site.config['ATOM_PATH'](lang)], 'always'
```
Kwpolska: @felixfontein: why is page number support not implemented for feeds?

tests/data/translated_titles/conf.py
felixfontein: Should we leave this change anyway?

nikola/plugins/task/indexes.py
felixfontein: Well, that would be a simple change. But:
 1. nobody asked for it so far (I think);
 2. there's still the big open question why the many .atom feed files are needed after all (#3016).
I'd like to see an answer to #3016 first.

tests/data/translated_titles/conf.py
Kwpolska: Whatever.

Kwpolska:
Thanks for contributing, @davidak!

PR #3026, opened 2018-04-14
Fix #3022 -- Explain index.html conflicts better

### LINKED ISSUE #3022: render_taxonomies:index.html conflicts with pages/index.html

**Python Version:** 3.5.3

**Nikola Version:** v7.8.14

**Operating System:** Debian

### Description:

I get the following error:

```
ERROR: Two different tasks can't have a common target.'output/index.html' is a target for render_taxonomies:output/index.html and render_pages:output/index.html.
```

My config has

```
PAGES = (
    ("pages/*.md", "", "page.tmpl"),
)
```

and `pages/index.md` exists.

This is similar to issue #2724 except that the problem is caused by taxonomies.  Making `taxonomy.should_generate_classification_page` return `False` in `nikola/plugins/task/taxonomies.py` is a workaround.

Now in addition to being a bug in the code, I think there's also a documentation problem here. This is a fresh nikola installation.  My conf.py doesn't mention taxonomy/taxonomies at all and searching google doesn't result obvious results on how to disable taxonomies (whatever that is).  Ahh, ok, the manual says "three taxonomy systems in Nikola" and tags and categories are enabled in the default config.  Still, maybe add the sentence about the three taxonomy systems to the config file.

Kwpolska:
New behavior:

```
$ nikola build
[2018-04-14T17:09:38Z] ERROR: Nikola: Post 'pages/index.rst': output path (output/index.html) conflicts with the blog index (output/index.html). Please change INDEX_PATH or disable index generation.
[2018-04-14T17:09:39Z] ERROR: Nikola: Post 'pages/index.rst': output path (output/es/index.html) conflicts with the blog index (output/es/index.html). Please change INDEX_PATH or disable index generation.
ERROR: Two different tasks can't have a common target.'output/index.html' is a target for render_taxonomies:output/index.html and render_pages:output/index.html.
```

cc @felixfontein @tbm

### DIFF
--- a/nikola/plugins/task/pages.py
+++ b/nikola/plugins/task/pages.py
@@ -45,14 +45,13 @@ class RenderPages(Task):
             "filters": self.site.config["FILTERS"],
             "show_untranslated_posts": self.site.config['SHOW_UNTRANSLATED_POSTS'],
             "demote_headers": self.site.config['DEMOTE_HEADERS'],
-            "disable_indexes": self.site.config["DISABLE_INDEXES_PLUGIN_INDEX_AND_ATOM_FEED"],
         }
         self.site.scan_posts()
         yield self.group_task()
         index_paths = {}
         for lang in kw["translations"]:
             index_paths[lang] = False
-            if not kw["disable_indexes"]:
+            if not self.site.config["DISABLE_INDEXES_PLUGIN_INDEX_AND_ATOM_FEED"]:
                 index_paths[lang] = os.path.normpath(os.path.join(self.site.config['OUTPUT_FOLDER'],
                                                      self.site.path('index', '', lang=lang)))
 


tbm:
I just applied the patch and this is much clearer.

Thank you!

nikola/plugins/task/pages.py
```diff
@@ -43,9 +45,17 @@ def gen_tasks(self):
             "filters": self.site.config["FILTERS"],
             "show_untranslated_posts": self.site.config['SHOW_UNTRANSLATED_POSTS'],
             "demote_headers": self.site.config['DEMOTE_HEADERS'],
+            "disable_indexes": self.site.config["DISABLE_INDEXES_PLUGIN_INDEX_AND_ATOM_FEED"],
```
felixfontein: Is it a good idea to put this into `kw`? After all, this value is only needed at task generation time, and its use does not require a rebuild of all pages if it changes.

nikola/plugins/task/pages.py
```diff
@@ -55,6 +65,12 @@ def gen_tasks(self):
                 else:
                     context = {'pagekind': ['story_page', 'page_page']}
                 for task in self.site.generic_page_renderer(lang, post, kw["filters"], context):
+                    if task['name'] == index_paths[lang]:
+                        # Issue 3022
+                        LOGGER.error(
+                            "Post {0!r}: output path ({1}) conflicts with the blog index ({2}). "
+                            "Please change INDEX_PATH or disable index generation.".format(
+                                post.source_path, task['name'], index_paths[lang]))
```
felixfontein: Maybe mention `DISABLE_INDEXES_PLUGIN_INDEX_AND_ATOM_FEED` as a way to disable index generation. That should make it easier for users to react on this message. (And they also see that this affects Atom feed generation.)

Kwpolska: Speaking of which, this config setting has a terrible name which should be changed in v8.

nikola/plugins/task/pages.py
Kwpolska: Sure, it’s not really needed.

PR #3036, opened 2018-04-15
Fixing RSS_PATH behavior.

### LINKED ISSUE #3024: RSS_PATH doesn't work as advertised (is path and filename, excluding .xml)
* Python Version: 3.5.3
* Nikola Version: v7.8.14
* Operating System: Debian

A fresh config says:

```
# Final location for the blog main RSS feed is:
# output / TRANSLATION[lang] / RSS_PATH / rss.xml
```

which is in line with other `_PATH` variables.

But it seems `RSS_PATH` is actually path+filename (and `.xml` is appended).

With `RSS_PATH = "blog/`I get `render_taxonomies:output/blog/.xml` (instead of `blog/rss.xml`)

With `RSS_PATH = blog/index.xml` I get `render_taxonomies:output/blog/index.xml.xml`

felixfontein:
Fixes #3024 so that `RSS_PATH` behaves as advertised.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,8 @@ New in master
 Features
 --------
 
+* Fixing behavior of RSS_PATH to do what the documentation
+  says it does (Issue #3024)
 * Add support for fragments in path handlers (Issue #3032)
 * Recognize both TEASER_END and (new) END_TEASER (Issue #3010)
 * New MARKDOWN_EXTENSION_CONFIGS setting (Issue #2970)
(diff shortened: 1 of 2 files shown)


Kwpolska:
Please:
* add a changelog entry
* show the new behavior (what is generated if `RSS_PATH = 'foo'`, `RSS_PATH = 'bar/'`?)

felixfontein:
`RSS_PATH == 'foo'` should result in `output/en/foo/rss.xml`, and `RSS_PATH == 'bar/'` in `output/en/bar//rss.xml`. So far I've only tested it with `RSS_PATH == ''` (the default value), which results in no behavior change (as expected).

Kwpolska:
Works fine. I’ll handle the v7 backport (which is a reasonable idea).

felixfontein:
Cool! Thanks also for backporting!

PR #3038, opened 2018-04-15
Add element role to navigation bar

### LINKED ISSUE #3035: HTML validation errors with default bootstrap3 theme
I see the following validation errors from https://validator.w3.org with the default bootstrap3 theme.

On all pages:

```
Error: Element div is missing one or more of the following attributes: role.
From line 36, column 9; to line 36, column 83
>↩        <div class="collapse navbar-collapse" id="bs-navbar" aria-expanded="false">↩     
```

On all posts (blog):

```
Error: The itemprop attribute was specified, but the element is not a property of any item.
From line 106, column 44; to line 106, column 90
or vcard"><span class="byline-name fn" itemprop="author">↩     
```

tbm:
Add element `role` to the navigation bar to avoid the validation
error: `Element div is missing one or more of the following
attributes: role.`

Fixes #3035

### DIFF
--- a/nikola/data/themes/bootstrap3-jinja/templates/base.tmpl
+++ b/nikola/data/themes/bootstrap3-jinja/templates/base.tmpl
@@ -32,7 +32,7 @@
             {% endif %}
             </a>
         </div><!-- /.navbar-header -->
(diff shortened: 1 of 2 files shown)


tbm:
@Kwpolska I know you don't care but I care about validation issues and it's a simple fix. :)

PR #3045, opened 2018-04-15
Make crumbs available in all pages

tbm:
Make the crumbs variable available in all pages so templates can
have access to it.

@Kwpolska suggested I create a plugin to have breadcrumbs for pages but I don't see any reason not to make `crumbs` available all the time so templates can have access to it.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2184,7 +2184,9 @@ class Nikola(object):
         context['description'] = post.description(lang)
         context['permalink'] = post.permalink(lang)
         if 'crumbs' not in context:
-            crumb_path = post.permalink(lang).lstrip('/').rstrip(self.config['INDEX_FILE'])
+            crumb_path = post.permalink(lang).lstrip('/')
+            if crumb_path.endswith(self.config['INDEX_FILE']):
+                crumb_path = crumb_path[:-len(self.config['INDEX_FILE'])]
             if crumb_path.endswith('/'):
                 context['crumbs'] = utils.get_crumbs(crumb_path.rstrip('/'), is_file=False)
             else:
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -2183,6 +2183,12 @@ class Nikola(object):
         context['title'] = post.title(lang)
         context['description'] = post.description(lang)
         context['permalink'] = post.permalink(lang)
+        if 'crumbs' not in context:
+            crumb_path = post.permalink(lang).lstrip('/').rstrip(self.config['INDEX_FILE'])
+            if crumb_path.endswith('/'):
+                context['crumbs'] = utils.get_crumbs(crumb_path.rstrip('/'), is_file=False)
+            else:
+                context['crumbs'] = utils.get_crumbs(crumb_path, is_file=True)
         if 'pagekind' not in context:
             context['pagekind'] = ['generic_page']
         if post.use_in_feeds:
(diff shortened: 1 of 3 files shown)


nikola/nikola.py
```diff
@@ -2183,6 +2183,7 @@ def generic_page_renderer(self, lang, post, filters, context=None):
         context['title'] = post.title(lang)
         context['description'] = post.description(lang)
         context['permalink'] = post.permalink(lang)
+        context['crumbs'] = utils.get_crumbs(post.destination_path(lang, extension))
```
Kwpolska: This may override another `crumbs` variable in the context.  Perhaps do `if 'crumbs' not in context`?

tbm:
I just noticed another issue. I'll post a new version soon.

tbm:
I think the new version works correctly regardless of whether `PRETTY_URLS` is true or false. At least it worked in my tests.

nikola/nikola.py
```diff
@@ -2183,6 +2183,11 @@ def generic_page_renderer(self, lang, post, filters, context=None):
         context['title'] = post.title(lang)
         context['description'] = post.description(lang)
         context['permalink'] = post.permalink(lang)
+        if 'crumbs' not in context:
+            if post.permalink(lang).endswith('/'):
```
Kwpolska: How does this work if `STRIP_INDEXES = False`?

tbm: If it's `False`', it's a file (which probably ends with `.html`) and files are dealt with in the else clause.

So for ```posts/foo.html``` we'd end up with
```
context['crumbs'] = [['.', 'posts'], ['#', 'foo.html']]
```

I've also tested `foo/bar/' which leads to
```
context['crumbs'] = [['..', 'foo'], ['#', 'bar']]
```

Hmm, I guess `PRETTY_URLS = True` and `STRIP_INDEXES = False` is not handled ideally (although not wrong either). `foo/bar/index.html` would be:
```
context['crumbs'] = [['..', 'foo'], ['.', 'bar'], ['#', 'index.html']]
```

Is this your concern?  Or did I miss some other case?

Kwpolska: Yeah, that last one is kinda concerning (although this use case is a strange one).  You can do `permalink.endswith(self.config['INDEX_FILE'])` and then do some extra logic.

tbm:
This covers all combinations I can think of.

nikola/nikola.py
```diff
@@ -2183,6 +2183,12 @@ def generic_page_renderer(self, lang, post, filters, context=None):
         context['title'] = post.title(lang)
         context['description'] = post.description(lang)
         context['permalink'] = post.permalink(lang)
+        if 'crumbs' not in context:
+            crumb_path = post.permalink(lang).lstrip('/').rstrip(self.config['INDEX_FILE'])
```
Kwpolska: rstrip doesn’t do what you think it does. The string is used as a list of characters to find and remove, not as a sequence. You need to remove it somehow else:

```python
crumb_path = post.permalink(lang).lstrip('/')
if crumb_path.endswith(self.config['INDEX_FILE']):
    crumb_path = crumb_path[:-len(self.config['INDEX_FILE'])]
```

Kwpolska:
Thanks for contributing this! :tada:

tbm:
Thank you.

Any chance of getting this into v7 too? I know it's more of a feature but you could argue not exporting the crumbs to posts is a bug ;)

PR #3058, opened 2018-04-19
Remove warning about Debian from issue template

### LINKED ISSUE #3057: Debian no longer offers nikola
The issue template says

> Nikola v7.6.4, as provided by Debian/Ubuntu, is NOT SUPPORTED.

Nikola was [removed from Debian in July 2017](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=863031) so I think it's time to update the issue template.

I don't know about Ubuntu but I can check (unless you know it already).

tbm:
Debian no longer offers Nikola, neither in the development tree
("Debian unstable") nor in any supported releases.

Fixes #3057


Kwpolska:
Not really important, but okay.

PR #3060, opened 2018-04-22
Remove max-width for images

Kwpolska:
IMO, this causes more harm than good.  Is there anyone or anything who depends on this or otherwise wants this behavior? (This needed working around on themes.getnikola.com *and* chriswarrick.com recently)

### DIFF
--- a/nikola/data/themes/bootstrap4-jinja/assets/css/theme.css
+++ b/nikola/data/themes/bootstrap4-jinja/assets/css/theme.css
@@ -10,10 +10,6 @@
     width: 150px;
 }
 
-img {
-    max-width: 90%;
-}
-
 .titlebox {
     text-align: right;
 }
--- a/nikola/data/themes/bootstrap4/assets/css/theme.css
+++ b/nikola/data/themes/bootstrap4/assets/css/theme.css
@@ -10,10 +10,6 @@
     width: 150px;
 }
 
-img {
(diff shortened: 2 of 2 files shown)


ralsina:
Well, if you have a SVG it often will be much more than 100% wide on the page.

felixfontein:
But why `max-width: 90%` then? I'd use `max-width: 100%` instead.

ralsina:
@felixfontein because it looked pretty.

ralsina:
Hope nobody was expecting a deep reason :-)

felixfontein:
@ralsina I was expecting something like that ;-)

@Kwpolska does `max-width: 100%` causes the same problems as `90%` for your page and the themes page?

Kwpolska:
100% just-so-happens to work in the specific cases (as they have `width: 100%`). I could accept setting max-width to 100%.

felixfontein:
I think having some kind of `max-width` for images is always a good idea. I mean, it can be disabled for more specific situations (`.specific-selector img { max-width: initial; }` in CSS), and it will fix probably enough blog posts where a too big image is included (for example a non-properly scaled SVG, or a JPEG directly copied from the camera's SD card) without postprocessing. And it does improve the situation when looking at a blog with a low-width device (such as a phone), in which situation also images with moderate width can screw up the layout.

PR #3064, opened 2018-04-26
Arbitrary metadata for galleries (#3050)

### LINKED ISSUE #3050: Arbitrary metadata in galleries’ metadata.yml files
It would be nice if `metadata.yml` supported any arbitrary metadata (inserted directly into photo_array dicts)

@jmcp: if you want to take it, go ahead.  Otherwise I can do it.

Kwpolska:
This fixes #3050.

cc @jmcp.  (btw, there was a bug with `order: 0` being ignored)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -18,7 +18,9 @@ Features
 
 * Produce a better error message when a template referenced in another
   template is missing (Issue #3055)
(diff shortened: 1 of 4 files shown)


jmcp:
Looks good - and thankyou for catching that bug.

Issue #2840, opened 2017-06-13
conf.py has too many features


Kwpolska:
(Those are only some propositions and examples. If someone disagrees with my opinions, or wants some other features gone, feel free to comment.)

### Unnecessary customization

- [ ] `TRANSLATIONS_PATTERN` — this is useful for backwards compatibility with multilingual sites created with Nikola v6 or older. Perhaps we could just offer some auto-upgrade and cut some code?
- [ ] `TAGLIST_MINIMUM_POSTS`
- [x] `FEED_PREVIEWIMAGE` — how many people don’t want this?
- [x] `USE_OPEN_GRAPH` — how many people don’t want this?
- [x] `SITEMAP_INCLUDE_FILELESS_DIRS` — how many people care?
- [ ] `INDEX_FILE` — how many people don’t want 'index.html'?

### Unneeded features

- [x] `USE_BASE_TAG` — this feature never helped anyone, is a quirk of the HTML spec, and has tripped some people (you must use nikola serve, or a httpd, and if you want multiple domain names, you need to disable that feature anyway)
- [ ] `SOCIAL_BUTTONS_CODE` — can also be done with `BODY_END`
- [ ] `LICENSE` — can be done inside `CONTENT_FOOTER` proper

### Split out to plugins?

- [ ] `DATE_FANCINESS`
- [ ] `WRITE_TAG_CLOUD`
- [x] `LESS_COMPILER`/`LESS_OPTIONS`/`SASS_COMPILER`/`SASS_OPTIONS` — those are already plugins but their options are still in conf.py
- [ ] `HYPHENATE` (→ filter?)
- [ ] `FILE_METADATA_REGEXP` / `FILE_METADATA_UNSLUGIFY_TITLES`

**Others:** We should rethink filter configuration.

felixfontein:
`TAGLIST_MINIMUM_POSTS`: I'm voting for keeping that. As soon as your blog gets bigger, it's really nice to restrict your tag list to tags with a certain minimum number of posts. Otherwise, your tag list is just way too long.

`FEED_PREVIEWIMAGE`: The only thing this is doing is disabling preview images (when set to `False`) if `previewimage` is in the metadata, right? So if you don't want preview images, don't add `previewimage` metadata fields. So for me, there's no reason to keep `FEED_PREVIEWIMAGE`.

`USE_OPEN_GRAPH`: I guess I don't care since I have my own theme without it anyway ;-)

[...]

PR #3065, opened 2018-04-26
Drop useless settings

Kwpolska:
The following settings have been removed. Nikola will now always behave as if the value was what is displayed afer the setting name.

* ``FEED_PREVIEWIMAGE =  True``
* ``SITEMAP_INCLUDE_FILELESS_DIRS = True``
* ``USE_OPEN_GRAPH = True``
* ``USE_BASE_TAG = False``

This is part of #2840. Not all settings listed there were deleted.


ralsina:
Other than the conflicts, LGTM +1

Issue #2761, opened 2017-05-13
Change drafts to be its own metadata instead of special tag


h4ckninja:
For a future version, I'd like to see Nikola change the way it handles drafts.

Today, drafts are handled in a way that isn't as straight-forward to the user by having a special tag.

Instead, I'd like to see drafts defined by a separate piece of metadata:

~~~
.. title: My Super Title Here About Python
.. slug: my-super-title-here-about-python
.. date: 2017-05-13 00:00:00 UTC
.. tags: python, pip, venv
.. section: Developers
.. description: My Python Post
.. status: published
~~~

**Why?**

By having an explicit status of draft or published, and with a default set in `conf.py`, the user can clearly see that a post is published or not. As an example, `conf.py` may have a default post setting:

~~~
DEFAULT_POST_STATUS='draft'
~~~

"But tags already do that." Sort of. I see this as a combination of User Experience and a technical abstraction of post creation to be extensible. By keeping special meaning out of tags, users:

- won't accidentally remove a draft tag while editing their post (UX)
- the user has to explicitly say a post is ready to be published (or explicitly say it is draft) making it clear what's happening to the content (UX)
- this can later on allow extensible post compiling to do things like have separate configs for publishing for only posts that are marked as having a status of published by having a config for dev and a config for prod.

felixfontein:
I'm in favor of this; I'd extend this to get rid of *all* special tags and use metadata instead. The other special tags besides `draft` I know of are `retired`, `private`, `mathjax`. (`retired` is marked as "remove in v8".)

Another (probably the main!) reason why I don't like the "use special tags as metadata" is that you cannot use these tags for real tagging. Like what if you want to have a tag `private` for posts about your private life? (As opposed to whatever else you're blogging about normally.) You simply cannot do that. Or if you write a post *about* mathjax and want to distinguish it from posts which *use* mathjax? No way to do that with tags. Or if you want to write a post about you being retired? You cannot use `retired` as a tag for that. Also, draft has [many different meanings](https://en.wikipedia.org/wiki/Draft), so you might want to tag a post with `draft` if it is about something in that list.

felixfontein:
Out of curiosity: why are tags used for this in Nikola at all? I'd speculate that this is rooted in the old-style metadata, where text being on a specific line had a specific interpretation, so it wasn't easy/feasible to add extra metadata. But with modern metadata (`.. xxx: yyy`), that's not relevant anymore.

[...]

PR #3068, opened 2018-05-01
[WIP] Conversion plugin for #2761

felixfontein:
Just checking in the stubs I had so far. Does nothing yet. Completes #2761, continues #3066.

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -243,6 +243,7 @@ class Post(object):
         is_private = False
         post_status = 'published'
         self._tags = {}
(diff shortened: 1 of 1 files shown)


Kwpolska:
I think this should be optional (in the plugins repo) and not a built-in plugin, since every user will need this at most once.

felixfontein:
Sure. A little `post.py` change will be necessary anyway, that part can still be in here I guess :-)

I'll work on this tomorrow and then update this branch. (Force push, so don't check this out yet ;-) )

PR #3082, opened 2018-05-09
Fix reverse option for post_list

tbm:
reverse is never None since it's True or False.

### DIFF
--- a/nikola/plugins/shortcode/post_list.py
+++ b/nikola/plugins/shortcode/post_list.py
@@ -154,7 +154,7 @@ class PostListShortcode(ShortcodePlugin):
 
         filtered_timeline = []
         posts = []
-        step = -1 if reverse is None else None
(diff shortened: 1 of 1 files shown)


Kwpolska:
Are you sure? The default argument passed to the function is `False`.

tbm:
> Are you sure? The default argument passed to the function is False.

My Python isn't good but it seems if something is False it's not None.

```
>>> reverse = False
>>> if reverse is None: print('yes')
... 
>>> reverse = True
>>> if reverse is None: print('yes')
... 
>>> reverse = None
>>> if reverse is None: print('yes')
... 
yes
>>> 
```

So basically the current test doesn't make sense because reverse is never "None".  It's either True or False.

Kwpolska:
It *is* None if the input comes from reST. And the `-1` argument makes stuff reversed. I fixed it in commit 2cb25fb.

PR #3085, opened 2018-05-09
Support CATEGORY_DESTPATH_NAMES with pages following destpath

Kwpolska:
This should make stuff work properly for all cases I can think of.

hat tip @tbm (I used your website for testing)

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -768,14 +768,6 @@ class Nikola(object):
             utils.LOGGER.error('CATEGORY_PAGES_FOLLOW_DESTPATH requires CATEGORY_ALLOW_HIERARCHIES = True, CATEGORY_OUTPUT_FLAT_HIERARCHY = False.')
             sys.exit(1)
 
-        # Needed to undo names for CATEGORY_PAGES_FOLLOW_DESTPATH
-        destpath_names_reverse = {}
-        for lang in self.config['TRANSLATIONS']:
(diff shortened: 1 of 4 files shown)


nikola/nikola.py
```diff
@@ -768,6 +768,14 @@ def __init__(self, **config):
             utils.LOGGER.error('CATEGORY_PAGES_FOLLOW_DESTPATH requires CATEGORY_ALLOW_HIERARCHIES = True, CATEGORY_OUTPUT_FLAT_HIERARCHY = False.')
             sys.exit(1)
 
+        # Needed to undo names for CATEGORY_PAGES_FOLLOW_DESTPATH
+        destpath_names_reverse = {}
+        for lang in self.config['TRANSLATIONS']:
+            destpath_names_reverse[lang] = {}
+            for k, v in self.config['CATEGORY_DESTPATH_NAMES'](lang).items():
+                destpath_names_reverse[lang][v] = k
+        self.config['_CATEGORY_DESTPATH_NAMES_REVERSE'] = utils.TranslatableSetting('_CATEGORY_DESTPATH_NAMES_REVERSE', destpath_names_reverse, self.config['TRANSLATIONS'])
```
felixfontein: What do you think about doing this computation in the categories plugin (in its `set_site` method) and storing the result there?
I'm currently tending to prefer this, but I'm not really sure...

PR #3087, opened 2018-05-12
Bootblog4 featured posts and sidebar

Kwpolska:
This adds featured posts and a sidebar to bootblog4. Testing needed.
screenshot 2018-05-13 00 16 25

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -570,6 +570,7 @@ class Nikola(object):
             'TEMPLATE_FILTERS': {},
             'THEME': LEGAL_VALUES['DEFAULT_THEME'],
             'THEME_COLOR': '#5670d4',  # light "corporate blue"
+            'THEME_CONFIG': {},
             'THUMBNAIL_SIZE': 180,
             'TRANSLATIONS_PATTERN': '{path}.{lang}.{ext}',
             'URL_TYPE': 'rel_path',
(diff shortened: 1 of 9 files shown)


ralsina:
Does it move the search form to the sidebar? That may fix our other issue :-)

Kwpolska:
`SEARCH_FORM` will stay in the top navigation, but with some other tricks needed (so it would collapse). The sidebar is optional; users are free to put anything — including the search form — in their sidebar.

nikola/plugins/task/indexes.py
```diff
@@ -117,6 +117,7 @@ def provide_context_and_uptodate(self, classification, lang, node=None):
             "title": self.site.config["INDEXES_TITLE"](lang) or self.site.config["BLOG_TITLE"](lang),
             "description": self.site.config["BLOG_DESCRIPTION"](lang),
             "pagekind": ["main_index", "index"],
+            "featured": [p for p in self.site.posts if p.post_status == 'featured' and lang in p.translated_to],
```
felixfontein: How about untranslated featured posts, when `SHOW_UNTRANSLATED_POSTS` is set to `True`?

nikola/nikola.py
```diff
@@ -2581,6 +2585,13 @@ def generic_index_renderer(self, lang, posts, indexes_title, template_name, cont
                     context['current_page'] = ipages_i - 1
                 context['prev_next_links_reversed'] = kw['indexes_static']
             context["permalink"] = page_links[i]
+            context["is_frontmost_index"] = i == 0
+
+            # Add dependencies to featured posts
+            if 'featured' in context:
```
felixfontein: I think we should mention that in the docstring of `generic_index_renderer`, in case someone wants to put something else called `featured` in there.

nikola/plugins/task/indexes.py
Kwpolska: Fair point, will do this in a while.

PR #3096, opened 2018-05-22
gallery: use JS appendChild instead of append

### LINKED ISSUE #3095: galleries don't work in Edge/Internet Explorer

**Python Version:**
3.6.5

**Nikola Version:**
v7.8.15 and v8.0.0b2

**Operating System:**
Arch Linux x86-64

### Description:
Galleries are broken on Internet Explorer and Edge.  Attempting to display one throws a JavaScript error:
  `Object doesn't support property or method 'append'`

The demo gallery is sufficient to reproduce this problem:
  https://getnikola.com/galleries/demo/

The default gallery is trying to use ParentNode.append(), which is not supported by either IE or Edge:

  https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append#Browser_compatibility

But since it looks like it's only trying to append one child at a time, it looks like it could use Node.appendChild(), which is universally supported.  At least, when I customize gallery.tmpl to call appendChild instead of append, everything starts working in IE and Edge.

evetsso:
It's only appending one child anyway, and append is not supported (yet?) by
IE and Edge.

Fixes #3095.

### Pull Request Checklist

- [X] I tested my changes.

### Description
Hopefully self-explanatory.  The Disqus JS appears to already by calling appendChild too.

### DIFF
--- a/nikola/data/themes/base-jinja/templates/gallery.tmpl
+++ b/nikola/data/themes/base-jinja/templates/gallery.tmpl
@@ -91,9 +91,9 @@ function renderGallery() {
         div.style.height = boxes[i].height + 'px';
         div.style.top = boxes[i].top + 'px';
         div.style.left = boxes[i].left + 'px';
-        link.append(img);
-        div.append(link);
-        container.append(div);
+        link.appendChild(img);
+        div.appendChild(link);
+        container.appendChild(div);
     }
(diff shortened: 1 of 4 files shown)


Kwpolska:
Sigh, why is IE still a thing?

PR #3120, opened 2018-07-17
Only listing directories that exist.

okin:
### Pull Request Checklist

- [x] I updated CHANGES.txt.
- [x] I tested my changes.

### Description

While testing Nikola 8.0.0b2 I discovered that listing installed themes does not work for me (setup with pipenv).

```
  File "/Users/niko/.virtualenvs/bikes.bembel.bytes-CZ5WGSwq/lib/python3.6/site-packages/doit/doit_cmd.py", line 177, in run
    return command.parse_execute(args)
  File "/Users/niko/.virtualenvs/bikes.bembel.bytes-CZ5WGSwq/lib/python3.6/site-packages/doit/cmd_base.py", line 127, in parse_execute
    return self.execute(params, args)
  File "/Users/niko/.virtualenvs/bikes.bembel.bytes-CZ5WGSwq/lib/python3.6/site-packages/nikola/plugin_categories.py", line 147, in execute
    return self._execute(options, args)
  File "/Users/niko/.virtualenvs/bikes.bembel.bytes-CZ5WGSwq/lib/python3.6/site-packages/nikola/plugins/command/theme.py", line 173, in _execute
    return self.list_installed()
  File "/Users/niko/.virtualenvs/bikes.bembel.bytes-CZ5WGSwq/lib/python3.6/site-packages/nikola/plugins/command/theme.py", line 293, in list_installed
    themes += [(i, os.path.join(tdir, i)) for i in os.listdir(tdir)]
FileNotFoundError: [Errno 2] No such file or directory: 'themes'
```

This PR fixes this by only listing existing directories.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Features
 Bugfixes
 --------
 
+* Fix listing installed themes if theme directory is missing.
 * Watch correct output folder in ``nikola auto`` (Issue #3119)
 * Fix post fragment dependencies when posts are only available in a
   non-default language (Issue #3112)
--- a/nikola/plugins/command/theme.py
+++ b/nikola/plugins/command/theme.py
@@ -290,7 +290,9 @@ class CommandTheme(Command):
         themes = []
         themes_dirs = self.site.themes_dirs + [resource_filename('nikola', os.path.join('data', 'themes'))]
         for tdir in themes_dirs:
-            themes += [(i, os.path.join(tdir, i)) for i in os.listdir(tdir)]
+            if os.path.isdir(tdir):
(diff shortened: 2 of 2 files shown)


Kwpolska:
What if the `themes` directory is a symlink? I guess `os.path.exists` would work better.

(Unrelated, but [pipenv and Nikola don’t work together well.](https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/))

okin:
@Kwpolska `os.path.isdir` follows symlinks so this is not a problem. In case the link is broken this will return `False`.
I would like to stick with `isdir` because it will return `False` in case we unexpectedly hit a file whereas `exists` would accept this.

Issue #2606, opened 2016-12-25
Improving language support


Kwpolska:
Current situation:

* messages files do not have any support for plural numbers, meaning some things may look awkward (numbers don’t appear often in messages) — but note eg. Microsoft famously sucks at it
* dates may be formatted incorrectly (eg. in Polish: 25 grudnia vs 25 grudzień; the first form is grammatically correct) — this is something computers, including Windows, are expected to get right

We can do better than that, if we use the [Babel](http://babel.pocoo.org/en/latest/) library, which will:

* provide real localisation support (`.po`/`.mo` files, `messages_*.py` will have to go) with support for [plural forms](https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/Plural-forms.html)
* have date formats from CLDR, along with good date declension etc.

cc @felixfontein, @ralsina

felixfontein:
(An earlier discussion about dates including code samples can be found in #2601.)

sukiletxe:
In <https://gitlab.com/sukiletxe/sukiletxe.gitlab.io> I have created a plugin which uses Babel for date formatting, to overcome the fact that Alpine Linux doesn't use "real" locales (they are correctly detected and such, but then the dates are in English). You can adapt that code, if you wish.  Note: It **only** formats dates.

Kwpolska:
That’s a bug in Alpine Linux, please complain to them.

sukiletxe:
Done. To  be clear, I never said it was your bug.

ralsina:
IIRC the main problem was that using the traditional gettext toolchain in anything other than Linux was a pain. If someone wants to change to whatever, I don't mind.

Issue #3121, opened 2018-07-23
Issues on Arch Linux (PKGBUILD and UnicodeDecodeError)


odysseusPolytropos:
Hi folks!

I tried to install nikola (and nikola-git) on my up-to-date Arch Linux and ran into several problems:

### Environment

**Python Version:*3.6.6*

**Nikola Version:*7.8.15-1* and *8.0.0b1.r0.gc48c3861-1*

**Operating System:*Arch Linux*

### Description:
1. PKGBUILD does not seem to be valid, when trying to install it (with yaourt) it mocks about invalid characters '>=' and '=' in the optdepends section. A brief look into [the Arch wiki on PKGBUILD](https://wiki.archlinux.org/index.php/PKGBUILD#optdepends) suggests that there are no version restrictions allowed in optdepends, since it is a plain information format. Removing the >=[VERSION]-number suffixes did the trick for me. I'm not an expert on PKGBUILDs, but it looks to me like this should be fixed.

2. When trying to execute nikola, e.g.
`nikola init --demo nikolatest`
I get long stacktrace ending with
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 37: ordinal not in range(128)
Error loading mako template system plugin. (Uploading full error log does not work at the moment for some reasons, here is a [pastebin](https://pastebin.com/bbQ0G3Ce)).

I read this [blogpost on unix locales](https://chriswarrick.com/pl/blog/2017/06/18/unix-locales-vs-unicode/) but could not figure out what is wrong with my setup.
`[user@computer]$ locale`
> LANG=de_DE.utf8
LC_CTYPE="de_DE.utf8"
LC_NUMERIC="de_DE.utf8"
LC_TIME="de_DE.utf8"
LC_COLLATE="de_DE.utf8"
LC_MONETARY="de_DE.utf8"
LC_MESSAGES="de_DE.utf8"
LC_PAPER="de_DE.utf8"
LC_NAME="de_DE.utf8"
LC_ADDRESS="de_DE.utf8"
LC_TELEPHONE="de_DE.utf8"
LC_MEASUREMENT="de_DE.utf8"
LC_IDENTIFICATION="de_DE.utf8"
LC_ALL=

I also tried to set `LOCALE_DEFAULT = "en_US.UTF-8"` as well as `de_DE.UTF-8` in `/usr/lib/python3.6/site-packages/nikola/conf.py.in`, with no change in result.

[...]

PR #3126, opened 2018-07-30
Use Babel instead of system locale to format dates

Kwpolska:
* Fixes issues #2606 and #3121.
* Also fixes the locale situation for users who have incomplete locales or misconfigured systems.
* Supersedes the `babeldates` plugin for v7.
* Adds a few extra nifty features with date formatting (but not much else).
* Removes a lot of ugly and dead code.

I’m not a fan of how the threading support currently looks in LocaleBorg, but without it, everything goes awry.  Also, our translators may need to fix the date strings if (a) my guesses were wrong or (b) the CLDR data isn’t perfect in the context (cf. Polish translation of `Posts for {month_year}`.)


tests/README.rst
```diff
@@ -1,106 +0,0 @@
-.. title: The Nikola Test Suite
-.. slug: tests
-.. date: 2012/03/30 23:00
-
-The Nikola Test Suite
```
felixfontein: Is there a reason why you removed the tests README.rst completely?

felixfontein: Ok, I see: "Remove terribly outdated test documentation"
Makes sense :)

Kwpolska: The document is fairly outdated, and after removing the now-irrelevant parts, it wouldn’t really be useful for anything.

Kwpolska:
@felixfontein Babel supports basically all locales spoken by humans and some of the obscure ones, too. [Data from CLDR is used in many products](http://cldr.unicode.org/#TOC-Who-uses-CLDR-), and some customizations are doable in the locale itself. I don’t think supporting overrides is worth it, especially at the price of added complexity (what if the plugin needs thread-safety, like `locale` did?)

felixfontein:
@Kwpolska Unfortunately there are languages/dialects where CLDR is not sufficient, or where no official spelling exists and spelling can vary by individual persons. An example is Swiss German (locale gsw). I checked the CLDR month names, they are mostly OK, [except the one for October](https://unicode.org/cldr/charts/latest/summary/gsw.html#1344), which contains two "o"s, and I don't like that and don't want to have that in my blog :) Currently I'm using a custom date formatter for that (since Python doesn't know about gsw anyway), but if that is removed I'll have to rework my themes to stop using Nikola's date formatting facilities, and also have to fork the archives plugin to be able to control the archive names.

I understand the argument about complexity, but I would shove that burden to the plugin. If the plugin really needs thread safety, it has to do that by itself.

Issue #3129, opened 2018-08-02
Templating in charts - Feature


MortalCatalyst:
Regarding the chart directive https://www.getnikola.com/handbook.html#chart

Can we use templates with the charts?

So data could be stored as csv or json and not manually written?

ralsina:
This is **so** close to working with the shortcode :-)

If you put your data in a json file and put that in the "data" metadata field, you could turn this:

```
{{% chart Bar title='Browser usage evolution (in %)' %}}
        x_labels='["2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012"]'%}}
        'Firefox', [None, None, 0, 16.6, 25, 31]
        'Chrome',  [None, None, None, None, None, None]
        'IE',      [85.8, 84.6, 84.7, 74.5, 66, 58.6]
        'Others',  [14.2, 15.4, 15.3, 8.9, 9, 10.4]
{{% /chart %}}
```

Into something like this:

```
{{% chart Bar title='Browser usage evolution (in %)' %}}
        x_labels='["2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012"]'%}}
        % for k, v in post.data.items:
        '${k}', ${v}
        % endfor 
{{% /chart %}}
```

Or something similar, perhaps more complicated (not tried it yet)

OTOH, I think it makes a lot more sense to just have chart support a data_file option that lets you specify everything in a file. Would that be ok?

MortalCatalyst:
Yes having all options in a chart data file sounds good. 

Especially using it like you have it.
% for k, v in post.data.items():
        '${k}', ${v}

Makes it easy to adapt to a data flow.

PR #3130, opened 2018-08-02
Chart data files

ralsina:
Add a data_file option to chart shortcode and directive as sort-of-per #3129

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1256,13 +1256,14 @@ chart
 
     .. code:: text
 
-        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)' %}}
-        x_labels='["2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012"]'%}}
+        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)'
+x_labels='["2002","2003","2004","2005","2006","2007"]' %}}
         'Firefox', [None, None, 0, 16.6, 25, 31]
         'Chrome',  [None, None, None, None, None, None]
         'IE',      [85.8, 84.6, 84.7, 74.5, 66, 58.6]
         'Others',  [14.2, 15.4, 15.3, 8.9, 9, 10.4]
-        {{% /chart %}}{{% /raw %}}
+        {{% /chart %}}
+        {{% /raw %}}
 
 doc
     Will link to a document in the page, see `Doc role for details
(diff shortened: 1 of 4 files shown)


docs/manual.rst
```diff
@@ -1256,13 +1256,39 @@ chart
...
-        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)' %}}
-        x_labels='["2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012"]'%}}
+        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)'
+x_labels='["2002","2003","2004","2005","2006","2007"]' %}}
         'Firefox', [None, None, 0, 16.6, 25, 31]
         'Chrome',  [None, None, None, None, None, None]
         'IE',      [85.8, 84.6, 84.7, 74.5, 66, 58.6]
         'Others',  [14.2, 15.4, 15.3, 8.9, 9, 10.4]
-        {{% /chart %}}{{% /raw %}}
+        {{% /chart %}}
```
Kwpolska: they need to be on the same line

docs/manual.rst
```diff
@@ -1256,13 +1256,39 @@ chart
 
     .. code:: text
 
-        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)' %}}
-        x_labels='["2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012"]'%}}
+        {{% raw %}}{{% chart Bar title='Browser usage evolution (in %)'
+x_labels='["2002","2003","2004","2005","2006","2007"]' %}}
```
Kwpolska: Is this a syntax change or a fix to the docs?

nikola/plugins/shortcode/chart.py
```diff
@@ -50,20 +50,31 @@ def handler(self, chart_type, **_options):
...
+        style_name = options.pop('style', 'BlueStyle')
         if '(' in style_name:  # Parametric style
             style = eval('pygal.style.' + style_name)
         else:
             style = getattr(pygal.style, style_name)
-        for k, v in _options.items():
-            try:
-                options[k] = literal_eval(v)
+        for k, v in options.items():
+            try:                options[k] = literal_eval(v)
```
Kwpolska: why is this on one line?

docs/manual.rst
ralsina: Just a fix for the docs, things were different lengths

docs/manual.rst
ralsina: Sigh. Fixing.

docs/manual.rst
Kwpolska: You moved x_labels from “inside” the shortcode to its arguments.

ralsina: Hmmm that was broken, I think.

PR #3134, opened 2018-08-08
gallery: don't generate index page on empty output folder

evetsso:
Empty output folder allows galleries to be placed at the site's top directory,
instead of inside e.g. a "galleries" subdirectory.  But without this change, the
plugin would want to generate an index page that collides with the site's main
index page.

### Pull Request Checklist

- [X] I tested my changes.

### Description

I don't think this requires a documentation change?  Attempting to use an empty output folder for galleries I think just follows naturally without needing it explicitly mentioned?


Kwpolska:
Come to think of it, this breaks sites that are gallery-centric (who want the gallery index as main). I slightly amended it in commit 12205959.

PR #3136, opened 2018-08-09
Add support for yaml bundles definitions

gwax:
### Pull Request Checklist

- [x] I tested my changes.

### Description

This PR adds support for `bundles.yaml` or `bundles.yml` as alternatives to `bundles` files in themes.

If both are present, the yaml files take precedence. If only `bundles` is present, behavior is unchanged.

`bundles.yaml` takes the form:

```yaml
assets/css/all.css:
  - input1.css
  - input2.css
```

and would replace `bundles` file of the form:

```
assets/css/all.css=input1.css,input2.css
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,7 @@ Features
 
 * Don’t generate gallery index if the destination directory is
   site root and it would conflict with blog index (Issue #3133)
+* Theme bundles may now be defined using yaml as ``bundles.yaml``
 
 Bugfixes
 --------
--- a/docs/creating-a-theme.rst
+++ b/docs/creating-a-theme.rst
@@ -814,14 +814,26 @@ Bundles
 -------
 
 If you have ``webassets`` installed and the ``USE_BUNDLES`` option set to True, Nikola can put several CSS or JS files together in a larger file,
-which makes sites load faster. To do that, your theme needs a ``bundles`` file where the syntax is::
+which makes sites load faster. To do that, your theme needs a ``bundles.yaml`` file where the syntax is::
 
-    outputfile1.js=thing1.js,thing2.js,...
-    outputfile2.css=thing1.css,thing2.css,...
+    outputfile1.js:
+      - thing1.js
+      - thing2.js
+      - ...
+    outputfile2.css:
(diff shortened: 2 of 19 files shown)


gwax:
Baseline is failing because:

* existing code has a bug wherein `assets/js/all-nocdn.js` is specified as `assets/css/all-nocdn.js` in `base` template; this has been fixed
* input assets in `bundles.yaml` files have been sorted alphabetically

Kwpolska:
:-1: from me. What benefits does this have over the old format? I can’t think of any. Since we haven’t migrated config to YAML yet, there is no point in forcing it for simple files like `bundles`.

> existing code has a bug wherein `assets/js/all-nocdn.js` is specified as `assets/css/all-nocdn.js` in base template; this has been fixed

Good catch, I fixed this on master.

> input assets in `bundles.yaml` files have been sorted alphabetically

You can’t always do that with JS — for example, if `bootstrap.min.js` loads before `jquery.min.js`, it will refuse to work.

gwax:
Per your point, I have removed the sorting.

gwax:
The primary benefit over the existing format, and the reason I want this change, is that it allows for bundle inputs to be on separate lines.

During theme development, this helps me in two big ways:

1. Open `base_helper.tmpl` and `bundles` side-by-side to make sure they are in sync with each other
2. Diff two bundles against each other. This is useful both for parent/child comparison and git change comparison.

In both cases, having to read a long line of entries, especially where I have to visually parse between `_`, `,`, and `.` can be difficult. I often find myself temporarily inserting line breaks after each comma and then removing them before committing.

I initially considered modifying the existing format to support multiple lines per bundle but supporting yaml seemed much easier and more foolproof.

Kwpolska:
I don’t think using YAML and making it the new default is worth the effort, especially because (a) YAML is not our base dependency, (b) this is minor, “backstage” stuff. I could, however, accept a PR in which `bundles.json` becomes an alternate accepted format.

gwax:
Instead of changing the file format, I have opened #3137 to change the underlying parser to add additional features while maintaining strict backwards compatibility and without adding new base dependencies.

Issue #3124, opened 2018-07-26
Headings are not rendered correctly from rst


d-s-e:
### Environment:
**Python Version:** 3.6.6
**Nikola Version:** 8.0.0b2
**Operating System:** Arch Linux

### Description:
When I start a post with a heading (after all the meta tags), then this heading is ignored.
If I add some text before the heading - one non-whitespace character is enough - the heading is rendered in the output as it should be.

ralsina:
In reSt a single unique heading at the beginning of a document is special and becomes the document title.

Kwpolska:
We ignore that document title in Nikola, unless you explicitly enable that (`USE_REST_DOCINFO_METADATA`).

Kwpolska:
I can think of a better way to handle this.

PR #3158, opened 2018-09-06
Remove doctitle if not using reST docinfo metadata

Kwpolska:
This is #2382, #3124, and probably a lot of other issues.

New behavior: doctitle (first header) disappears only if `USE_REST_DOCINFO_METADATA` is true. I also re-implemented the feature in a less hacky way.

The old (but not well-known, even by the core devs) setting `NO_DOCUTILS_TITLE_TRANSFORM` was removed.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -21,6 +21,8 @@ Features
 Bugfixes
 --------
 
+* Do not remove first heading in document (reST document title)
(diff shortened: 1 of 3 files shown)


ralsina:
I like it. Even though it changes behaviour that has been there for a long time, that behaviour was really confusing anyway.

PR #3175, opened 2018-10-03
Use Markdown 3.0, fix some warnings

ralsina:
While Markdown 3.0 worked, it caused a lot of warnings because the extension API had changed. This PR makes Nikola work with 3.0 and ONLY with 3.0

Additionally, it fixes a python 3.7 warning and a pytest warning.


Kwpolska:
> ONLY with 3.0

Do you intend to add 2.x compatibility for this?

ralsina:
Wasn't planning on it. It would have to be done by adding version checks
and conditional code :-/

Kwpolska:
Tests pass, so I merged this. It’s certainly been a while this has been waiting without much reason to believe this should not be merged.

PR #3186, opened 2018-11-27
Using Twitter cards does not require an opt-in anymore.

okin:
### Pull Request Checklist

- [x] I tested my changes.

### Description

When the Twitter card support was implemented this was something that required an opt-in in their developer platform.
This is not the case anymore and hints about this may be misleading because people do not find a way to opt-in.


Kwpolska:
Good catch!

PR #3194, opened 2018-12-27
 Manual: Fix broken link for tagcloud

jonasstein:
fix broken link: changed v7 -> v8 in URL

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -7,7 +7,7 @@
 .. has_math: true
 .. author: The Nikola Team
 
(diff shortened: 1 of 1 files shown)


docs/manual.rst
```diff
@@ -7,7 +7,7 @@
 .. has_math: true
 .. author: The Nikola Team
 
-:Version: 8.0.1
+:Version: 8.0.2
```
Kwpolska: This is Nikola’s version (not the manual’s), and this field is updated automatically on release..

PR #3204, opened 2019-01-08
Fix several ResourceWarning: unclosed file

BoboTiG:
### Pull Request Checklist

- [ ] I tested my changes.

### Description

Fix several `ResourceWarning: unclosed file` and uniformize usage of the `with` context manager to prevent resource leaks.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,9 +10,6 @@ Bugfixes
 * Remove mention of Twitter cards requiring an opt-in.
   This is not true anymore - anyone can use them.
 * fancydates now workwith listdate items (eg. archives)
-* Several ``ResourceWarning: unclosed file``.
-  Also uniformize usage of the ``with`` context manager to
-  prevent resource leaks.
 
 Features
 --------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,9 @@ Bugfixes
 * Remove mention of Twitter cards requiring an opt-in.
   This is not true anymore - anyone can use them.
 * fancydates now workwith listdate items (eg. archives)
+* Several ``ResourceWarning: unclosed file``.
+  Also uniformize usage of the ``with`` context manager to
+  prevent resource leaks.
 
 Features
 --------
--- a/nikola/filters.py
+++ b/nikola/filters.py
@@ -146,19 +146,20 @@ def runinplace(command, infile):
 def yui_compressor(infile, executable=None):
     """Run YUI Compressor on a file."""
     yuicompressor = executable
-    if not yuicompressor:
-        try:
-            subprocess.call('yui-compressor', stdout=open(os.devnull, 'w'), stderr=open(os.devnull, 'w'))
-            yuicompressor = 'yui-compressor'
-        except Exception:
-            pass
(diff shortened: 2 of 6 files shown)


CHANGES.txt
```diff
@@ -10,6 +10,9 @@ Bugfixes
 * Remove mention of Twitter cards requiring an opt-in.
   This is not true anymore - anyone can use them.
 * fancydates now workwith listdate items (eg. archives)
+* Several ``ResourceWarning: unclosed file``.
```
Kwpolska: Internal changes should not be put in CHANGES.txt.

nikola/filters.py
```diff
@@ -146,19 +146,20 @@ def runinplace(command, infile):
...
-        except Exception:
-            pass
-    if not yuicompressor:
-        try:
-            subprocess.call('yuicompressor', stdout=open(os.devnull, 'w'), stderr=open(os.devnull, 'w'))
-            yuicompressor = 'yuicompressor'
-        except Exception:
-            raise Exception("yui-compressor is not installed.")
-            return False
+    with open(os.devnull, 'w') as devnull:
```
Kwpolska: Use [`subprocess.DEVNULL`](https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL) since we only support 3.4+ nowadays.

nikola/image_processing.py
```diff
@@ -188,8 +188,8 @@ def resize_svg(self, src, dst, max_size, bigger_panoramas):
                 op = gzip.GzipFile(dst, 'wb')
             else:
                 op = open(dst, 'wb')
-            op.write(lxml.etree.tostring(tree))
-            op.close()
+            with op:
```
Kwpolska: I’d rather not, `.close()` is more obvious.

setup.py
```diff
@@ -34,6 +34,9 @@
 standard_exclude_directories = ('.*', 'CVS', '_darcs', './build',
                                 './dist', 'EGG-INFO', '*.egg-info')
 
+with open('README.rst').read() as f:
```
Kwpolska: `.read()` shouldn’t be here

nikola/filters.py
BoboTiG: Cool, I was not sure I can use it.

setup.py
BoboTiG: Already fixed :+1:

nikola/image_processing.py
BoboTiG: OK, but if there is an error in `lxml.etree.tostring(tree)` the file would not be closed. But still, I revert this change.

nikola/plugins/command/serve.py
```diff
@@ -252,8 +252,8 @@ def send_head(self):
         filtered_bytes = None
         if ctype == 'text/html':
             # Comment out any <base> to allow local resolution of relative URLs.
-            data = f.read().decode('utf8')
-            f.close()
+            with f:
```
Kwpolska: This still does `with f`

Kwpolska:
Thanks!

BoboTiG:
My pleasure :)

PR #3211, opened 2019-01-22
Fix #3210 -- Correctly handle separators for "ignored_assets"

### LINKED ISSUE #3210: Cannot use "ignored_assets" properly on Windows with custom themes
### Environment

**Python Version:** Python 3.7.1

**Nikola Version:** Nikola v8.0.1

**Operating System:** Windows 10

### Description:

The option "ignored_assets" used in theme meta files does not work properly on Windows: the relative path is only recognized when using a backslash, i.e. css\theme.css instead of css/theme.css ([see doc](https://getnikola.com/theming.html#theme-meta-files)).

### Solution:

Normalize the path to the convention of the operating system when parsing the theme meta file.

de-souza:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Fix Issue https://github.com/getnikola/nikola/issues/3210 by normalizing the relative path to the convention of the operating system when parsing the theme meta file.

### DIFF
--- a/nikola/plugins/task/copy_assets.py
+++ b/nikola/plugins/task/copy_assets.py
@@ -67,7 +67,7 @@ class CopyAssets(Task):
         theme_ini = utils.parse_theme_meta(main_theme)
         if theme_ini:
             ignored_assets = theme_ini.get("Nikola", "ignored_assets", fallback='').split(',')
-            ignored_assets = [asset_name.strip() for asset_name in ignored_assets]
+            ignored_assets = [os.path.normpath(asset_name).strip() for asset_name in ignored_assets]
         else:
             ignored_assets = []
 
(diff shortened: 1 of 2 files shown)


nikola/plugins/task/copy_assets.py
```diff
@@ -67,7 +67,7 @@ def gen_tasks(self):
         theme_ini = utils.parse_theme_meta(main_theme)
         if theme_ini:
             ignored_assets = theme_ini.get("Nikola", "ignored_assets", fallback='').split(',')
-            ignored_assets = [asset_name.strip() for asset_name in ignored_assets]
+            ignored_assets = [os.path.normpath(asset_name).strip() for asset_name in ignored_assets]
```
Kwpolska: It might be better to call `.strip` before we call `normpath`.

PR #3218, opened 2019-02-05
Partially fix #3216

### LINKED ISSUE #3216: Incremental build seems not to be working
### Environment

**Python Version:**
Python 3.5.2

**Nikola Version:**
Nikola v8.0.1

**Operating System:**
Ubuntu 16.04

### Description:

If I run `nikola build` twice in a row, according to [this page](https://getnikola.com/features/fast.html) I would expect it to do nothing; and yet:
```
$ nikola build 2>&1 | wc -l
1910
```

It is less than a full rebuild:
```
$ nikola build -a 2>&1 | wc -l
4077
```
If I compare the actual output, it seems that the incremental rebuild doesn't do the `render_posts:cache/posts/...` tasks and the `render_taxonomies:output/...` for the `atom` and `xml` feeds.

Do I have some configuration or template in my site that makes Nikola misbehave? Are there some ways to debug this?

For the record, the source of my site is [here](https://gitlab.com/mardy/mardy.it).

ralsina:
Minor improvement to the JSON CustomEncoder so it handles defaultdicts better (also tuples and sets)


ralsina:
This probably breaks things for people with locales not supported by momentjs, I guess?

felixfontein:
Hmm, is it a good idea to sort lists and tuples?

ralsina:
@felixfontein on second thought I am guessing it is wrong, since it would mean changing the order of navigation links or somesuch will not trigger a rebuild :-(

CHANGES.txt
```diff
@@ -4,6 +4,7 @@ New in master
 Bugfixes
 --------
 
+* Improve the custom JSON encoder to better support defaultdicts, tuples and sets (Issue #3216)
```
Kwpolska: That doesn’t match the code.

nikola/nikola.py
```diff
@@ -171,8 +171,7 @@
         'sr_latin': 'sr_Latn',
     },
     'RTL_LANGUAGES': ('ar', 'fa', 'he', 'ur'),
-    'MOMENTJS_LOCALES': defaultdict(
-        str,
+    'MOMENTJS_LOCALES': dict(
```
Kwpolska: Let’s standardize on `{`.

Kwpolska:
Merging, although further debugging is necessary for #3216 on Python 3.5. (This fixes 3.6 and newer.)

PR #3233, opened 2019-02-19
bootstrap4, bootblog4: don't load moment.js when date_fanciness=0

### LINKED ISSUE #3231: [bootstrap4-jinja] [bootbase4-jinja] jinja themes should not load moment.js when DATE_FANCINESS = 0
### Environment

**Nikola Version:** 8.0.1

### Description:

While `base-jinja` correctly checks `date_fanciness` before loading `moment.js`, as seen in https://github.com/getnikola/nikola/blob/2414a5e53ad7c4188e330ff0276301a2500359d4/nikola/data/themes/base-jinja/templates/base_helper.tmpl#L69-L71
some other themes load this 319 kB JavaScript dependency regardless of the setting.

Offending themes include bootstap4-jinja and bootbase4-jinja, which each define their own `late_load_js`, e.g. https://github.com/getnikola/nikola/blob/2414a5e53ad7c4188e330ff0276301a2500359d4/nikola/data/themes/bootstrap4-jinja/templates/base_helper.tmpl#L64-L71

There may be others, I have not done an exhaustive search.

(There's probably a larger discussion to be had about whether it is _ever_ appropriate to load that 319 kB of javascript, given that the required locales seem to be known at compile-time, but that's for #3232.)

keturn:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Fixes #3231, prevents bootstrap-based themes from loading moment.js when fancy dates are not in use.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -23,6 +23,8 @@ Bugfixes
 * Remove mention of Twitter cards requiring an opt-in.
   This is not true anymore - anyone can use them.
 * fancydates now workwith listdate items (eg. archives)
+* bootstrap4 and bootblog4 themes no longer load moment.js when
+  fancydates are off. (Issue #3231)
 
 Features
 --------
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1632,6 +1632,7 @@ For Mako:
 
 .. code:: html
 
+    % if date_fanciness != 0:
     <!-- required scripts -- best handled with bundles -->
     <script src="/assets/js/moment-with-locales.min.js"></script>
     <script src="/assets/js/fancydates.js"></script>
@@ -1642,12 +1643,14 @@ For Mako:
     fancydates(${date_fanciness}, ${js_date_format});
     </script>
     <!-- end fancy dates code -->
+    %endif
 
 
 For Jinja2:
 
 .. code:: html
 
+    {% if date_fanciness != 0 %}
     <!-- required scripts -- best handled with bundles -->
     <script src="/assets/js/moment-with-locales.min.js"></script>
     <script src="/assets/js/fancydates.js"></script>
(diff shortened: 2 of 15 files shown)


keturn:
Is any automated test coverage desired for template changes? On the one hand, that conditional block is kind of gnarly, with bundles and CDN and date-fanciness all interacting. On the other hand, it may be hard to write readable tests for.

Unless you have end-to-end coverage (including javascript) that is testing fancy dates?

keturn:
There is one other case that's not yet handled by this PR:

When USE_CDN = False and USE_BUNDLES = True, `moment` is included in the bundle, regardless of fanciness.  (This is probably true of all themes, not just the bootstrap ones.)

Does the asset bundler have a way to make inclusion of specific assets conditional on settings?

keturn:
Looks like tests failed on account of changing the baseline site.

Kwpolska:
> When USE_CDN = False and USE_BUNDLES = True, `moment` is included in the bundle, regardless of fanciness. (This is probably true of all themes, not just the bootstrap ones.)

> Does the asset bundler have a way to make inclusion of specific assets conditional on settings?

No, it doesn’t, at least not now. An option could be added, but I don’t really think it’s worth it. If we really wanted to, we could move `moment` outside of bundles. Do you want to do that, or should I merge as-is?

> Looks like tests failed on account of changing the baseline site.

Correct, this will be taken care of when this is merged.

keturn:
Personally, I'm okay with one of the options that doesn't use the bundle (especially now that HTTP/2 seems to be widespread), but looking back at the default `conf.py` it looks like USE_CDN=False; USE_BUNDLES=True; DATE_FANCINESS=0 is the _default_ configuration, and including moment and not using it _by default_ does seem bad, yeah.

So the process for moving Moment outside of the bundles would be

- [ ] remove `moment-with-locales.min.js` from all `themes/*/bundles` files
- [ ] adjust all themes `late_load_js` blocks to move moment's `script` tag outside the `if not use_bundles` conditionals

is that right?

Kwpolska:
Yes, that’s how it would work.

keturn:
Okay, I believe that's all done now.

Kwpolska:
Thanks for contributing!

PR #3240, opened 2019-02-27
Missing Dependency

manuth:
### Pull Request Checklist

- [x] I tested my changes.

### Description
I have noticed that `flake8` is missing in the `requirements`-files. I've added the dependency to the `requirements-tests.txt`-file.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -38,6 +38,7 @@ Bugfixes
 * fancydates now workwith listdate items (eg. archives)
 * bootstrap4 and bootblog4 themes no longer load moment.js when
   fancydates are off. (Issue #3231)
+* Add missing `flake8`-dependency
 
 New in v8.0.1
 =============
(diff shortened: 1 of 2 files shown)


Kwpolska:
I wouldn’t call `flake8` a testing dependency, it’s not required to run Nikola’s test suite, the checker is completely separate.

Right now, I removed the CHANGES.txt notice for this and #3241, because (a) those were in the v8.0.2 section (they should be in a new `master` section), but also (b) those are internal, infrastructural changes, which are not relevant to Nikola’s users.

manuth:
This sure makes sense - I was just looking for a `requirements`-file which contains packages for developement, such as `pylint`, `PyInvoker` and stuff like this. `requirements-tests.txt` was the file which looked the most dev-ish.

I'm fine with putting dev-related requirements into a `requirements-dev.txt` or even completely dropping `flake8` from the requirements.

But I do think It'd make it a bit easier to follow the  [https://github.com/getnikola/nikola/blob/master/CONTRIBUTING.rst](Contributing-Guide).
What do you think?

Kwpolska:
I’d rather drop it, a fourth `requirements` file is just going to be clutter.

PR #3241, opened 2019-02-27
Exclude auto-generated files from flake8

manuth:
### Pull Request Checklist

- [x] I tested my changes.

### Description
All files located at `nikola/themes/base/data/messages` are generated automatically.
You may want to exclude it from `flake8`s style-check.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+exclude = nikola/data/themes/base/messages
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+exclude = nikola/data/themes/base/messages
```
ralsina: Could you put this in setup.cfg instead? One fewer file to remember we put configuration in :-)

manuth: Thanks for the advice - I just moved it to `setup.cfg`

ralsina:
Thanks for helping!

PR #3243, opened 2019-02-27
Add classes to `body` based on pagekind, permalink and lang

jean:
This allows styling on a per-pagekind, -page or -language basis,
without requiring template overrides.

For a Thai page in a Thai site, this results in body tags as follows:
`<body class="story_page page_page  2018 schedule  lang_th lang_ltr">`

For an English page in a Thai site:
`<body class="story_page page_page  en 2018 party  lang_en lang_ltr">`

We could filter out that "en", but I don't know if that's worthwhile.

### DIFF
--- a/nikola/data/themes/bootstrap4/templates/base.tmpl
+++ b/nikola/data/themes/bootstrap4/templates/base.tmpl
@@ -8,7 +8,7 @@ ${base.html_headstart()}
 </%block>
 ${template_hooks['extra_head']()}
 </head>
-<body>
+<body class="${' '.join(pagekind)} ${permalink.replace('/', ' ')} lang_${lang} lang_${'rtl' if is_rtl else 'ltr'}">
 <a href="#content" class="sr-only sr-only-focusable">${messages("Skip to main content")}</a>
 
 <!-- Menubar -->


jean:
If this is OK, I can update the other included themes with the change.

ralsina:
I don't mind. @Kwpolska ?

Kwpolska:
I don’t see the point in this.

* pagekind *might* make some sense, but for very few use-cases. I wouldn’t bother though.
* `permalink` makes no sense to me. There are also many chances for destructive CSS to be triggered in unexpected ways from unexpected places.
* `lang` and `rtl` are already available in `<html>`, and you can use that for styling or scripting.

jean:
Could I ask you to reconsider?

`lang`/`rtl` is clearly redundant, and I agree the current permalink split is too simplistic. How about something like `path-${permalink.replace('/', '-')`, with the language component stripped out? 

> There are also many chances for destructive CSS to be triggered in unexpected ways from unexpected places.

That is a risk, but adding CSS rules meant for a specific page without being able to scope them to that page risks triggering them in unrelated parts of the site. E.g. in our "_conference party_" page, image callouts should look different from our "_talks_" page. 

It's clearly useful, I've been using it from the beginning in Plone, and both Wordpress and Drupal also include the page or node id as a class on the `body`.

It's by far the easiest way to allow a designer to collaborate without needing to know anything about templates, and without having to copy templates into your theme.

Here are some examples:

- **Plone** (page id: `foundation`): `<body id="visual-portal-wrapper" class="template-summary_view portaltype-folder site-ploneorg section-foundation icons-off thumbs-on frontend viewpermission-view userrole-anonymous" [...]>` (https://plone.org/foundation)
- **Wordpress** (page id): `<body class="page-template-default page page-id-68 mp6 customizer-styles-applied group-blog highlander-enabled highlander-light">` (from the [editors-picks](https://discover.wordpress.com/editors-picks/) page)
  - **Wordpress site** linked from editor's picks (post id):`<body class="post-template-default single single-post postid-74283 single-format-standard custom-background wp-embed-responsive mp6 customizer-styles-applied user-background has-header-image highlander-enabled highlander-light custom-colors">` (from [tarotforthetimes](https://tarotforthetimes.org/2019/02/26/five-day-tarot-forecast-2-25-3-1/))
- **Drupal** (node id, node type): `<body class="html not-front not-logged-in no-sidebars page-node page-node- page-node-3021742 node-type-casestudy drupalorg-site-main new-style" >` (from https://www.drupal.org/case-study/the-wildlife-trusts)
  - **Drupal site** linked from case studies (path: `actions`): `<body class="not-front path-actions">` (from https://www.wildlifetrusts.org/actions)
- **Hugo site** (path: `how-it-works`): `<body class="is-boxed has-animations how-it-works list">` (from https://www.paceup.com/en/how-it-works/ , linked from from https://gohugo.io/showcase/).

Kwpolska:
Feel free to add this to your own templates, but I’m not really convinced. Are there really that many pages out there that make use of this?

jean:
- https://github.com/search?q=body+class suggests it's not an unusual need. 
- This [Wordpress tutorial](https://www.wpbeginner.com/wp-themes/wordpress-body-class-101-tips-and-tricks-for-theme-designers/) teaches it: 
  > Body Class (`body_class`) is a WordPress function that gives the body element different classes, so theme authors can style their sites effectively with CSS. [...] By adding this little element, you give yourself the flexibility of easily modifying the look of each page by using CSS.
  > [...]
  > About a year ago, we needed to style a WordPress menu item that was only on a specific page. We decided to dig deep to find a filter to add a special navigation class to the menu items which would be added using conditional statement. For example, if it is a single page, then add the class “Blog” to the menu items. After spending all the time in figuring out all that and writing a post about it, one of our users pointed out that **we could’ve easily done that with existing body class** [...]
- This [question](https://stackoverflow.com/q/37661815/274318) shows it being used: _Building a site in Drupal 8, using classy subtheme. Run into a puzzling theming issue - adding body class [...] **Themers use this to customize page display,** in my case using it to define a few sections of my site so I can change color and format._

I can continue collecting examples if you like .. 

From my perspective it's just by far the simplest way to tweak the looks of individual pages. In my case, for last year's PyCon Thailand we needed some help styling the schedule which was rendered as an RST table. I could ask a friend to have a look at the rendered HTML for that page, they wrote CSS that coloured table cells based on the type of timeslot, and I could be confident that no other tables would be affected. They didn't have to know that the site was based on Nikola, and I didn't have to touch any template.

Issue #2140, opened 2015-10-11
AttributeError: 'bool' object has no attribute 'format'


yamila-moreno:
Context: building a nikola with data imported from wordpress, python3.5 and version 7.7.2, I got the next error:
```
AttributeError: 'bool' object has no attribute 'format'
```
After debugging, the problem are the deprecation warnings. Once fixed, the build works properly.

Kwpolska:
Further information on Twitter; tl;dr: RSS_LINKS_APPEND_QUERY → FEED_LINKS_APPEND_QUERY does not create TranslatableSettings properly.

Kwpolska:
That was actually a copy-paste error. Fixed in master.

PR #3247, opened 2019-03-01
Fail meaningfully for RSS_LINKS_APPEND_QUERY

tedder:
### Pull Request Checklist

- [x] I tested my changes.

### Description
Valid values for `FEED_LINKS_APPEND_QUERY` (or `RSS_LINKS_APPEND_QUERY`) are `None`, `False`, or a string. It's sort of devious, because presumably if False is allowed, True is also allowed. Further, the error is more buried since the config string is abstracted and passed down the stack, making it difficult to understand what is actually failing.

Perhaps there's a reasonable default that could be used in the truthy-true-but-not-string case, but having a descriptive failure is a step in that direction.

Tested locally with false/true/string.

Added myself to authors since this is about my fourth PR, I guess.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1675,7 +1675,7 @@ class Nikola(object):
 
         feed_append_query = None
         if rss_links_append_query:
-            if isinstance(rss_links_append_query, bool):
+            if rss_links_append_query is True:
                 raise ValueError("RSS_LINKS_APPEND_QUERY (or FEED_LINKS_APPEND_QUERY) cannot be True. Valid values are False or a formattable string.")
             feed_append_query = rss_links_append_query.format(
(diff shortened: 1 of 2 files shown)


tedder:
Note, I thought this might be fixed by #2140. I'm linking it, though, as it's related.

nikola/nikola.py
```diff
@@ -1675,6 +1675,8 @@ def generic_rss_feed(self, lang, title, link, description, timeline,
 
         feed_append_query = None
         if rss_links_append_query:
+            if isinstance(rss_links_append_query, bool):
```
Kwpolska: A better, more obvious check would be `if rss_links_append_query is True:`

tedder: good call. I'll commit that.

PR #3249, opened 2019-03-19
Add Slideshare reST directive

roskakori:
This adds a slideshare reST directive to inline slides stored on https://www.slideshare.net.


Kwpolska:
This should be in the `plugins` repo, and ideally with a shortcode available for other input formats. Please move it there.

PR #3250, opened 2019-04-02
gallery was losing ordering

black-puppydog:
sorry for the quick and dirty PR.
It seems that the sorting was actually getting lost in older python versions. IIRC, python 3.6 introduced `dict`s that retain the insertion order, but it shouldn't really be relied on. I'm on python 3.5 (with nikola and gallery_directive both installed today), and the dict implementation was jumbeling up the sort that happens a few lines before as it seems.

I don't have time to do the checklist below here, but it's a trivial fix and I thought I'd open this, if only as a reminder for my future self.

### Pull Request Checklist

- [ ] I tested my changes.

### DIFF
--- a/nikola/plugins/task/galleries.py
+++ b/nikola/plugins/task/galleries.py
@@ -32,6 +32,7 @@ import io
 import json
 import mimetypes
 import os
+from collections import OrderedDict
 
 try:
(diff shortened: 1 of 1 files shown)


Kwpolska:
The fix looks okay, so I merged it. Thanks!

PR #3272, opened 2019-07-07
Fixing «Comments» link not translatable

cpkio:
### Description
Fixes a small bug where «Comments» link wasn't translatable no matter what language selected.

### DIFF
--- a/nikola/data/themes/base-jinja/templates/comments_helper_disqus.tmpl
+++ b/nikola/data/themes/base-jinja/templates/comments_helper_disqus.tmpl
@@ -30,7 +30,7 @@
 
 {% macro comment_link(link, identifier) %}
     {% if comment_system_id %}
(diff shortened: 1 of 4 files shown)


nikola/data/themes/base-jinja/templates/comments_helper_disqus.tmpl
```diff
@@ -30,7 +30,7 @@
 
 {% macro comment_link(link, identifier) %}
     {% if comment_system_id %}
-    <a href="{{ link }}#disqus_thread" data-disqus-identifier="{{ identifier }}">Comments</a>
+    <a href="{{ link }}#disqus_thread" data-disqus-identifier="{{ identifier }}">${messages(Comments)}</a>
```
Kwpolska: You're missing quotes around the string. Moreover, Jinja2 has different syntax to Mako.

Kwpolska:
Thanks!

PR #3276, opened 2019-07-29
"Use consistent html writer (plugin#294)

matclab:
Writer html5_polyglot is used in plugins/compile/rest/__init__.py whereas the
rst2hmtml is initialized with `writer_name='html'` which according to
docutils doc http://docutils.sourceforge.net/docs/user/html.html defaults
to `html4css1`. This discrepency prevent plugins adding role to work

Closes https://github.com/getnikola/plugins/issues/294

### Pull Request Checklist

- [x] I tested my changes.

### DIFF
--- a/nikola/plugins/compile/rest/__init__.py
+++ b/nikola/plugins/compile/rest/__init__.py
@@ -304,7 +304,7 @@ def add_node(node, visit_function=None, depart_function=None):
 def rst2html(source, source_path=None, source_class=docutils.io.StringInput,
              destination_path=None, reader=None,
              parser=None, parser_name='restructuredtext', writer=None,
-             writer_name='html', settings=None, settings_spec=None,
+             writer_name='html5_polyglot', settings=None, settings_spec=None,
              settings_overrides=None, config_section='nikola',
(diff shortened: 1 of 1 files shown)


ralsina:
LGTM, the test failures are the expected HTML changes.

matclab:
If the HTML output change is a problem, we can still replace all occurrences of `html5_polyglot` with `html4css1`.

Kwpolska:
The new writer does something really dumb: \`\`double backticks\`\` render as a `<span>` tag now instead of `<tt>`. Apparently, `<tt>` is deprecated in HTML 5. A better tag to output would be `<code>`. We should write a custom writer that does the right thing or go back to `html4css1`.

ralsina:
Is it a <span> with a reasonable class? In that case we can just add it to
one of the stylesheets.

Kwpolska:
It is styled (`.pre` is in the rst stylesheets) but I would prefer a dedicated tag to this.

ralsina:
I prefer to be html5-compliant if it doesn't look bad, but don't care all
that much one way or the other

Issue #3291, opened 2019-08-18
extra_footer block in base.tmpl


tdimitrov:
I'm building few custom pages for a Nikola powered site and I stumbled upon a minor issue. I wanted to override the footer of just one of the pages. As it is a single page CONTENT_FOOTER doesn't help, as it will update the content of all pages/posts.

My second idea was to add a small template override, similar to this one, but for the footer:

```
<%block name="extra_head">
    ${parent.extra_head()}
    <link rel="stylesheet" href="assets/bootstrap-social.css">
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
</%block>
```

I noticed that no such block exists in base.tmpl:

```
<div class="container" id="content" role="main">
    <div class="body-content">
        <!--Body content-->
        ${template_hooks['page_header']()}
        <%block name="content"></%block>
        <!--End of body content-->

        <footer id="footer">
            ${content_footer}
            ${template_hooks['page_footer']()}
        </footer>
    </div>
</div>
```

Do you think it's a good idea to add extra_footer block, which one can easily override for their needs?

Please check my PR, it's done for bootstrap4 theme.

tdimitrov:
Link to the PR: https://github.com/getnikola/nikola/pull/3292

Kwpolska:
#3292 was merged.

PR #3292, opened 2019-08-18
Issue #3291 - extra_footer block in base.tmpl

tdimitrov:
Add extra_footer block in bootstrap4/templates/base.tmpl

### Pull Request Checklist

- [x] I tested my changes.

### DIFF
--- a/nikola/data/themes/base/templates/base_footer.tmpl
+++ b/nikola/data/themes/base/templates/base_footer.tmpl
@@ -5,6 +5,7 @@
         <footer id="footer">
             <p>${content_footer}</p>
             ${template_hooks['page_footer']()}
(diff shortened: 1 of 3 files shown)


Kwpolska:
We don’t have this for other parts of the page. We might want to add an `extra_` block to other places. Opinions? @ralsina?

ralsina:
Don't see why not.

tdimitrov:
I've pushed another commit, which adds the block to the rest of the mako templates.

Kwpolska:
@gwax Jinja2 themes are auto-generated from the Mako themes.

Kwpolska:
I added `extra_header` in commit c589a1b32.

PR #3294, opened 2019-09-11
themes: add microformats tags for post authors

mardy:
Based on http://microformats.org/wiki/h-entry and
http://microformats.org/wiki/h-card

Signed-off-by: Alberto Mardegan <mardy@users.sourceforge.net>

### Pull Request Checklist

- [x] I tested my changes.

### Description

This adds microformat tags to the author data already present in post headers. I plan to add more data for each author, but that's not a trivial change so I'll prepare an issue describing a proposal.

### DIFF
--- a/nikola/data/themes/base-jinja/templates/post_header.tmpl
+++ b/nikola/data/themes/base-jinja/templates/post_header.tmpl
@@ -31,9 +31,9 @@
     <header>
         {{ html_title() }}
         <div class="metadata">
-            <p class="byline author vcard"><span class="byline-name fn" itemprop="author">
+            <p class="byline author vcard p-author h-card"><span class="byline-name fn p-name" itemprop="author">
                 {% if author_pages_generated %}
-                    <a href="{{ _link('author', post.author()) }}">{{ post.author()|e }}</a>
+                    <a class="u-url" href="{{ _link('author', post.author()) }}">{{ post.author()|e }}</a>
                 {% else %}
                     {{ post.author()|e }}
                 {% endif %}
--- a/nikola/data/themes/base/templates/post_header.tmpl
+++ b/nikola/data/themes/base/templates/post_header.tmpl
@@ -31,9 +31,9 @@
     <header>
         ${html_title()}
         <div class="metadata">
-            <p class="byline author vcard"><span class="byline-name fn" itemprop="author">
+            <p class="byline author vcard p-author h-card"><span class="byline-name fn p-name" itemprop="author">
                 % if author_pages_generated:
-                    <a href="${_link('author', post.author())}">${post.author()|h}</a>
+                    <a class="u-url" href="${_link('author', post.author())}">${post.author()|h}</a>
                 % else:
                     ${post.author()|h}
                 % endif


Kwpolska:
Do we really need more strange classes in our HTML? What tools actually use and benefit from those?

mardy:
> Do we really need more strange classes in our HTML? What tools actually use and benefit from those?

It's the recommended standard by the indieweb, and given that Nikola already supports parts of it (the `h-entry`) it seems normal to continue on that road. If you are concerned about the proliferation of classes, then maybe we can probably remove the `vcard` and `fn` classes, given that they are coming from an [obsoleted version of microformats](http://microformats.org/wiki/hcard), and `author`, which comes from the [hentry](http://microformats.org/wiki/hentry).

mardy:
For comparison, the in the `nikola/data/themes/base/templates/post.tmpl` both the current `h-entry` and the old `hentry` classes are supported.

Kwpolska:
It does, but the question is, do we really need that? Do we need any of these? Show me some tools that actually care about these (in whichever of the [15 competing standards](https://xkcd.com/927/)) and that have a significant enough userbase.

mardy:
> It does, but the question is, do we really need that? Do we need any of these?

Yes, if one wants to implement [webmentions](https://www.w3.org/TR/webmention/).

> Show me some tools that actually care about these (in whichever of the [15 competing standards](https://xkcd.com/927/)) and that have a significant enough userbase.

I honestly don't understand all this pushback on a trivial change that just improves web interoperability. If we go by the userbase logic, Nikola wouldn't have been written — everybody uses Facebook and Instagram, while a few geeks use Blogger and a handful of people use Wordpress :-)

The standard is there, but while even open-source projects push back on supporting it, it's just silly to expect to find a significant userbase for it. What I can tell you is that [support was added in Jekyll](https://github.com/jekyll/minima/pull/160) and that's why Jekyll has a [nice page in the IndieWeb wiki](https://indieweb.org/Jekyll) while Nikola is [not even listed](https://indieweb.org/static_site_generator). Improving microformats support is a very easy thing we can do with very little cost, and it can help us earn a strong support in the IndieWeb community; certainly it's not a "significant enough userbase", but it's a community with a rather high ratio of developers and otherwise technically skilled/aware bloggers: which seems to me the ideal community which would support Nikola's development.

Kwpolska:
> I honestly don't understand all this pushback on a trivial change that just improves web interoperability.

No change is trivial. Trivial changes add up, and they need maintenance. We are not really invested in microformats, and it shows, because

> maybe we can probably remove the vcard and fn classes, given that they are coming from an obsoleted version of microformats, and author, which comes from the hentry.

(and no, `author` does not necessarily come from `hentry`, it comes from logical class names for items you may want to style in CSS)

> What I can tell you is that support was added in Jekyll and that's why Jekyll has a nice page in the IndieWeb wiki while Nikola is not even listed.

It’s a wiki, which means you can add Nikola to the list. Moot point.

ralsina:
Merging. Let's err on the side of niceness and assume @mardy will maintain it since he's using it in his own site.

mardy:
I'll certainly will! Thanks, Roberto!

PR #3297, opened 2019-09-20
Create Furlan (Friulano) translation file

aoanla:
Professionally sourced translation into the Northern Italian language, Furlan (known popularly as Friulano).

### Pull Request Checklist

- [x] I tested my changes.


Kwpolska:
Translations are handled on Transifex: https://www.transifex.com/ralsina/nikola/
Please request the addition of Furlan there and upload your translations when the team is approved.

aoanla:
The thing about transifex, as you've noted, is that it adds a bunch more work to supply translations (and it isn't clear that the translation quality is better - for example, see the issue in the italian provided transifex translation, where they've ignored the instructions and translated "Read in English" literally, rather than into the equivalent of "Read in Italian", which is what's required. 

I suggest not using transifex in future, as quality is not guaranteed.

Kwpolska:
We aren’t using Transifex for quality reasons, we don’t source our translations from Transifex’s professional services. We’re doing that to open translations to non-programmers. Moreover, when we update the strings, translators get e-mail notifications and can update the translations, without needing to do the full flow with a GitHub pull request.

I can copy over these translations to Transifex myself if you don’t want to create an account there. And if you have any fixes for the Italian translation, feel free to fix them yourself or point them out (and provide a better translation).

Kwpolska:
I copied over the translation text to Transifex and added it to Nikola. Thanks for contributing!

PR #3301, opened 2019-09-25
cli improvements to add support for argparse, --output and --define flags

cav71:
Hi,
I have a pull request to add to Nikola cli few features:

add -o|--output flag to override the default OUTPUT_FOLDER config
use of argparser to parse arguments
add a -D|--define flag to override values in the conf.py
Thanks

### DIFF
--- a/nikola/__main__.py
+++ b/nikola/__main__.py
@@ -51,6 +51,9 @@ from .utils import (LOGGER, STDERR_HANDLER, STRICT_HANDLER,
                     ColorfulStderrHandler, get_root_dir, req_missing,
                     sys_decode)
 
+from . import cli
+
+
 try:
     import readline  # NOQA
 except ImportError:
@@ -65,8 +68,11 @@ config = {}
 _RETURN_DOITNIKOLA = False
 
 
-def main(args=None):
+
+def main(options=None):
     """Run Nikola."""
+    options, args = cli.parse_args(options) if isinstance(options, (list, tuple, None.__class__)) else options
+
     colorful = False
     if sys.stderr.isatty() and os.name != 'nt' and os.getenv('NIKOLA_MONO') is None and os.getenv('TERM') != 'dumb':
         colorful = True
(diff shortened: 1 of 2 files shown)


nikola/__main__.py
```diff
@@ -65,8 +68,11 @@
 _RETURN_DOITNIKOLA = False
 
 
-def main(args=None):
+
+def main(options=None):
     """Run Nikola."""
+    options, args = cli.parse_args(options) if isinstance(options, (list, tuple, None.__class__)) else options
```
Kwpolska: This is an ugly hack, please think of something else, perhaps you could just keep the old API?

nikola/cli.py
```diff
@@ -0,0 +1,30 @@
+import argparse
```
Kwpolska: Why is this a separate file? It can live in `__main__.py` IMO.

nikola/cli.py
```diff
@@ -0,0 +1,30 @@
...
+    parser = argparse.ArgumentParser()
+    add_arguments(parser)
+    options, args = parser.parse_known_args(args)
+    options = process_options(options) or options
+    return options, args
+
+
+def add_arguments(parser):
+    parser.add_argument('--conf', dest='conf_filename')
+    parser.add_argument('-D', '--define', dest='defines', action='append')
```
Kwpolska: Nikola’s settings are often more complex than just strings. This setting can be useful for some cases, although I don’t think we really need it, because of the limitations, and because it’s trivial to just edit `conf.py` and undo if the setting doesn’t work for you.

nikola/cli.py
```diff
@@ -0,0 +1,30 @@
...
+    parser.add_argument('-o', '--output')
```
Kwpolska: To change the output folder once doesn’t make much sense (just copy an existing `output` directory somewhere), to change it regularly, you probably want a different `conf.py` file anyway.

Kwpolska:
Overall, I’m :-1: on the changes made. We don’t really need to use argparse. The --output and --define flags would not have a lot of practical uses, since they can be easily replaced by `--conf`, which is also more flexible than using the new flags.

Any other opinions? @ralsina?

gwax:
@cav71 Could you describe your use case for wanting to override configuration values from the command line?

ralsina:
I say -1

Kwpolska:
Closing due to maintainer disapproval.

PR #3307, opened 2019-10-18
update pytest fixture

okin:
### Pull Request Checklist

- [x] I tested my changes.

### Description

The fixture used an old and deprecated decorator. It has been updated to current style.
The fixture has been changed so that in case an test fails it will return back to the original path anyway.


Kwpolska:
Thanks!

PR #3316, opened 2019-11-07
Fix demoting headers

okin:
### Pull Request Checklist

- [x] I tested my changes.

### Description

The tests for header demotion weren't running and after activating them they failed.
Tests have been converted to make use of pytests parametrization feature to avoid duplication of test code.

The demotion function has been correct so that all cases work as expected.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -1381,14 +1381,26 @@ def demote_headers(doc, level=1):
     if level == 0:
         return doc
     elif level > 0:
-        r = range(1, 7 - level)
+        levels = range(1, 7 - (level - 1))
+        levels = reversed(levels)
     elif level < 0:
-        r = range(1 + level, 7)
-    for i in reversed(r):
-        # html headers go to 6, so we can’t “lower” beneath five
-        elements = doc.xpath('//h' + str(i))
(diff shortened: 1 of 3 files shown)


nikola/utils.py
```diff
@@ -1381,14 +1381,26 @@ def demote_headers(doc, level=1):
...
+            # html headers can't go lower than 1
+            after = 1
+        if after > 6:
+            # html headers go until 6
+            after = 6
+
+        if before == after:
+            continue
+
+        elements = doc.xpath('//h%i' % before)
```
Kwpolska: We prefer modern string formatting, `doc.xpath('//h{}'.format(before))`

nikola/utils.py
```diff
@@ -1381,14 +1381,26 @@ def demote_headers(doc, level=1):
...
+        for element in elements:
+            element.tag = 'h%i' % after
```
Kwpolska: (Also here, `'h{}'.format(after)`)

okin:
I'm totally fine with using the newer str.format. Code has been updated!

Kwpolska:
Thanks for fixing this!

PR #3317, opened 2019-11-10
Make use of pytest for all tests

okin:
### Pull Request Checklist

- [x] I tested my changes.

### Description

As part of Hacktoberfest I started to look into the tests of Nikola and found that they...

* use a mixture of test styles (xUnit-style tests with unittest.TestCase; pytest functions)
* test more things than one thing in one test
* not all tests have been running
* had side-effects
* sometimes relied on other tests to be run first

I decided to port all the tests to pytest.
Pytest has been used as a test runner for some while now and provides more features that come in handy.
Pytest tests are written as simple functions which makes writing a test easier.
In combination with fixtures we are able to reduce the amount of setup code and we are able to effictively reduce side effects by using fixtures with [autouse](https://docs.pytest.org/en/latest/fixture.html#autouse-fixtures-xunit-setup-on-steroids).

This PR makes use of test parametrization and fixtures.
A lot of the tests have been refactored beyond a simple switch in test style to make them more explicit and easier to understand.


tests/test_plugin_importing.py
```diff
@@ -1,14 +1,2 @@
...
-class ImportPluginsTest(unittest.TestCase):
-    def test_importing_command_import_wordpress(self):
-        import nikola.plugins.command.import_wordpress  # NOQA
-
-    def test_importing_compile_rest(self):
-        import nikola.plugins.compile.rest  # NOQA
-
-    def test_importing_plugin_compile_markdown(self):
-        import nikola.plugins.compile.markdown    # NOQA
+def test_importing_plugin_task_galleries():
```
Kwpolska: This could be merged with the equally short `test_commands.py`.

tests/test_rst_compiler.py
```diff
@@ -1,238 +1,227 @@
...
+from nikola.plugins.compile.rest import vimeo
+from nikola.utils import _reload, LocaleBorg
+
+from .base import FakeSite, FakePost
+
+import pytest
+
+
+def test_ReST_extension(tempdir):
+    "Simple test for our base class :)"
```
Kwpolska: The docstring is probably outdated.  Also, we prefer `"""triple quotes"""` for docstrings.

tests/test_rst_compiler.py
```diff
@@ -1,238 +1,227 @@
...
-        _reload(nikola.plugins.compile.rest)
 
-    def test_vimeo(self):
-        """ Test Vimeo iframe tag generation """
-        self.basic_test()
-        self.assertHTMLContains("iframe",
-                                attributes={"src": ("https://player.vimeo.com/"
-                                                    "video/VID"),
-                                            "height": "400", "width": "600"})
+def getHtmlFromRst(temp_dir, rst):
```
Kwpolska: Please use `snake_case` here and in `assertHTMLContains`.

tests/test_rst_compiler.py
okin: There will be a few other places where single quotes are in place. I'll check for these!

Kwpolska: You could try to use the `black` automatic formatter on tests only. It’s a great tool overall; converting all of Nikola to `black` would make a huge mess, but since you’re editing practically all files under `tests/`, it will work out here nicely. Feel free to pick a length limit greater than the default (88).

tests/test_plugin_importing.py
okin: Merged them into `test_plugins.py`.

okin:
Work on this is still ongoing but nearing completion.

In the meantime I have split up the larger integration test module into separate modules. During this process I also added the tests that run the check commands to most integration tests as I couldn't think of why any of these tested settings should not be able to get a positive check result.
This did increase the runtime of the tests and if this is of concern the calls to the check commands could be removed from the tests.

There was one test for doing a wordpress import which I have converted to an automated tests aswell. This resides alongside the other integration tests.

I was also able to remove some appearantly old or unused parts of the test setups and now hopefully everything that is in the tests is actually required to run them.

My next step is working on formatting - probably with help of `black`.

tests/test_rst_compiler.py
```diff
@@ -1,238 +1,237 @@
...
-        self.basic_test()
-        self.assertHTMLContains("span", attributes={"class": "math"},
-                                text=r"\(e^{ix} = \cos x + i\sin x\)")
+    with io.open(infile, "w+", encoding="utf8") as post_file:
+        post_file.write(rst)
 
+    compiler = nikola.plugins.compile.rest.CompileRest()
+    compiler.set_site(FakeSite())
+    compiler.site.post_per_input_file[infile] = FakePost(outfile)
+    compiler.compile(infile, outfile)
```
Kwpolska: Couldn’t we just call `compiler.compile_string` and avoid dealing with temp files?

tests/test_command_import_wordpress.py
```diff
@@ -1,294 +1,116 @@
...
+    # assert not import_command.url_map
+    assert not module.links
+    import_command.url_map = {}
+
+    write_metadata = mock.MagicMock()
+    write_content = mock.MagicMock()
+    write_attachments_info = mock.MagicMock()
+    download_mock = mock.MagicMock()
+
+    with mock.patch(
```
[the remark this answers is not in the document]
okin: It works totally fine without.
Removed the nesting.

tests/test_rst_compiler.py
okin: We get one rather unimportant line less covered but have less setup required for tests.
I'd say this is a good tradeoff!

tests/integration/test_check_failure.py
```diff
@@ -0,0 +1,51 @@
+"""The demo build should pass 'nikola check' and fail with missing files."""
```
Kwpolska: Where is it tested that the demo site passes `nikola check`?

tests/integration/test_redirection_absolute.py
```diff
@@ -0,0 +1,49 @@
...
+from .test_empty_build import (  # NOQA
+    test_archive_exists,
+    test_avoid_double_slash_in_rss,
+    test_check_files,
+    test_check_links,
+    test_index_in_sitemap,
+)
+
+
+@pytest.fixture(scope="module")
```
Kwpolska: Please restructure this.

1. Merge all three `test_redirection_*.py` files into one (create all three redirects in one site)
2. Write a test to check that each of the three redirects works (file is created and has a meta refresh tag with the appropriate URL)

tests/integration/test_section_page_collision.py
```diff
@@ -0,0 +1,96 @@
...
+    assert "This is Post 0" not in page
+
+
+@pytest.fixture(scope="module")
+def build(target_dir):
+    """
+    Add subdirectories and create a post in section "sec1" and a page
+    with the same URL as the section index.
+
+    It also enables post sections.
```
Kwpolska: [Sections were a feature that was dropped in Nikola v8.](https://getnikola.com/blog/upgrading-to-nikola-v8.html#sections-were-replaced-by-categories) This test hasn’t been doing anything interesting for quite a while. That said, the layout created by sections can be recreated by `CATEGORY_DESTPATH_AS_DEFAULT = True; CATEGORY_PAGES_FOLLOW_DESTPATH = True`. It might be useful to activate that with `patch_config` in this test and rename it.

tests/integration/test_check_failure.py
okin: This can be found in `integration/test_demo_build.py`.

Kwpolska: It isn’t obvious without reading/being aware of `test_demo_build.py`.

tests/integration/test_building_in_subdir.py
```diff
@@ -0,0 +1,45 @@
...
+
+    add_post_without_text(os.path.join(target_dir, "posts"))
+
+    with cd(build_dir):
+        result = __main__.main(["build"])
+        assert result == 0
+
+
+@pytest.fixture(scope="module")
+def build_dir(target_dir):
```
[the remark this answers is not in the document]
okin: Nope, not anymore. If it will be needed in the future it can easily be re-introduced.

tests/integration/test_archive_full.py
```diff
@@ -0,0 +1,53 @@
...
+)
+def test_full_archive(build, output_dir, path):
+    """Check existance of archive pages"""
+    expected_path = os.path.join(output_dir, *path)
+    assert os.path.isfile(expected_path)
+
+
+@pytest.fixture(scope="module")
+def build(target_dir, test_dir):
+    """Fill the site with demo content and build it."""
```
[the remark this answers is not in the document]
okin: I moved the setup of the demo site into a function that is re-used through the tests that use it. What do you think?

tests/integration/test_check_failure.py
okin: I get your point and added some documentation. Let me know what you think about this!

tests/integration/test_archive_full.py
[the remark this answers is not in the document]
Kwpolska: It’s much better now!

tests/integration/test_building_in_subdir.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.fixture(scope="module")
+def build(target_dir):
+    """Fill the site with demo content and build it."""
+    prepare_demo_site(target_dir)
+
+    build_dir = os.path.join(target_dir, "posts")
+
+    with cd(build_dir):
+        result = __main__.main(["build"])
+        assert result == 0
```
Kwpolska: This `assert` does not appear anywhere else, why is it there?

tests/integration/test_check_failure.py
Kwpolska: That’s more obvious, thanks.

tests/integration/test_redirection_absolute.py
okin: I've merged all redirection tests into one file (`tests/integration/test_redirections.py`) and added the tests.
What do you say about it now?

Kwpolska: Much better.

tests/integration/test_building_in_subdir.py
okin: It's been like this for ages and I think it has been used to check for a valid build since the build is triggered in a different directory.

Since we are employing more tests now that check for valid builds (and fail if no proper build has been made) I'd say we remove it.
What do you think?

tests/integration/test_section_page_collision.py
okin: Nikola failed me with the various attempts I had a re-enabling the configuration and make the test do something useful.

Do you have any objections in dropping this test module alltogether?

tests/integration/test_building_in_subdir.py
Kwpolska: If a non-zero exit code will guarantee other tests in this file failing, and if that’s the case for all tests where we build something, then this `assert` can be dropped.

tests/integration/test_section_page_collision.py
Kwpolska:
```
$ tree posts
posts
├── 1.rst
├── a
│   ├── 1.rst
│   └── 2.rst
├── b
│   └── 1.rst
└── b.rst
```

conf.py changes

```python
CATEGORY_ALLOW_HIERARCHIES = True
CATEGORY_DESTPATH_AS_DEFAULT = True
CATEGORY_DESTPATH_TRIM_PREFIX = True
CATEGORY_PAGES_FOLLOW_DESTPATH = True
```

results:
* `/output/posts/a/index.html` is a category index (“Posts about a”) with two posts
* `/output/posts/b/index.html`… crashes with current master, so that needs fixing

I’ll take it from there, fixing both the code and the test, or dropping it if it’s impossible.

Kwpolska:
LGTM! Before merging, please resolve `tests/integration/test_building_in_subdir.py` (remove the assert if that’s safe). And if you wish to do any sort of rebasing/commit squashing/history simplification, now is the time. (It’s not necessary.)

PR #3320, opened 2019-11-25
Fix #3319 - drop extra_header/extra_footer from base theme

### LINKED ISSUE #3319: Named block 'extra_header' not allowed inside of def 'html_header'
Hi,

after updating to v8.0.3 I get this error:

```
(nikola) dzu@krikkit:~/blog (master)$ nikola build
Scanning posts....done!
[2019-11-24T11:24:07Z] ERROR: Nikola: Error loading tasks. An unhandled exception occurred.
[2019-11-24T11:24:07Z] ERROR: Nikola: mako.exceptions.CompileException: Named block 'extra_header' not allowed inside of def 'html_header' in file '/home/dzu/src/python/nikola/lib/python3.7/site-packages/nikola/data/themes/base/templates/base_header.tmpl' at line: 16 char: 5
[2019-11-24T11:24:07Z] NOTICE: Nikola: To see more details, run Nikola in debug mode (set environment variable NIKOLA_DEBUG=1) or use NIKOLA_SHOW_TRACEBACKS=1
(nikola) dzu@krikkit:~/blog (master)$ 
```

This is from the announcement mail of v8.0.3:

* Add ``extra_header`` and ``extra_footer`` blocks to templates
  (Issue #3291)

So I wonder if I have to do any additional things to complete the update or if this is a real bug.

Thanks in advance for any advice!

Kwpolska:
Mako does not allow blocks inside functions. The blocks have to be removed from these themes. They are still available in bootstrap4/bootblog4.

cc @laodzu @tdimitrov

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,9 @@ Features
 Bugfixes
 --------
 
+* Remove ``extra_header`` and ``extra_footer`` from ``base`` theme
+  due to incompatibility with Mako. The blocks are still available
+  in ``bootstrap4`` and ``bootblog4``. (Issue #3319 via #3291)
 * Show tracebacks when ``conf.py`` cannot be imported
 * Fix loading complex config files that import modules (Issue #3314)
 * Fix behavior of header demotion for negative values
--- a/nikola/data/themes/base-jinja/templates/base.tmpl
+++ b/nikola/data/themes/base-jinja/templates/base.tmpl
@@ -13,11 +13,11 @@
 <body>
     <a href="#content" class="sr-only sr-only-focusable">{{ messages("Skip to main content") }}</a>
     <div id="container">
-         {{ header.html_header() }}
-         <main id="content">
+        {{ header.html_header() }}
+        <main id="content">
             {% block content %}{% endblock %}
-         </main>
-         {{ footer.html_footer() }}
+        </main>
+        {{ footer.html_footer() }}
     </div>
     {{ base.late_load_js() }}
     {% if date_fanciness != 0 %}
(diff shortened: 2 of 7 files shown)


laodzu:
So with this fix my site builds again, but the output misses the navigation bar on the left hand side from the zen theme and is thus not usable. Moreover I cannot even build the clean site from nikola init with THEME = "zen".  Given the fact that my own site uses this theme and builds, irritates me.

In the meantime I wrote a small shell script to check all available themes on the "nikola init" site and the results with v8.0.3 show some problems.  Some of those are fixed in this branch, but not all.  So I will open a new bug for this.

Kwpolska:
@laodzu zen works for me with this PR in a demo site with the `conf.py` modified as the `zen` theme expects

laodzu:
Fixes #3319

laodzu:
For completeness - to fix the missing navigation bar on my own site I needed to update NAVIGATION_LINKS, notably the third arguments like this:

Old:
            ('/index.html', 'Home', 'icon-home'),

New:
            ('/index.html', 'Home', 'fa fa-home'),

I am currently going through the theme repo to find when this change was made neccessary, but somehow my setup continued working until now.  Maybe removing the cache would have shown the problem earlier, but this is speculation.

Kwpolska:
https://github.com/getnikola/nikola-themes/commit/a2450f228b7e6992dccaa376cef7c885d90304d3#diff-2b0faaad684d5ff37e9e8c039a038217

The change was made on 10 May 2018, and this was the only version available for v8. Caching is irrelevant, but installing a new version of `zen` would lead to the icons breaking.

laodzu:
I also understand this now.  Checking the git status of my site shows that experiencing the problems I tried to update the theme and this then made those changes neccesary.  I simply wanted to make sure that I am not reporting spurious problems that could be fixed on my side before raising the issue.

Kwpolska:
It worked for people who used bootstrap4/bootblog4, and it will still work for them.  Base-based themes didn’t work at all.

PR #3323, opened 2019-11-27
Fix possibly undefined exit function

okin:
### Pull Request Checklist

- [x] I tested my changes.

### Description

We have been using a function that might be undefined. This use is actually recommended against by Python - https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module

We use the function from sys instead that works the same.


Kwpolska:
Our users don’t really need to care about internal changes like this one. I’ll delete this from the changelog on master.

PR #3330, opened 2019-12-18
Refactored imports

okin:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Removed imports that we made to be compatible with now unsupported versions of Python (< 3.5).

Moved some builtin module imports to the top of the module - there is no point in having them as a lazy import as these are always present.

The largest part is a reworked import order:
First come Python builtins.
After that outside depenendencies followed by our own modules.
Lastly there are the conditional / optional imports.

Imports have been grouped together to make it easy to see where an import belongs.

The imports are ordered to first have complete module imports followed by specific imports.
The order is then alphabetically.

### DIFF
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -33,17 +33,9 @@ import re
 
 import lxml
 import piexif
+from PIL import ExifTags, Image
 
 from nikola import utils
-
-try:
-    from PIL import ExifTags, Image  # NOQA
-except ImportError:
-    try:
-        import ExifTags
-        import Image
-    except ImportError:
-        Image = None
 
 EXIF_TAG_NAMES = {}
 
@@ -91,9 +83,10 @@ class ImageProcessor(object):
 
     def resize_image(self, src, dst, max_size, bigger_panoramas=True, preserve_exif_data=False, exif_whitelist={}, preserve_icc_profiles=False):
         """Make a copy of the image in the requested size."""
-        if not Image or os.path.splitext(src)[1] in ['.svg', '.svgz']:
+        if os.path.splitext(src)[1] in ['.svg', '.svgz']:
             self.resize_svg(src, dst, max_size, bigger_panoramas)
             return
+
         im = Image.open(src)
 
         if hasattr(im, 'n_frames') and im.n_frames > 1:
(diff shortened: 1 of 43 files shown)


nikola/image_processing.py
```diff
@@ -27,25 +27,23 @@
...
+import lxml
 import piexif
 
 from nikola import utils
 
-Image = None
 try:
     from PIL import ExifTags, Image  # NOQA
 except ImportError:
     try:
```
Kwpolska: We can probably get rid of the nested try-except, `import Image` has been deprecated for years and is not supported by anything reasonably modern.

nikola/plugin_categories.py
```diff
@@ -38,7 +39,6 @@
 from .utils import LOGGER, first_line, get_logger, req_missing
 
 try:
```
Kwpolska: try-except is unnecessary, it was there to guard `import typing` for 3.4

nikola/plugins/command/import_wordpress.py
```diff
@@ -63,6 +49,16 @@
     get_default_translations_dict
 )
 
+try:
+    import html2text
+except ImportError:
+    html2text = None
```
Kwpolska: This probably needs `# NOQA`, and if it doesn’t, it would also not be needed for `phpserialize` and other instances of try-import-except-None.

nikola/plugins/task/galleries.py
```diff
@@ -33,31 +33,26 @@
...
 
 from nikola.plugin_categories import Task
 from nikola import utils
 from nikola.image_processing import ImageProcessor
 from nikola.post import Post
 
+try:
+    from PIL import Image  # NOQA
+except ImportError:
+    import Image
```
Kwpolska: `import Image` can be dropped here as well.

Kwpolska:
Please also resolve the merge conflict (caused by #3331).

nikola/image_processing.py
okin: If this is the case I'm dropping the nested import.

nikola/plugins/command/import_wordpress.py
okin: I just checked this:
* If the import is there to have something available in a module that isn't used (i.e. for backwards compatiblity) you need `# NOQA`. In this case there isn't usually try-except.
* If it is an optional import (handled with try-except) you can remove the `# NOQA`.

I'll be removing it for the latter.

PR #3331, opened 2019-12-18
There is no reason why the sitemap plugin shouldn't be a simple module.

okin:
### Pull Request Checklist

- [X] I tested my changes.

### Description

I didn't see a reason why the sitemap sitemap should be in it's own directory.
Therefore I changed it to be a simple module.

Please let me know if I'm missing something here that isn't obvious!


Kwpolska:
From git history, there used to be a reason in Nikola v5, there was a `sitemap_gen.py` third-party module there, but it was killed in #395 in March 2013.

Issue #3333, opened 2019-12-26
Reduce dependency footprint


ghost:
Nikola currently has 26 dependencies, including these:

~~~
Babel
blinker
docutils
doit
logbook
mako
Markdown
natsort
piexif
Pygments
PyRSS2Gen
python-dateutil
requests
setuptools
unidecode
Yapsy
~~~

and these transitive:

~~~
certifi (from requests)
chardet (from requests)
cloudpickle (from doit)
idna (from requests)
MarkupSafe (from mako)
pytz (from Babel)
six (from python-dateutil)
urllib3 (from requests)
~~~

and these native extensions:

~~~
Pillow
lxml
~~~

could some of these be removed? I know personally I dont have any use for the
Pillow imaging library and maybe some others.

ralsina:
Yes, you do need the Pillow extension if you want to have images in your
website.

cup:
@ralsina I dont think thats true. My understanding is Pillow would only be needed if you wish to do image manipulation. If you merely have images that are already processed, I cant imagine how Pillow would be needed.

ralsina:
Nikola processes images. Last I checked you can actually remove Pillow and
end with a less functional installation if you want

cup:
@ralsina something like this:

~~~diff
diff --git a/requirements-extras.txt b/requirements-extras.txt
index 62de7ce..db6f87d 100644
--- a/requirements-extras.txt
+++ b/requirements-extras.txt
@@ -15,2 +15,4 @@ ruamel.yaml>=0.15
 toml>=0.9.2

+Pillow>=2.4.0
+lxml>=3.3.5
diff --git a/requirements.txt b/requirements.txt
index 805f780..b8fd84d 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 doit>=0.30.1
 Pygments>=1.6
-Pillow>=2.4.0
 python-dateutil>=2.6.0
 docutils>=0.13
@@ -7,5 +6,4 @@ mako>=1.0.0
 Markdown>=3.0.0
 unidecode>=0.04.16
-lxml>=3.3.5
 Yapsy>=1.11.223
 PyRSS2Gen>=1.1
~~~

ralsina:
I would rather not. Pillow is needed for image galleries and image
thumbnails and both are core features.

Kwpolska:
lxml is mandatory for HTML processing. All pages we render go through that.

-- 
[...]

PR #3334, opened 2019-12-26
Drop logbook

Kwpolska:
Inspired by #3333, get rid of the `logbook` dependency. We don’t use any of its advanced features and with 70 SLOC, we can get rid of it.

Code that imports `logbook` might or might not work. Any such code has been eliminated from Nikola core, and from plugins (`drop-logbook` branch in `nikola-plugins`, PR getnikola/plugins#319).  Calling `logger.notice` will still work for the foreseeable future (it’s patched into our loggers).

### DIFF
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -139,7 +139,7 @@ class ImageProcessor(object):
             else:
                 im.save(dst, icc_profile=icc_profile)
         except Exception as e:
-            self.logger.warn("Can't process {0}, using original "
+            self.logger.warning("Can't process {0}, using original "
                              "image! ({1})".format(src, e))
             utils.copy_file(src, dst)
 
@@ -182,7 +182,7 @@ class ImageProcessor(object):
             op.write(lxml.etree.tostring(tree))
             op.close()
         except (KeyError, AttributeError) as e:
-            self.logger.warn("No width/height in %s. Original exception: %s" % (src, e))
+            self.logger.warning("No width/height in %s. Original exception: %s" % (src, e))
             utils.copy_file(src, dst)
 
     def image_date(self, src):
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -140,7 +140,7 @@ class ImageProcessor(object):
                 im.save(dst, icc_profile=icc_profile)
         except Exception as e:
             self.logger.warning("Can't process {0}, using original "
-                             "image! ({1})".format(src, e))
+                                "image! ({1})".format(src, e))
             utils.copy_file(src, dst)
 
     def resize_svg(self, src, dst, max_size, bigger_panoramas):
--- a/nikola/log.py
+++ b/nikola/log.py
@@ -78,16 +78,16 @@ class ColorfulFormatter(logging.Formatter):
 
 def get_logger(name: str, handlers=None) -> logging.Logger:
     """Get a logger with handlers attached."""
-    l = logging.getLogger(name)
+    logger = logging.getLogger(name)
     if handlers is not None:
         for h in handlers:
-            l.addHandler(h)
-    return l
+            logger.addHandler(h)
+    return patch_notice_level(logger)
 
 
 # For compatibility with old code written with Logbook in mind
 # TODO remove in v8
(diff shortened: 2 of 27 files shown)


nikola/image_processing.py
```diff
@@ -139,8 +139,8 @@ def resize_image(self, src, dst, max_size, bigger_panoramas=True, preserve_exif_
             else:
                 im.save(dst, icc_profile=icc_profile)
         except Exception as e:
-            self.logger.warn("Can't process {0}, using original "
-                             "image! ({1})".format(src, e))
+            self.logger.warning("Can't process {0}, using original "
+                                "image! ({1})".format(src, e))
```
felixfontein: How about monkeypatching `notice` and `warn` methods to the loggers returned by `utils.get_logger`, so that 3rd party plugins using these don't simply break? Or should this PR only go into a new major version?

Kwpolska: `notice` is already monkeypatched. `warn` is supported in `logging` but deprecated.

felixfontein: Ah, I overlooked it. Thanks!

nikola/utils.py
```diff
@@ -113,57 +110,8 @@
...
-
-
-def get_logger(name, handlers=None):
-    """Get a logger with handlers attached."""
-    l = logbook.Logger(name)
-    l.handlers += STDERR_HANDLER
-    return l
-
-
-STDERR_HANDLER = [ColorfulStderrHandler(
```
felixfontein: Almost all plugins I'm using now fail with
```
AttributeError: module 'nikola.utils' has no attribute 'STDERR_HANDLER'
```
when doing `utils.get_logger('xxx', utils.STDERR_HANDLER)`. Also grepping in https://github.com/getnikola/plugins/ shows tons of plugins using `STDERR_HANDLER`. I guess we should still provide a dummy `STDERR_HANDLER` so that existing code won't break.

I've tried adding
```
STDERR_HANDLER = None
```
somewhere in this file, and with that, all my blogs and pages run through fine.

Kwpolska: Fixed in commit 5cd3e140a.

nikola/log.py
```diff
@@ -0,0 +1,151 @@
...
+    """Stream handler with colors."""
+
+    _colorful = False
+
+    def format(self, record: logging.LogRecord) -> str:
+        """Format a message and add colors to it."""
+        message = super().format(record)
+        return self.get_wrap_string(record).format(message)
+
+    def get_wrap_string(self, record: logging.LogRecord) -> str:
```
okin: Why not call this something very obvious like `wrap_in_color`?

nikola/log.py
```diff
@@ -0,0 +1,151 @@
...
+    handler = logging.StreamHandler()
+    handler.setFormatter(
+        ColorfulFormatter(
+            fmt="[%(asctime)s] %(levelname)s: %(name)s: %(message)s",
+            datefmt="%Y-%m-%dT%H:%M:%SZ",
+        )
+    )
+    if logging_mode == LoggingMode.QUIET:
+        logging.root.handlers = []
+        return
```
okin: Since we aren't setting any handlers this probably could live above the configuration of `handler`.

nikola/plugins/compile/rest/__init__.py
```diff
@@ -26,6 +26,7 @@
 
 """reStructuredText compiler for Nikola."""
 
+import logging
```
okin: I'd move this one line down for the imports to be ordered but that might just be me :woman_shrugging: :smile:

nikola/log.py
```diff
@@ -0,0 +1,152 @@
...
+        handlers.append(StrictModeExceptionHandler())
+
+    logging.root.handlers = handlers
+
+
+configure_logging()
+
+
+# For compatibility with old code written with Logbook in mind
+# TODO remove in v8
```
okin: This should be removed in v9 I assume.

Kwpolska: Correct, fixed.

Kwpolska:
Thanks for the reviews!

PR #3335, opened 2019-12-30
Fix category_destpath conflicts with posts and add tests for those

Kwpolska:
As discovered in #3317, there were leftover tests for the sections feature, that was removed with Nikola v8. In an attempt to use the replacement CATEGORY_DESTPATH features, it turned out that the old sections would avoid conflicts when the section index path (now category index path) was also the destination path of a post. This PR restores the behavior and creates a new test for this feature.


tests/integration/test_category_destpath.py
```diff
@@ -0,0 +1,83 @@
+"""Test if section indexes avoid pages."""
```
okin: I think the docstring should be rewritten since it does not test the obsolete _sections_ feature.

tests/integration/test_category_destpath.py
```diff
@@ -0,0 +1,83 @@
...
+    assert os.path.isfile(_make_output_path(cat1, "index"))
+    assert os.path.isfile(_make_output_path(cat2, "index"))
+
+    # Are their contents correct?
+    with open(_make_output_path(cat1, "index"), "r", encoding="utf-8") as fh:
+        page = fh.read()
+        assert "Posts about cat1" in page
+        assert "test-destpath-p1" in page
+        assert "test-destpath-p2" in page
+        assert "test-destpath-p3" not in page
```
okin: There is no need for the file to still be open when these assertions are made. I'd suggest lowering their indentation.

tests/integration/test_category_destpath.py
```diff
@@ -0,0 +1,83 @@
...
+
+    with open(_make_output_path(cat2, "index"), "r", encoding="utf-8") as fh:
+        page = fh.read()
+        assert "Posts about cat2" not in page
+        assert "This is a post that conflicts with cat2." in page
```
okin: There is no need for the file to still be open when these assertions are made. I'd suggest lowering their indentation.

tests/integration/test_category_destpath.py
```diff
@@ -0,0 +1,83 @@
...
+
+    cat1 = os.path.join(output_dir, "posts", "cat1")
+    cat2 = os.path.join(output_dir, "posts", "cat2")
+
+    # Do all files exist?
+    assert os.path.isfile(_make_output_path(cat1, "index"))
+    assert os.path.isfile(_make_output_path(cat2, "index"))
+
+    # Are their contents correct?
+    with open(_make_output_path(cat1, "index"), "r", encoding="utf-8") as fh:
```
okin: Instead of re-creating the same index again I'd suggest storing the path to the cat1 index in a variable and refer to it here and in the file existance test above.

tests/integration/helper.py
```diff
@@ -4,36 +4,46 @@
 
 from ..helper import cd
 
-__all__ = ["add_post_without_text", "append_config", "cd", "patch_config"]
+__all__ = ["add_post_without_text", "create_simple_post", "append_config", "cd", "patch_config"]
```
okin: I think for overall readability it is the easiest to have these ordered alphabetically.

Issue #3327, opened 2019-12-18
Split up nikola.utils


okin:
The module `nikola.utils` does a lot of things and some more namespaces would be helpful.

Among the thing that are to be found there:
* Rendering utilities
* Command base class
* Time handling utlities
* Zip archive handling
* Asset handling
* Translation
* RSS feed utlities
* CLI interaction
* ... and even more!

This is a lot of things for one module.

I'd suggest splitting the module up into smaller pieces by adding new submodules to `nikola.utils` that group functionality.

PR #3336, opened 2019-12-31
Remove indirection for accessing str / bytes.

okin:
# Pull Request Checklist

- [x] I tested my changes.

### Description

With Nikola being Python 3 only this isn't necessary anymore.

This also makes attempting #3327 easier because it reduces some smaller dependency to `nikola.utils`.

### DIFF
--- a/nikola/utils.py
+++ b/nikola/utils.py
@@ -86,8 +86,8 @@ __all__ = ('CustomEncoder', 'get_theme_path', 'get_theme_path_real',
            'get_theme_chain', 'load_messages', 'copy_tree', 'copy_file',
            'slugify', 'unslugify', 'to_datetime', 'apply_filters',
            'config_changed', 'get_crumbs', 'get_tzname', 'get_asset_path',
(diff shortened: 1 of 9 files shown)


nikola/utils.py
```diff
@@ -104,11 +104,6 @@
 # Are you looking for 'generic_rss_renderer'?
 # It's defined in nikola.nikola.Nikola (the site object).
 
-# Aliases, previously for Python 2/3 compatibility.
```
Kwpolska: I don’t think we can remove this, plugins could still use those.

okin: I checked for those in getnikola/plugins.
Maybe this is a change that should be held off until a larger release.

Kwpolska: We can change everything now, add a standard `# TODO: remove in v9` deprecation notice to the codebase, and that will be taken care of in the future.

Kwpolska: I’ll handle this from here.

PR #3343, opened 2020-01-30
Pass correct parameters to shortcodes in jupyter notebooks

mstimberg:
The wrong call passed a boolean as the filename and the language name instead of the `post` object.

I took the liberty to add my name to the authors list – the change is trivial, but it took me a while to figure out why I was getting these weird values in the shortcodes :)

### Pull Request Checklist

- [x] I tested my changes.


Kwpolska:
Thanks for fixing this!

PR #3356, opened 2020-03-01
Fix simple typo: resutling -> resulting

timgates42:
There is a small typo in nikola/plugins/command/import_wordpress.py.
Should read `resulting` rather than `resutling`.


felixfontein:
@timgates42 thanks a lot for fixing this! :)

PR #3358, opened 2020-03-18
Make code listings prettier and use word wrap

Kwpolska:
Nikola’s code listings are pretty ugly right now.

ugly code

The most prominent issues are the fact that the line numbers get an ugly white background, and at the bottom, there are scrollbars. Well, there are the scrollbar placeholders, because if they weren’t there, we’d break alignment with line numbers if the code was too long. And if code is too long, you get a horizontal scrollbar, the UX of which is hit-and-miss and largely depends on your pointing device.

In an intention to improve this, and also fix supakeen/pinnwand#4, I built a BetterHtmlFormatter, which lets us fix these issues. The line numbers adapt in style to the rest of the code block, and if a line is too long, browser-based word wrap applies.

image

I tested this on Safari, Firefox, and Chrome, but extra testing is always welcome.

I vendored this package in, because I believe this will be the easiest for users and downstream distributors.


Kwpolska:
Thanks for reviewing!

PR #3368, opened 2020-04-04
Fix #3361

### LINKED ISSUE #3361: Dependency issue in gallery building with -n4
Building my site, if I use multiprocessing, I get errors about specific images "not existing".

It seems to me like the page tries to get the image size but the image is not processed yet. Probably a missing dependency.

ralsina:
Building the gallery indexes requires reading sizes of **destination** images, so they should also depend on destination image files.

### DIFF
--- a/nikola/plugins/task/galleries.py
+++ b/nikola/plugins/task/galleries.py
@@ -308,7 +308,7 @@ class Galleries(Task, ImageProcessor):
                 yield utils.apply_filters({
                     'basename': self.name,
                     'name': dst,
-                    'file_dep': file_dep,
+                    'file_dep': file_dep + file_dep_dest,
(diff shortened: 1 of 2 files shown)


nikola/plugins/task/galleries.py
```diff
@@ -308,7 +308,7 @@ def gen_tasks(self):
                 yield utils.apply_filters({
                     'basename': self.name,
                     'name': dst,
-                    'file_dep': file_dep,
+                    'file_dep': file_dep + file_dep_dest,
```
Kwpolska, line 311: It should be enough to just do `+ dest_img_list` here.

ralsina: Yeah, good point

PR #3370, opened 2020-04-04
Fix #3326 added support for default preview images

### LINKED ISSUE #3326: Add support for default previewimage
The previewimage meta tag is used for Open Graph metadata, among other things. However, not all pages and blog posts will have their own image. It would be cool to be able to provide a generic **default image via the config**, without having to add it manually to all pages that don't feature one already.

(New pages and posts can be generated with the matching previewimage line, of course, but retrofitting all old files is a pain.)

ralsina:
It's pretty basic but it works.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -292,14 +292,14 @@ slug
     other symbols may cause issues in URLs. (required)
 
     So, if the slug is "the-slug" the page generated would be "the-slug.html" or
-    "the-slug/index.html" (if you have the pretty URLs option enabled) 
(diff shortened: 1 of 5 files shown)


docs/manual.rst
```diff
@@ -382,6 +382,9 @@ previewimage
 
        .. previewimage: /images/looks_great_on_facebook.png
 
+    If a post has no `previewimage` it will try ti use the `DEFAULT_PREVIEW_IMAGE`
```
Kwpolska, line 385: try to use*

PR #3371, opened 2020-04-04
Make gallery images lazy loading

ralsina:
Minor fix, useful for very large galleries.


ralsina:
@Kwpolska yes, it's pretty green

image

Issue #1771, opened 2015-05-31
Use thumbnails in gallery index


humitos:
Gallery index is showing just a list of folder's name for each gallery. I don't like this, so I changed the behavior to build something better. An example here:

http://argentinaenpython.com.ar/galeria/

In the example you can see that I'm showing the title, preview image and a description. I'm using `.. previewimage:`, `.. description:` and `.. title:` from the `index.txt` file inside the gallery folder.

Here is part of the code that I'm building to make this work as in the example, but I need some help from you guys to make it perfect for a PR.

``` diff
diff --git a/nikola/conf.py.in b/nikola/conf.py.in
index d6776fd..54438a1 100644
--- a/nikola/conf.py.in
+++ b/nikola/conf.py.in
@@ -645,6 +645,19 @@ COMMENT_SYSTEM_ID = ${COMMENT_SYSTEM_ID}
 # Enable comments on picture gallery pages?
 # COMMENTS_IN_GALLERIES = False

+# Use a thumbnail (defined by ".. previewimage:") in list of galleries
+# for each gallery
+# GALLERIES_USE_THUMBNAIL = False
+
+# Columns of thumbnails list in galleries index
+# GALLERIES_COLUMNS = 3
+
+# Image to use as thumbnail for those galleries that doesn't have one
+# GALLERIES_DEFAULT_THUMBNAIL = ""
+
+# HTML title for /GALLERY_FOLDERS/index.html
+# GALLERY_INDEX_TITLE = "galleries"
+
 # What file should be used for directory indexes?
 # Defaults to index.html
 # Common other alternatives: default.html for IIS, index.php
diff --git a/nikola/data/themes/bootstrap3/templates/gallery.tmpl b/nikola/data/themes/bootstrap3/templates/gallery.tmpl
index 3d6c01a..5ae23e7 100644
--- a/nikola/data/themes/bootstrap3/templates/gallery.tmpl
+++ b/nikola/data/themes/bootstrap3/templates/gallery.tmpl
@@ -15,11 +15,40 @@
     </p>
     %endif
     %if folders:
-    <ul>
-    % for folder, ftitle in folders:
-        <li><a href="${folder}"><i class="glyphicon glyphicon-folder-open"></i>&nbsp;${ftitle}</a></li>
+    % if GALLERIES_USE_THUMBNAIL:
+    % for folder, ftitle, fpost in folders:
+   % if loop.index % GALLERIES_COLUMNS == 0:
+   <div class="row">
+   % endif
[...]

PR #3372, opened 2020-04-04
Implement #1771

ralsina:
Note:

* This is a **very** basic implementation of "thumbnails in gallery list" issue #1771
* Most of the code is shamelessly copied from @humitos PR
* In action: http://ralsina.me/galleries/index.html
* This requires #3326 because it includes its changes.

I don't want this to be perfect, I just want:

1. That it doesn't break existing sites
2. That it sort of works?


nikola/data/themes/base-jinja/templates/gallery.tmpl
```diff
@@ -15,12 +15,26 @@
...
-        <li><a href="{{ folder }}"><i
-        class="icon-folder-open"></i>&nbsp;{{ ftitle|e }}</a></li>
-    {% endfor %}
-    </ul>
+        {% if galleries_use_thumbnail %}
+            {% for (folder, ftitle, fpost) in folders %}
+            <div class="thumnbnail-container">
+                <a href="{{ folder }}" class="thumbnail image-reference" title="{{ ftitle|e }}">\
+                    {% if fpost and fpost.previewimage %}
+                        <img src="{{ fpost.previewimage }}" alt="{{ ftitle|e }}" loading="lazy" style="max-width:{{ thumbnail_size }}px; max-height:{{ thumbnail_size }}px;" />
```
Kwpolska, line 23: This could lead to a path issue. If the user provides an image filename, the previewimage would be valid inside the gallery, but not outside. There is no documentation that states what is required. At the same time, it might be a good idea to include the previewimage as og:image and other open graph metadata in gallery pages.

ralsina: I'll add a mention in the docs about that.

And yeah, it would be nice to have that in og:image, I'll add it in a followup PR

ralsina: Actually, the docs for previewimage metadata say "image path relative to BASE_URL"

Kwpolska: `open_graph_metadata` uses this to handle all cases:

```mako
<meta property="og:image" content="${url_replacer(permalink, post.previewimage, lang, 'absolute')}">
```

ralsina: Not really sure that would work here since the permalink is not the post's link but the parent folder's?

Kwpolska: Should work if we pass `folder`, or `permalink+"/"+folder`.

ralsina: Ok, I'll take a shot today during lunch

nikola/data/themes/base-jinja/templates/gallery.tmpl
```diff
@@ -15,12 +15,26 @@
...
-    <ul>
-    {% for folder, ftitle in folders %}
-        <li><a href="{{ folder }}"><i
-        class="icon-folder-open"></i>&nbsp;{{ ftitle|e }}</a></li>
-    {% endfor %}
-    </ul>
+        {% if galleries_use_thumbnail %}
+            {% for (folder, ftitle, fpost) in folders %}
+            <div class="thumnbnail-container">
+                <a href="{{ folder }}" class="thumbnail image-reference" title="{{ ftitle|e }}">\
```
Kwpolska, line 21: This backslash is invalid in Jinja2, and shouldn’t really be needed. I removed it.

ralsina: Thx!

nikola/data/themes/base-jinja/templates/gallery.tmpl
ralsina: Not doing this because I am not sure how :-(

So, basically: previewimage is meant to be "image path relative to BASE_URL" until further notice.

Issue #3338, opened 2020-01-06
Multiple image thumbnail sizes


gpaumier:
Hello,

As a theme designer, I would like to be able to configure multiple image thumbnail sizes instead of only one in `conf.py`. This is needed in order to have [responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) through the use of [`srcset` and `sizes`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset), which are now [widely supported by browsers](https://caniuse.com/#search=srcset).

This allows to make sure that the browser loads an image appropriate for the viewport size and pixel density (e.g. [retina displays](https://en.wikipedia.org/wiki/Retina_display)) of the user, which can vary widely across devices.

Ideally the size would be part of the new filename, e.g. `image-400w.jpg`, `image-1200w.jpg`, `image-2400w.jpg` for configured sizes of 400, 1200 and 2400 pixels.

ralsina:
It should be "simple" to create a plugin to do this. I'll take a shot now.

ralsina:
My idea of how to implement it so far:

New setting (example):

SRCSET_IMAGE_SIZES=(400,1200,2400)

Then a task plugin that would take images from images/ (or whatever, same folder we support now) and create the images with the right names/locations.

Then a srcset shortcode that would create the right HTML thing.

ralsina:
On second though, I don't have the energy to do this right now.

ralsina:
I modified our image scaling code so that it can take multiple sizes at once.

Next step would be to add a config option to replace/enhance  our current thumb size ones defining multiple thumb sizes.

PR #3376, opened 2020-04-10
Refactor resize_image to support multiple target sizes

ralsina:
This could eventually lead to supporting things like #3338

In the meantime, gallery rendering in my site takes ~~112~~ 260 seconds when it takes 364 using master.

Still there are opportunities for improving performance:

* EXIF cleanup could be done only once per loop
* Probably something else

But this is not so invasive and gives a large perf improvement, so let's merge it if it's ok.

### DIFF
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -112,6 +112,7 @@ class ImageProcessor(object):
         # The jpg exclusion is Issue #3332
         is_animated = hasattr(_im, 'n_frames') and _im.n_frames > 1 and extension not in {'.jpg', '.jpeg'}
 
+        exif = None
         if "exif" in _im.info:
             exif = piexif.load(_im.info["exif"])
             # Rotate according to EXIF
--- a/nikola/image_processing.py
+++ b/nikola/image_processing.py
@@ -103,23 +103,19 @@ class ImageProcessor(object):
             max_sizes = [max_size]
         if len(max_sizes) != len(dst_paths):
             raise ValueError('resize_image called with incompatible arguments: {} / {}'.format(dst_paths, max_sizes))
+        extension = os.path.splitext(src)[1].lower()
+        if extension in {'.svg', '.svgz'}:
+            self.resize_svg(src, dst_paths, max_sizes, bigger_panoramas)
+            return
 
-        im = None
+        im = Image.open(src)
 
         for dst, max_size in zip(dst_paths, max_sizes):
-            extension = os.path.splitext(src)[1].lower()
-            if extension in {'.svg', '.svgz'}:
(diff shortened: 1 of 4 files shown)


nikola/image_processing.py
```diff
@@ -81,69 +81,94 @@ def filter_exif(self, exif, whitelist):
...
+        * Either max_size or max_sizes should be set
+        * Either dst or dst_paths should be set
+        """
+
+        if dst_paths is None:
+            dst_paths = [dst]
+        if max_sizes is None:
+            max_sizes = [max_size]
+        if len(max_sizes) != len(dst_paths):
+            raise ValueError(f'resize_image called with incompatible arguments: {dst_paths} / {max_sizes}')
```
Kwpolska, line 105: We still support 3.5, which does not have f-strings.

ralsina: Should we? Changing it anyway :-)

Kwpolska: 3.5 is generally usable, and Ubuntu 16.04 is still supported.

ralsina: Oh, well. Changed.

ralsina:
Marking again as WIP because the delta on the demo site looks suspicious.

ralsina:
Locally there is no difference in the output between this and master :shrug:

ralsina:
Ok, so this PR is making all images small for some reason. Fixing.

ralsina:
This PR is done, review much appreciated @Kwpolska ... in the end the performance improvement is a modest 25% in my site, but it's something :-)

nikola/image_processing.py
```diff
@@ -81,111 +81,132 @@ def filter_exif(self, exif, whitelist):
...
+                _im = _im.transpose(Image.ROTATE_270)
             elif value in (7, 8):
-                im = im.transpose(Image.ROTATE_90)
+                _im = _im.transpose(Image.ROTATE_90)
             if value in (2, 4, 5, 7):
-                im = im.transpose(Image.FLIP_LEFT_RIGHT)
+                _im = _im.transpose(Image.FLIP_LEFT_RIGHT)
             exif['0th'][piexif.ImageIFD.Orientation] = 1
+            exif = self.filter_exif(exif, exif_whitelist)
+        except KeyError:
```
Kwpolska, line 129: Is that KeyError meant to be coming from `_im.info["exif"]`? We could replace it with an `'exif' in _im.info` or `_im.info.get('exif')`. Same applies to `exif['0th']`.

ralsina: Probably (who can remember? not me :-D) will change.

ralsina: Done

PR #3380, opened 2020-04-13
[WIP] Refactor Post.__init__

ralsina:
Not a functional change, just break the monster into bits

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -81,6 +81,7 @@ class Post(object):
     _paragraph_count = None
     _remaining_paragraph_count = None
     post_status = 'published'
+    has_oldstyle_metadata_tags = False
 
     def __init__(
         self,
@@ -152,7 +153,6 @@ class Post(object):
         self.publish_later = False if self.current_time is None else self.date >= self.current_time
 
         self._tags = {}
-        self.has_oldstyle_metadata_tags = False
         for lang in self.translated_to:
             if isinstance(self.meta[lang]['tags'], (list, tuple, set)):
                 _tag_list = self.meta[lang]['tags']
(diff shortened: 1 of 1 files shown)


nikola/post.py
```diff
@@ -307,7 +305,6 @@ def __init__(
                     self.has_oldstyle_metadata_tags = True
 
         # While draft comes from the tags, it's not really a tag
```
Kwpolska, line 307: I think that’s a leftover comment from before the `status` meta key?

ralsina: Probably. Have not gotten that far yet

nikola/post.py
```diff
@@ -92,176 +102,53 @@ def __init__(
...
-
-        if 'updated' not in default_metadata:
-            default_metadata['updated'] = default_metadata.get('date', None)
-
-        self.updated = to_datetime(default_metadata['updated'], tzinfo)
-
-        if 'title' not in default_metadata or 'slug' not in default_metadata \
-                or 'date' not in default_metadata:
+        # These are the required metadata fields
+        if 'title' not in default_metadata or 'slug' not in default_metadata:
```
Kwpolska, line 137: if `date` was removed from the required metadata set, then maybe we could remove it from the message as well?

ralsina:
If tests pass, this is about as much as I can stomach refactoring it right now.

Further refining of each smaller chunk will be done in the future.

ralsina:
Yes, good point.

nikola/post.py
ralsina: Hmmm I am not sure I did the right thing here. Testing it.

ralsina: Confirmed, this is ok, it fails like this without ever reaching that message:

```
[2020-04-13T15:40:35Z] ERROR: Nikola: ValueError: Missing date in file posts/foo.rst
```

PR #3390, opened 2020-04-15
Add a source('lang') method to Post

### LINKED ISSUE #3384: Add "get post contents" method to Post
To support "posts without sources" (#2659) we need to add a method to get the post's raw content, similar to `text()` (`source(lang)` maybe?)

It also simplifies compiler code, while we're at it.

ralsina:
Fix #3384

This method returns the unprocessed source for the post in the given language.
If the post is not translated to that language, uses the same logic as compile() to return the "correct" data.

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -762,7 +762,6 @@ class Post(object):
             source_data = self.compiler.split_metadata(data, self, lang)[1]
         return source_data
 
-
     def text(self, lang=None, teaser_only=False, strip_html=False, show_read_more_link=True,
              feed_read_more_link=False, feed_links_append_query=None):
         """Read the post file for that language and return its compiled contents.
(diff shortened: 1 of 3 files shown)


nikola/post.py
```diff
@@ -987,6 +1002,24 @@ def source_ext(self, prefix=False):
...
+        METADATA_FORMAT.
+
+        This has caveats:
+
+        * If you are extracting metadata from filename?
+        * If you are extracting metadata from document contents?
+
+        Because of how it's going to be written, that's not going to be used anymore.
+        """
+        breakpoint()
```
Kwpolska, line 1018: Debugging leftover?

ralsina:
Oh yeah

nikola/post.py
ralsina: I forgot to remove the started but-not-done method.

PR #3393, opened 2020-04-19
Fix #3392

### LINKED ISSUE #3392: write_metadata removes content
When trying to use this new feature to make my old 2-post files single file, this happened.

image

ralsina:
Not really happy with this solution, opinion @Kwpolska ?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -52,8 +52,8 @@ Internal
 
 * Added Post.source() method to get a Post's object unprocessed 
   contents.
-* Made is_two_file a read-only property after Post creation (Issue
-  #3392)
+* Added Post.save() method to modify Post contents.
+* Made is_two_file a property that uses save()
 
 New in v8.0.4
 =============
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -52,6 +52,8 @@ Internal
 
 * Added Post.source() method to get a Post's object unprocessed 
   contents.
+* Made is_two_file a read-only property after Post creation (Issue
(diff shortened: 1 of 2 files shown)


ralsina:
> I’m also not sure about it, but I don’t think anything will be negatively affected by this change.

The alternative is to make the setter actually read the post source, change the setting, and re-write it correctly, which is quite scary.

ralsina:
I am going to do the ambitious thing.

ralsina:
I used this branch in my site to do this, and IT WORKED!

```python
two_post_files = [p for p in site.timeline if p.is_two_file]

for p in two_post_files:
    p.is_two_file = False
    print(p.title())
```

I no longer have any separate .meta files.

Kwpolska:
I found one legitimate-ish use-case, the `pkgindex_scan` plugin updated is_two_file, because those are two-file posts, even though the other file is INI (`.plugin` or `.theme`) and not `.meta` (`pkgindex_compiler` can extract this as metadata). I updated `pkgindex_scan` to modify `_is_two_file` instead, and added a more appropriate split_metadata method to `pkgindex_compiler`.

ralsina:
Oh, nice. I am guessing we may also need a separate function to determine the "second file path", really. Internally Nikola treats it like "the same as the post file + .meta"

ralsina:
Well, as long as Nikola is the one extracting the metadata: yes.

BUT if the compiler plugin offers to extract it via supports_metadata, then
that takes precedence, I think?

PR #3405, opened 2020-05-02
Variant of Typogrify filter without ampersand

### LINKED ISSUE #3404: Typogrify filter places <span> into certain page <title> blocks
### Environment

**Python Version:** Python 3.6.10

**Nikola Version:** 8.0.4 when issue noticed

**Operating System:** Gentoo Linux

### Description:

When using the standard Typogrify filter certain page or post `<title>` blocks may end up with `<span>` tags inside. This is due to the ampersand filter wrapping the character. This issue only presents when the page or post title contains an ampersand character. I attemped several fixes in the templates with the `striphtml()` function, and various combinations of Mako filter to no avail. It seems those fixes do not work because the filters run after the template is parsed.

ncdulo:
### Pull Request Checklist

- [x] I tested my changes.

### Description
In reference to issue #3404 this pull request adds a variation of the base Typogrify filter without the ampersand filtering. The only change made to the standard `filters.typogrify` is removal of the `typo.amp()` call. This has been tested on my own site using Nikola 8.0.4, and from a test site using Nikola `master`. The work for this PR is based off of `master`.

This pull fixes #3404

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -84,6 +84,7 @@
 * `Miguel Ángel García <https://github.com/magmax>`_
 * `Mitch Chapman <https://github.com/mchapman87501>`_
 * `mrabbitt <https://github.com/mrabbitt>`_
+* `ncdulo <https://github.com/ncdulo>`_
 * `Neil MartinsenBurrell <https://github.com/neilmb>`_
 * `Niels Böhm <https://github.com/blubberdiblub>`_
 * `Niko Wenselowski <https://github.com/okin>`_
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -48,6 +48,8 @@ Bugfixes
   #3332)
 * Use correct language for hyphenation in posts that are not
   translated to all languages (Issue #3377)
+* Prevent ``<span>`` tags from appearing in certain page ``<title>``
+  blocks when using Typogrify (Issue #3404)
 
 Internal
 --------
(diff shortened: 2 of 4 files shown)


Kwpolska:
Perhaps the filters could just be fixed to ignore some tags?

We could take the `typogrify` function, change the applyfilters thing, and fix this (and possibly other bugs) without an extra filter and without having still more broken things: https://github.com/mintchaos/typogrify/blob/master/typogrify/filters.py#L348

ncdulo:
That sounds like a great solution. I was thinking something similar the other day. More of a way to pass a set of parameters to the `typogrify` function which define a selection of filters to run. That way a user would be pick exactly the set of Typogrify filters to run. I did not further pursue that solution because I was unsure of how those parameters might be passed around, and just haven't really had the energy to dig too deeply into it.

I'm at work so I can't take a deep look at the code right now, but will check it out later on. From what you mention, it looks like yours would be a better fitting solution. I'm all for reducing the duplication, and preventing other issues from arising.

ncdulo:
With this pull request being two months old, and life situations pulling me away from working towards a more appropriate fix, I don't foresee myself able to commit to any further work on this in the near future. I'd rather not leave this hanging, open on the PR board if no work is being done. If someone wants to pick this up and handle the issue more properly than the solution I came up with that would be great. I'd really like to see the original issue fixed in Nikola. Otherwise it may be time to close this one.

Thanks to the project developers for taking the time to offer suggestions towards a better fix here, and through other mediums. I'll continue maintaining the original fix implemented in my private (for unrelated reasons) Nikola source repo that powers my website.

Kwpolska:
I implemented the full fix in pull request #3447.

PR #3409, opened 2020-05-07
Add support for posts without titles (fix #3408)

### LINKED ISSUE #3408: Add support for microblog-style posts without titles
Like Dave Winer and others, I like to have some posts without titles. This is easy enough to implement with the templates. But there are a number of different conditions that can cause the title to be absent:

* Use of the `hidetitle` metadata
* Title is the "NO TITLE" placeholder
* Title is literally blank
* Post type: I'm using "micro" to indicate a microblog post.

As I say, I've got this working reasonably well _in my fork of one particular theme_. But my templates need to have various duplicated `if` conditions scattered through them, and anyone wanting similar functionality doesn't benefit. I would be good if there were a built-in feature by which templates could identify posts for which they should not render the title.

I propose adding a function to `Post.py`, such as `should_show_title()`. It would check the first three of the points above. For the fourth, I'm thinking of a new config item, something like `TYPES_WITHOUT_TITLES`, which would contain an optional list of of post types for which titles are not to be shown.

I plan to start work on this, but I'm open to suggestions for alternative approaches, corrections to my thinking, or what have you. I'm new to the Nikola codebase, so I'll be learning it as I go.

devilgate:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Allows theme authors to hide the title on a post. This is useful if you blog like [Dave Winer](http://scripting.com) or like to create Twitter-style microblog posts.

The title will be hidden if any of the following is true:

- the post type is in a list set in the new config variable `TYPES_TO_HIDE_TITLE`;
- the title is empty;
- the title is set to 'NO TITLE';
- the metadata item `hidetitle` is set on the post.

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -1091,12 +1091,12 @@ class Post(object):
             return ext
 
     def should_hide_title(self):
-        """Returns True if this post's title should be hidden. Use in templates to manage posts without titles."""
+        """Return True if this post's title should be hidden. Use in templates to manage posts without titles."""
         return self.title().strip() in ('NO TITLE', '') or self.meta('hidetitle') or \
             self.meta('type').strip() in self.types_to_hide_title
 
     def should_show_title(self):
-        """Returns True if this post's title should be displayed. Use in templates to manage posts without titles."""
+        """Return True if this post's title should be displayed. Use in templates to manage posts without titles."""
         return not self.should_hide_title()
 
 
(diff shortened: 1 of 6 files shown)


nikola/post.py
```diff
@@ -1089,6 +1090,15 @@ def source_ext(self, prefix=False):
         else:
             return ext
 
+    def should_hide_title(self):
+        """Return True if this post's title should be hidden. Use in templates to manage posts without titles."""
+        return self.title().strip() in ('NO TITLE', '') or self.meta('hidetitle') or \
```
Kwpolska, line 1095: Is there any specific reason why `NO TITLE` is used? Is a blank string not enough?

devilgate: If you import a blog from WordPress, and if there are posts without titles in the original, then they are given the "NO TITLE" string in Nikola.

ralsina: Yeah, decisions made 10 years ago :-D

Issue #2257, opened 2016-02-21
Support breaking pages/posts into multiple pages


tritium21:
It would be nice if posts could be broken into multiple parts.

For example, you could do something like insert a comment ``.. PAGE_BREAK`` into a page, then nikola could create a second page, and automatically paginate between them.

The structure would result in something like:

   /posts/my-post.html
   /posts/my-post-page-2.html

or
   /posts/my-post/index.html
   /posts/my-post/page-2/index.html

This is a low priority suggestion, but would be very handy for me.  My pages tend to be in the 20k words range.

ralsina:
If client-side pagination is good enough, you can use something like http://flaviusmatis.github.io/simplePagination.js/ and make it paginate over sections, which have a nice and handy div around them.

tritium21:
I think that may be a better idea.

felixfontein:
If this is implemented in Nikola, it would be good if also a full version of the page/post is created. I hate server-side pagination when I notice there's a second page at the bottom of the page and my internet connection is gone... :)

hilcharge:
Hi,

Is this feature still desired? If yes, I am interested in helping out with it or at least taking a shot at it.

Kwpolska:
@hilcharge Yes, it still is. Feel free to work on it, and to ask about any details you are unsure of.

hilcharge:
Hi @Kwpolska  and others, 
Have taken a shot at pagination of a long post:  
https://github.com/getnikola/nikola/pull/3418

It is just based on extra css/js in the template, not any parsing of rst etc

There are some questions raised in the pull request though, most pressing:
 - To include by theme or some other way?
 - How to parameterize the page-length - currently it is hard-coded at 10-paragraphs - was struggling to think of way to pass this value to the JS from template or elsewhere.

Kwpolska:
I think that an implementation of the pagination feature should:

* work and integrate with all themes (with small, single-line changes to templates), not require a special theme (we tried that for Jupyter, that means really bad UX),
* primarily be based on user markers, not a paragraph count (just like we do for teasers),
* not require JS to operate.

The implementation would operate with the same/similar structure as mentioned in the original issue, and would primarily be a change in how posts are rendered (probably in the render_pages task plugin, but the page URLs might also require changes to the Post class or other things).

hilcharge:
Just to clarify about the point "not require JS to operate": 

If I understand correctly, avoiding javascript is not compatible with client-side-only pagination.
[...]

PR #3418, opened 2020-06-03
Paginated post

hilcharge:
### Pull Request Checklist

- [ x ] I’ve read the [guidelines for contributing](https://github.com/getnikola/nikola/blob/master/CONTRIBUTING.rst).
- [ x ] I updated AUTHORS.txt and CHANGES.txt (if the change is non-trivial) and documentation (if applicable).
- [ x ] I tested my changes.

### Description

This is adding some pagination functionality, based on javascript for browser-based pagination of long post. The pagination is triggered by a script and related CSS that can be included in a template. (see paginated-post.tmpl for a sample).

Based on requested / help-wanted feature https://github.com/getnikola/nikola/issues/2257

*Disclosure*: probably not ready for merge yet. Hoping to get some feedback. 

The main functionality is in nikola-pagination.js.

To use:

1. make sure the pagination theme assets are being used
2. add in the following to template:
```
<link type="text/css" rel="stylesheet" href="../../assets/css/simplePagination.css"/>
... <!-- Then at bottom: -->
<script type="text/javascript" src="../../assets/js/jquery.min.js"></script>
<script type="text/javascript" src="../../assets/js/jquery.simplePagination.js"></script>
<script type="text/javascript" src="../../assets/js/nikola-pagination.js"></script>`
```
3. Build the post, and check that the pagination is working - it is just browser-based, not server-side

Some sticking points / questions: 
- how to parameterize the page-length - currently it is hard coded at 10 paragraphs. 
- is "theme" appropriate way to include this feature? Or where else to store. 
- basically just need to provide some extra css, js, and a template which includes these.
- This affects documentation placement as well (currently docs are just within the JS)
- sample template file just has the script/  link tags added in directly, might be missing a preferred syntax
 - suggestions for adding a test -- maybe `tests/integration/test_paginated_post` ?


Kwpolska:
This implementation is incorrect. It creates a new theme, which only contains assets and a template, so it isn’t a valid Nikola theme. The selected JS library is also worrying because of a far-reaching CSS.

I think that an implementation of the pagination feature should:

* work and integrate with all themes (with small, single-line changes to templates), not require a special theme (we tried that for Jupyter, that means really bad UX),
* primarily be based on user markers, not a paragraph count (just like we do for teasers),
* not require JS to operate.

The implementation would operate with the same/similar structure as mentioned in the original issue, and would primarily be a change in how posts are rendered (probably in the render_pages task plugin, but the page URLs might also require changes to the Post class or other things).

Kwpolska:
I’ll close this PR, but feel free to open a new one with a different approach.

PR #3424, opened 2020-07-01
Fix date handling for post_list. Fixes #3420

### LINKED ISSUE #3420: In the `post-list` shortcode, the `now` feature doesn't work; and date handling generally doesn't behave as expected
### Environment

**Python Version:**

3.7.7 

**Nikola Version:**

v8.0.4

**Operating System:**

MacOS Catalina

### Description:

I've got some code like this in a page:

```python
{{% post-list date=day==now template=post_list_with_expansion.tmpl %}}{{% /post-list %}}
```

The template is one of my own, but it's pretty simple.

Running `nikola build` gives this:

```
Scanning posts........done!
.  render_posts:cache/pages/index.html
TaskError - taskid:render_posts:cache/pages/index.html
PythonAction Error
  File "/Users/martin/.venv/nikola-dev/lib/python3.7/site-packages/doit/action.py", line 437, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/Users/martin/Projects/nikola/nikola/post.py", line 716, in compile
    lang)
  File "/Users/martin/Projects/nikola/nikola/plugins/compile/markdown/__init__.py", line 122, in compile
    output, shortcode_deps = self.compile_string(data, source, is_two_file, post, lang)
  File "/Users/martin/Projects/nikola/nikola/plugins/compile/markdown/__init__.py", line 111, in compile_string
    output, shortcode_deps = self.site.apply_shortcodes_uuid(output, shortcodes, filename=source_path, extra_context={'post': post})
  File "/Users/martin/Projects/nikola/nikola/nikola.py", line 1726, in apply_shortcodes_uuid
    replacement, _deps = shortcodes.apply_shortcodes(v, self.shortcode_registry, self, filename, lang=lang, extra_context=extra_context)
  File "/Users/martin/Projects/nikola/nikola/shortcodes.py", line 370, in apply_shortcodes
    res = f(*args, **kw)
  File "/Users/martin/Projects/nikola/nikola/plugins/shortcode/post_list.py", line 201, in handler
    filtered_timeline = [p for p in filtered_timeline if date_in_range(utils.html_unescape(date), p.date, now=_now)]
  File "/Users/martin/Projects/nikola/nikola/plugins/shortcode/post_list.py", line 201, in <listcomp>
    filtered_timeline = [p for p in filtered_timeline if date_in_range(utils.html_unescape(date), p.date, now=_now)]
  File "/Users/martin/Projects/nikola/nikola/packages/datecond/__init__.py", line 80, in date_in_range
    right = int(value)
ValueError: invalid literal for int() with base 10: 'now'
```

I tried various ways of quoting "now", but that just caused other errors. Turns out it's the code in the `date_in_range` function in `datecond.__init__`:

```python
        if attribute in ('weekday', 'isoweekday'):
            left = getattr(date, attribute)()
            right = int(value)
        elif attribute:
            left = getattr(date, attribute)
            right = int(value)
        elif value == 'now':
            left = date
            right = now or datetime.datetime.now()
```

It fails on the ` left = getattr(date, attribute)` line. If I move the `elif value == 'now':` block up to before that, it works.

A simple enough fix, which I'm happy to apply. But when I say, "It works," I just mean the build doesn't fail. I don't actually get a list of today's posts displayed. That is because we're using `date time.now()` for all the comparisons, which is a time. I think we should be using `date.today()`, instead.

My inclination is to change the functionality of `now` in `post-list`. I think most people would expect it to behave in terms of days, not times. But maybe not. An alternative is to add a `today` option to `post-list`, and make it behave the way I expect.

What do people think?

devilgate:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Changes the way the `now` special value is used in the `post_list` directive. It now uses days, rather than timestamps.


Kwpolska:
This will be fixed in a different, more flexible way, and also as part of the `datecond` package. (I had a comment written in #3420, but apparently forgot to send it.)

PR #3437, opened 2020-07-13
Remove empty return to allow else clause in post-list shortcode.

devilgate:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Removes an empty return which prevented the template being called if no posts were returned, and hence stopped the template's `else` clause from executing.


CHANGES.txt
```diff
@@ -4,6 +4,7 @@ New in v8.1.1
 Bugfixes
 --------
 
+* Allow else clause in post-list plugin. (Issue #3436)
```
Kwpolska, line 7: This should be in the “new in master” block, please rebase on origin/master and fix the changelog.

devilgate: Sorry, what's the “new in master” block? I can't find a section like that.

Kwpolska: You’re looking at an outdated copy of this file, pull from upstream.

image

devilgate: Ah! Got it. Thanks.

PR #3439, opened 2020-07-14
fix the path handlers on Windows

PoByBolek:
### Pull Request Checklist

- [x] I tested my changes.
- [x] I added more tests ;)

### Description

If one of the path components returned by a path handler contains a forward slash (``/``), ``os.path.normpath()`` turns that into a backward slash (``\``) on Windows.

For example, if I have set ``TAGS_INDEX_PATH`` to ``"blog/tags.html"``, then ``"link://tag_index"`` would render as ``"/blog%5Ctags.html"`` on Windows.

I *think* this is related to #2649 except that I noticed it because the ``permalink`` template variable in the taxonomy overviews contained ``"%5C"`` instead of ``"/"``.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -44,6 +44,7 @@
 * `Felix Fontein <https://github.com/felixfontein>`_
 * `Felix Schwarz <https://github.com/FelixSchwarz>`_
 * `Florian Finkernagel <https://github.com/TyberiusPrime>`_
+* `Florian Zimmermann <https://github.com/PoByBolek>`_
 * `follower <https://github.com/follower>`_
 * `George Leslie-Waksman <https://github.com/gwax>`_
 * `Grzegorz Śliwiński <https://github.com/fizyk>`_
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Bugfixes
 * Fix handling of duplicate plugins on Windows
 * Allow else clause in post-list plugin. (Issue #3436)
 * Ensure `type` metadata value from plugins is preserved (Issue 3445)
+* Fix path handlers with slashes on Windows
 
 New in v8.1.1
 =============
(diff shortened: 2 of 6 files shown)


PoByBolek:
Hm, maybe the root cause is that the taxonomy plugins simply return the ``*_INDEX_PATH`` or ``*_PATH`` config values without further processing (see [authors.py:97], [tags.py:105], etc). Shouldn't they return a list of path components? Or are the ``*_PATH`` variables simply not intended to hold nested paths?

  [authors.py:97]: https://github.com/getnikola/nikola/blob/980161de10cc961a8329f1e61f8e2e3d758a4fa1/nikola/plugins/task/authors.py#L97. 
  [tags.py:105]: https://github.com/getnikola/nikola/blob/980161de10cc961a8329f1e61f8e2e3d758a4fa1/nikola/plugins/task/tags.py#L105

Kwpolska:
It is fine to have a nested structure in `*_PATH`. It might be better if we split it earlier. Perhaps you could instead modify the path handlers to split the `*_PATH` variables (by `/` and `\`)?

PoByBolek:
Sure I can. I was just thinking that sanitizing the paths in a central location would prevent other issues like this. But I guess you are right, it's better to have consistent path values as soon as possible instead of letting the weird ones bubble all the way up to `Nikola.path()`.

Do you really want me to split on both `/` and `\`? Wouldn't it be better to say that Nikola internally only deals with forward slashes?

Kwpolska:
Using forward slashes everywhere would probably be okay, Windows accepts paths with `/`.

PoByBolek:
Is this okay so far? Do you need anything else?

Kwpolska:
This got lost in my inbox, sorry.

Kwpolska (approved):

Kwpolska:
Thanks for fixing this!

PR #3446, opened 2020-07-24
Preserve value of type metadata from plugins. Fixes #3445

### LINKED ISSUE #3445: Post type metadata cannot be overridden by plugins
I'm writing a plugin that needs to set the `type` metadata (specifically to "micro", for microblog posts). But as it stands, The value that the plugin sets always gets replaced by the default value, "text". So I'm making a core change under this number to let my plugin work.

devilgate:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Ensures that, if the `type` metadata is set by a plugin, that value is written to the post file.

### DIFF
--- a/nikola/plugins/command/new_post.py
+++ b/nikola/plugins/command/new_post.py
@@ -243,7 +243,7 @@ class CommandNewPost(Command):
         date_path_opt = options['date-path']
         date_path_auto = self.site.config['NEW_POST_DATE_PATH'] and content_type == 'post'
         date_path_format = self.site.config['NEW_POST_DATE_PATH_FORMAT'].strip('/')
-        post_type = options['type'] if 'type' in options else 'text'
+        post_type = options.get('type', 'text')
 
         if wants_available:
             self.print_compilers()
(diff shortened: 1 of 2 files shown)


nikola/plugins/command/new_post.py
```diff
@@ -243,6 +243,7 @@ def _execute(self, options, args):
         date_path_opt = options['date-path']
         date_path_auto = self.site.config['NEW_POST_DATE_PATH'] and content_type == 'post'
         date_path_format = self.site.config['NEW_POST_DATE_PATH_FORMAT'].strip('/')
+        post_type = options['type'] if str('type') in options else 'text'
```
Kwpolska, line 246: Why not `if 'type' in options`?

devilgate: Ah! My IDE was reporting that `type` clashes with a built-in. The other way works too, but I tried it this way to see if it removed the warning, which it did. I can change it back if you want.

Kwpolska: You might have mixed up a `type` variable and a `type` string, please change to a regular string in the `if` condition.

devilgate: I think that's what it was. I've made the change.

PR #3459, opened 2020-09-13
Add support for the utterances comment system (https://utteranc.es)

### LINKED ISSUE #3453: Native support for Utterances comment system
See https://utteranc.es/

Utterances gives you a script tag that you can add to your site, but it's not clear where that should go in the conf.py.

asmeurer:
### Pull Request Checklist

- [ ] I tested my changes.

### Description

Fixes #3453

There is one issue with this, which is that the configuration options for
Utterances are stored in GLOBAL_CONTEXT['utterances_config'], similar to isso,
but one configuration option is required (either issue-term or issue-number).
How can we make this error in the build if it isn't provided? Otherwise, the
only indication that it isn't working is an error in the console log.

### DIFF
--- a/nikola/data/themes/base/templates/comments_helper_utterances.tmpl
+++ b/nikola/data/themes/base/templates/comments_helper_utterances.tmpl
@@ -20,7 +20,4 @@
 
 
 <%def name="comment_link_script()">
-    %if comment_system_id and 'index' in pagekind:
-        <script src="${comment_system_id}js/count.min.js" data-utterances="${comment_system_id}" data-utterances-lang="${lang}"></script>
-    %endif
 </%def>
--- a/nikola/data/themes/base/templates/comments_helper_utterances.tmpl
+++ b/nikola/data/themes/base/templates/comments_helper_utterances.tmpl
@@ -1,0 +1,26 @@
+## -*- coding: utf-8 -*-
+<%def name="comment_form(url, title, identifier)">
+    %if comment_system_id:
+        <div data-title="${title|h}" id="utterances-thread"></div>
+        <script src="https://utteranc.es/client.js" repo="${comment_system_id}"
+        % if utterances_config:
+        % for k, v in utterances_config.items():
+        ${k}="${v}"
+        % endfor
+        % endif
+        ></script>
+    %endif
+</%def>
+
+<%def name="comment_link(link, identifier)">
+    %if comment_system_id:
+        <a href="${link}#utterances-thread">${messages("Comments")}</a>
+    %endif
+</%def>
+
+
+<%def name="comment_link_script()">
+    %if comment_system_id and 'index' in pagekind:
+        <script src="${comment_system_id}js/count.min.js" data-utterances="${comment_system_id}" data-utterances-lang="${lang}"></script>
+    %endif
+</%def>
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -100,6 +100,7 @@ LEGAL_VALUES = {
         'isso',
         'muut',
         'commento',
+        'utterances',
     ],
     'TRANSLATIONS': {
         'af': 'Afrikaans',
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -836,6 +836,13 @@ class Nikola(object):
             utils.LOGGER.error('CATEGORY_PAGES_FOLLOW_DESTPATH requires CATEGORY_ALLOW_HIERARCHIES = True, CATEGORY_OUTPUT_FLAT_HIERARCHY = False.')
             sys.exit(1)
 
+        # The Utterances comment system has a required configuration value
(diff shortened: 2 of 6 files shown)


nikola/data/themes/base/templates/comments_helper_utterances.tmpl
```diff
@@ -0,0 +1,26 @@
...
+<%def name="comment_link(link, identifier)">
+    %if comment_system_id:
+        <a href="${link}#utterances-thread">${messages("Comments")}</a>
+    %endif
+</%def>
+
+
+<%def name="comment_link_script()">
+    %if comment_system_id and 'index' in pagekind:
+        <script src="${comment_system_id}js/count.min.js" data-utterances="${comment_system_id}" data-utterances-lang="${lang}"></script>
```
Kwpolska, line 24: This looks like an isso leftover, right? If uterrances doesn’t have a comment-link thing (that produces "%d comments", like Disqus does), leave this function empty.

Kwpolska (changes requested):

nikola/data/themes/base/templates/comments_helper_utterances.tmpl
asmeurer: Oh right. I missed this. I'll have to see if it supports that.

asmeurer: I guess it's a requested feature, but not yet supported https://github.com/utterance/utterances/issues/36. So I'll leave this out for now. If there are any updates on that issue we can return to this.

asmeurer:
I added a check that the required config is defined. 

This is ready for review from my end. I did a basic test of this on my own blog, and it seems to work. You may want to test it yourself to make sure I didn't miss anything. I mostly copied the theme code based on the other systems, so hopefully everything there is still good. 

Something that I might want to add to my own blog, but I'm not sure if I should put anything about it in this PR, is some boilerplate text under the "comments" header like "This blog uses [Utterances](https://utteranc.es) for comments. You can comment below, or directly on the corresponding GitHub issue \<link to issue here>." The reason is that commenting on the page requires authorizing Utterances on your GitHub account, which some people might not like, but you don't actually have to do that if you comment on the issue directly.

nikola/nikola.py
```diff
@@ -836,6 +836,13 @@ def __init__(self, **config):
             utils.LOGGER.error('CATEGORY_PAGES_FOLLOW_DESTPATH requires CATEGORY_ALLOW_HIERARCHIES = True, CATEGORY_OUTPUT_FLAT_HIERARCHY = False.')
             sys.exit(1)
 
+        # The Utterances comment system has a required configuration value
+        if self.config.get('COMMENT_SYSTEM') == 'utterances':
+            utterances_config = self.config.get('GLOBAL_CONTEXT', {}).get('utterances_config', {})
+            if not ('issue-term' in utterances_config or 'issue-number' in utterances_config):
+                utils.LOGGER.error("COMMENT_SYSTEM = 'utterances' must have either GLOBAL_CONTEXT['utterances_config']['issue-term'] or GLOBAL_CONTEXT['utterances_config']['issue-term'] defined.")
+                sys.exit(1)
```
Kwpolska, line 844: I wouldn’t exit, it’s not that fatal of an issue IMO.

nikola/nikola.py
```diff
@@ -836,6 +836,13 @@ def __init__(self, **config):
             utils.LOGGER.error('CATEGORY_PAGES_FOLLOW_DESTPATH requires CATEGORY_ALLOW_HIERARCHIES = True, CATEGORY_OUTPUT_FLAT_HIERARCHY = False.')
             sys.exit(1)
 
+        # The Utterances comment system has a required configuration value
+        if self.config.get('COMMENT_SYSTEM') == 'utterances':
```
Kwpolska, line 840: You also need to add `'utterances'` to the list on [line 102 of `nikola.py`](https://github.com/getnikola/nikola/blob/eea7dbdb0a25a7d154f35068d77c5c2e2b975c13/nikola/nikola.py#L102).

Kwpolska (changes requested):

Kwpolska (approved):

Kwpolska:
> Something that I might want to add to my own blog, but I'm not sure if I should put anything about it in this PR, is some boilerplate text under the "comments" header like "This blog uses Utterances for comments. You can comment below, or directly on the corresponding GitHub issue <link to issue here>." The reason is that commenting on the page requires authorizing Utterances on your GitHub account, which some people might not like, but you don't actually have to do that if you comment on the issue directly.

This boilerplate text would require translating, and I’m not sure if we should add a new translation string for a comment system.

Kwpolska:
Thanks for contributing this!

asmeurer:
Nice. Now I just need to figure out how I can best migrate my old Disqus comments into Utterances, and finally ditch Disqus.

PR #3460, opened 2020-09-16
Support the Markdown extension for Quotebacks

devilgate:
### Pull Request Checklist

- [x] I tested my changes.

### Description

[Quotebacks](https://quotebacks.net) is a recently-developed JavaScript library for applying a standard styling to blockquotes on web pages. As released, it also includes a browser plugin for Chrome, that helps with capturing the quotes and applying the correct HTML. 

However, if you don't use Chrome, and/or you prefer to write in Markdown, it's not ideal. 

Luckily [Matt Webb](http://interconnected.org/home/2020/06/16/quotebacks) has written an [extension for Python-Markdown](https://github.com/genmon/quotebacks-mdx) which allows the writer to activate the Quotebacks styling using simple Markdown.

This change incorporates that extension into Nikola.

To use it you'll have to have the `quotebacks.js` file installed, and have suitable code in your template to ensure that it's referenced in posts. Or just suitable code to import it from the CDN.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@ New in master
 Features
 --------
 
+* Add support for Matt Webb's Quotebacks Markdown extension.
 * Add ``navbar_custom_bg`` theme option to ``bootstrap4`` and document
   options for ``bootstrap4`` better (Issue #3443)
 * Add Marathi translation
--- a/nikola/plugins/compile/markdown/quotebacks_mdx.plugin
+++ b/nikola/plugins/compile/markdown/quotebacks_mdx.plugin
@@ -1,0 +1,14 @@
+[Core]
+name = quotebacks_mdx
+module = quotebacks_mdx
+
+[Nikola]
+compiler = markdown
+PluginCategory = CompilerExtension
+
+[Documentation]
+author = Matt Webb (tweaked by Martin McCallion for Nikola compatibility)
+version = 0.1
+website = https://github.com/genmon/quotebacks-mdx
+description = Markdown extension for simpler Quotenacks support (https://quotebacks.net/).
+
--- a/nikola/plugins/compile/markdown/quotebacks_mdx.py
+++ b/nikola/plugins/compile/markdown/quotebacks_mdx.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+# Quotebacks extension for Python-Markdown
+
+Use by including `QuotebacksExtension()` in the extensions list.
+
+Converts Markdown of the form:
+
+```markdown
+> QUOTED-CONTENT
+>
+> -- AUTHOR, [LINKTEXT](URL)
+```
+
+to the quotebacks format.
+"""
+
+
+import logging
+import markdown
+import re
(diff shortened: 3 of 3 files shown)


Kwpolska:
I don’t think this should be built into Nikola, especially in a partially functional form — and I don’t think we should add this to default themes. Please make this an independent plugin and make a PR against getnikola/plugins.

devilgate:
While I take your point, it raises two questions for me:

1. Can one make a plugin that installs a `.py` file into a particular directory, as is needed here? And if so, how?
2. Do you think it might be better not to instantly close a PR like this one? Instead, maybe leave it open for a day or two, to give other people a chance to see it and possibly comment? Opinions might differ.

Kwpolska:
1. It’s not needed. If the plugin is correctly categorized (subclass of MarkdownExtension + compiler=markdown in .plugin), Nikola should load it from site/plugins just fine. If it does not, report a bug or make a PR to fix this (but there is 
2. If you want to discuss a new feature for Nikola core, please create an issue. As a maintainer, I decided to reject this PR, because new plugins in core are not something that should be added without significant benefits to users. I believe that maintainers should have the right to make decisions about the future path of the project, even if some people do not agree with them, especially if those changes would add more work for the maintainers.

To properly implement Quotebacks in core, you would need to:

1. Add it to the default templates and `npm_assets`…
2. But since not everyone wants it, would that require adding a new config setting, or a dynamic template variable like “page_uses_quotebacks”?
3. And wouldn’t it be better as a general thing, like a shortcode, or at least a similar reST extension?

There are too many questions, drawbacks, added maintenance and requirements for adding this. Because of this, I decided to reject this pull request. As I mentioned before, feel free to open a PR against getnikola/plugins, where none of the issues apply, and which is the correct place for extra functionality that may interest some users (but not a majority of them).

Does this mean that Nikola core won’t get any new features? No. But features in core must be useful for a large group of users, without breaking existing sites (what if someone is using this quoting style on their site — it’s not that uncommon, and now they get some fancy HTML, and possibly a third-party library if steps 1/2/3 were implemented — and now the quotes look weird?), and the more maintenance work they require, the more “useful” it should be to users. And yes, this can be subjective metric, but Quotebacks does not meet it IMO.

Also, while Quotebacks doesn’t seem to depend on an external service (and in fact, seems to just add some fancy CSS? Why would you need JS for this? [and I’m not the only one wondering what it even does](https://github.com/Blogger-Peer-Review/quotebacks/issues/33)), but if it was, what if the service closed down? We already went through this with annotations, which just closed down one day and we had to live with a mostly blank `annotations_helper.tmpl`, which is also why I would prefer not to introduce a new third-party thing.

devilgate:
Thanks for that, Chris, and sorry for causing problems. When I suggested keeping this open for discussion, I was forgetting that it is a pull request, not an issue. Of course I should have opened an issue first.

Regarding what Quotebacks are and why people would use them: I was confused myself at first, and I'm still not _totally_ convinced about them. But I think the idea is to have a common form for blockquotes that might become as commonplace and recognisable as embedded tweets. And using a JS library when you could just do it with CSS? Dropping a library in is just easier than hacking around with CSS, I guess.

PR #3475, opened 2020-11-11
Fix #3471 — fix previewimages for non-server-root sites

### LINKED ISSUE #3471: Incorrect src attribute for images in featured posts in bootblog4 theme
### Environment

**Python Version:** 3.8.5

**Nikola Version:** 8.1.1

**Operating System:** Ubuntu 20.10

### Description:

It appears that Nikola 8.1.1 has introduced either a bug or a backwards-incompatible change that makes images in featured posts using the bootblog4 template render incorrecty.

My featured posts include these headings: 
```
.. status: featured
.. previewimage: /galleries/face-a-day-2020/20200817-me.png
```

My nikola site is rendered under a path: 
```python
# conf.py
SITE_URL = "https://carles.barrobes.com/drawing/"
# BASE_URL not set
```

- With nikola 8.0.4, this would generate a correct `src` attribute `galleries/face-a-day-2020/20200817-me.png`
- In nikola 8.1.1, this generates a relative path that doesn't work `drawing/galleries/face-a-day-2020/20200817-me.png` as it generates an invalid image URL of `https://carles.barrobes.com/drawing/drawing/galleries/face-a-day-2020/20200817-me.png` (duplicated `drawing` in the path).

The source of the issue seems to have been introduced here: https://github.com/getnikola/nikola/commit/e559cfda090edacbca0769395703fdbde50691cb

Kwpolska:
This fixes previewimage path generation for non-server-root sites and simplifies the code for all cases.

This is #3471, cc @txels.

(There used to be a “not server root” label. Then it was removed, because we got better at it… but now, here’s a bug that this label would apply to. Although it’s my fault for overcomplicating this in the first place.)

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -1085,11 +1085,13 @@ class Post(object):
 
         image_path = self.meta[lang]['previewimage']
         if not image_path:
-            return self._default_preview_image
+            image_path = self._default_preview_image
 
-        # This is further parsed by the template, because we don’t have access
-        # to the URL replacer here.  (Issue #1473)
-        return image_path
+        if image_path.startswith("/"):
+            # Paths starting with slashes are expected to be root-relative, pass them directly.
+            return image_path
+        # Other paths are relative to the permalink. The path will be made prettier by the URL replacer later.
+        return urljoin(self.permalink(lang), image_path)
 
     def source_ext(self, prefix=False):
         """Return the source file extension.
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -1087,7 +1087,7 @@ class Post(object):
         if not image_path:
             image_path = self._default_preview_image
 
-        if image_path.startswith("/"):
+        if not image_path or image_path.startswith("/"):
             # Paths starting with slashes are expected to be root-relative, pass them directly.
             return image_path
         # Other paths are relative to the permalink. The path will be made prettier by the URL replacer later.
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,6 +19,8 @@ Features
 Bugfixes
 --------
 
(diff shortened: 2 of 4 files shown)


nikola/post.py
```diff
@@ -1085,11 +1085,13 @@ def previewimage(self, lang=None):
 
         image_path = self.meta[lang]['previewimage']
         if not image_path:
-            return self._default_preview_image
+            image_path = self._default_preview_image
 
-        # This is further parsed by the template, because we don’t have access
-        # to the URL replacer here.  (Issue #1473)
-        return image_path
+        if image_path.startswith("/"):
```
txels, line 1090: Changing this line to `if image_path and image_path.startswith("/"):` works for me. Otherwise I get an exception.

```python
    if image_path.startswith("/"):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

txels:
Thanks for working on a fix this quickly!

nikola/post.py
txels: Ah but your CI is [already telling you this](https://travis-ci.org/github/getnikola/nikola/jobs/742854436#L679) ;)

Kwpolska: I added a slightly different fix, seems to work now in all cases (and I forgot to look at CI, assuming it’s an expected failure of baseline testing). Can you test again?

txels: Seems to work, cheers.

txels: I'd myself have gone for an explicit `image_path is None` rather than `not image_path` ;)

Kwpolska: For `image_path = ""`, returning a blank screen is also fine.

txels: Ah good point. I only had my case in mind :)

PR #3477, opened 2020-11-15
Start GitHub Actions

felixfontein:
Actions won't be run if this isn't in master.

Since this isn't running anything yet, it should work fine and not start failing right away.


felixfontein:
Interestingly, they actually seem to run in this PR. I guess I can just continue without this being merged :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Basic tests (Python ${{ matrix.python }} on ${{ matrix.image }})
+    strategy:
+      matrix:
+        python:
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
+        image:
+          - unbutu-18.04
```
Kwpolska, line 23: 1. There’s a typo. (u**nb**utu)
2. We should limit the Ubuntu versions we use as much as possible. It looks like you can install everything in 20.04: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md — so we should use that, or `ubuntu-latest` (which will become 20.04 soon: https://github.com/actions/virtual-environments/issues/1816)
3. This should use `setup-python`, as in this example: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python

I’ll try to commit a fix to all those issues.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    strategy:
+      matrix:
+        python:
+          - 3.8
+    runs-on: ubuntu-18.04
+    steps:
+      - run: echo hello
+
+  basereq:
+    name: Compare to baseline
```
Kwpolska, line 49: `basereq` is used to build the demo site with the base `requirements.txt` (without extras) and see if it doesn’t crash.

Kwpolska (changes requested):

.github/workflows/ci.yml
felixfontein: Please don't, I'm actively working on this.

Kwpolska: This issue was assigned to me. In the future, please ask before working on an assigned issue.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+  nikola:
+    name: Basic tests (Python ${{ matrix.python }} on ${{ matrix.image }})
+    strategy:
+      fail-fast: false
+      matrix:
+        python:
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
```
Kwpolska, line 22: Please add 3.9.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+            image: ubuntu-16.04
+    runs-on: '${{ matrix.image }}'
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '${{ matrix.python }}'
+      - name: Install test requirements
```
Kwpolska, line 43: Before installing requirements, we need to run this:

    pip install -U --upgrade-strategy=eager pip setuptools wheel

.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+        run: |
+          pip install --upgrade-strategy eager -Ur requirements-tests.txt
+      - name: Install coverage requirements
```
Kwpolska, line 46: Let’s drop Codacy, we don’t really use that for anything.

.github/workflows/ci.yml
felixfontein: I used the same Ubuntu versions that were mentioned in the travis config. I was using this list: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+          - 3.8
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '${{ matrix.python }}'
+      - name: Install requirements
```
Kwpolska, line 106: This also requires `pip install -U --upgrade-strategy=eager pip setuptools wheel`

Kwpolska (changes requested):

.github/workflows/ci.yml
Kwpolska: Travis uses 16.04 for 3.5 because they don’t offer 3.5 on newer versions. The provider change lets us switch to a more modern OS version for all tests.

Kwpolska (approved):

Kwpolska:
This now works with macOS and Windows.

@felixfontein, please review the last 3 commits I made.

@ralsina, could you review this? Thanks.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+jobs:
+  nikola:
+    name: Nikola test suite (Python ${{ matrix.python }} on ${{ matrix.image }})
+    strategy:
+      fail-fast: false
+      matrix:
+        python: [3.5, 3.6, 3.7, 3.8, 3.9]
+        image:
+          - ubuntu-latest
+        include:
```
felixfontein, line 18: Neat, I haven't seen that in use before, it makes adding special cases a lot easier (and nicer)!

felixfontein (commented):
Cannot approve since I create this PR, but all new changes look good :)

README.rst
```diff
@@ -9,8 +9,8 @@ In goes content, out comes a website, ready to deploy.
 .. image:: https://img.shields.io/requires/github/getnikola/nikola.svg
    :target: https://requires.io/github/getnikola/nikola/requirements/?branch=master
 
-.. image:: http://img.shields.io/travis/getnikola/nikola.svg
-   :target: https://travis-ci.org/getnikola/nikola
+.. image:: https://github.com/getnikola/nikola/workflows/Nikola%20CI/badge.svg
+   :target: https://github.com/getnikola/nikola/actions
 
 .. image:: http://img.shields.io/coveralls/getnikola/nikola.png
```
felixfontein, line 15: PS: is this one taking its data from codacy? If yes, I guess it should be removed.

Kwpolska: It’s taking data from Coveralls, which is another service — one that we haven’t used for even longer. Good catch, I’ll remove it.

PR #3489, opened 2020-12-20
Accept string datetime representations for formatting.

arusahni:
### Pull Request Checklist

- [x] I tested my changes.

### Description

This addresses an issue that I asked about on the mailing list, but received no response from the community about.

`post.formatted_date()` accepts an optional second argument that allows one to format an arbitrary date. This is very convenient for template authors. However, If attempting to render custom post metadata fields, there is currently no way to convert the parsed `post.meta('fieldname')` value to a datetime for formatting.  I see two possible solutions:

1. Create a `post.parse_datetime(dt: str) -> datetime` method
2. Update the existing `post.formatted_date()` method to support strings.

I opted for the latter, here, because it feels more intuitive from the templating perspective.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,8 @@ Features
 
 * Provide the full ``GLOBAL_CONTEXT`` to the post list shortcode plugin
   (Issue #3481)
+* Accept and parse string datetime representations passed to
+  ``post.formatted_date()``
 
 New in v8.1.2
 =============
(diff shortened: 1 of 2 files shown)


Kwpolska:
This solves one problem and it uses an ugly workaround. The best solution would be to write a plugin that converts your custom meta field to a `datetime` using `to_datetime`, as we already do for `date` and `updated` — then you could use `formatted_date` as-is. Another solution that is simpler but also acceptable would be to expose `to_datetime` in templates. We don’t do that by default (and I don’t think we should), but you can do it using `GLOBAL_CONTEXT` in `conf.py`.

arusahni:
Sounds good.

PR #3509, opened 2021-02-07
Add NIKOLA_LOG_FORMAT & NIKOLA_LOG_TIME_FORMAT environment variables

matthew16550:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Add `NIKOLA_LOG_FORMAT` & `NIKOLA_LOG_TIME_FORMAT` environment variables to configure the logging format.

Inspired by some plugin debugging where I needed to see millisecond timestamps and the process ids during auto builds.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,8 @@ Features
   by specifying a dict of ``{".extension": [options]}`` (Issue #3492)
 * Allow boolean/integer ``pretty_url`` post meta values in YAML/TOML
   (Issue #3503)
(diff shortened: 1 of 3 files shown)


Kwpolska:
Debugging some very unusual timing conditions would be the only real use-case for this. We don’t have too many of those, and if you’re debugging stuff, you likely make other modifications to the code to get better debug information — so in this very rare case, you can just edit `log.py` with the desired format. It’s definitely not worth anincrease in the number of configurable things.

PR #3513, opened 2021-02-18
New 'auto_command_starting' signal when 'nikola auto' is starting.

matthew16550:
### Pull Request Checklist

- [x] I tested my changes.

### Description

When adding support for the PlantUML PicoWeb server (getnikola/plugins#378) I would like `nikola auto` to start one PicoWeb server and share it with all child build processes.  So the `plantuml` plugin needs to know when `nikola auto` is happening.

The simplest solution I've thought of is a new `auto_command_starting` signal but perhaps something like these more generic solutions would be preferred?
*  A new `command_starting` signal including command name as a param
* `BasePlugin.get_command_name()`

(Haven't looked closely if either of those is actually possible)

### DIFF
--- a/nikola/plugins/command/auto/__init__.py
+++ b/nikola/plugins/command/auto/__init__.py
@@ -149,7 +149,7 @@ class CommandAuto(Command):
             req_missing(['aiohttp'], 'use the "auto" command')
         elif Observer is None:
             req_missing(['watchdog'], 'use the "auto" command')
-            
+
         blinker.signal('auto_command_starting').send(self.site)
 
         if sys.argv[0].endswith('__main__.py'):
--- a/nikola/plugins/command/auto/__init__.py
+++ b/nikola/plugins/command/auto/__init__.py
@@ -136,9 +136,8 @@ class CommandAuto(Command):
     ]
 
     def _execute(self, options, args):
-        blinker.signal('auto_command_starting').send(self.site)
-
         """Start the watcher."""
+
         self.sockets = []
         self.rebuild_queue = asyncio.Queue()
         self.reload_queue = asyncio.Queue()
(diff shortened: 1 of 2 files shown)


matthew16550:
The build is failing on an unrelated test and I haven't been able to reproduce it locally as yet. 😢 

```
    def test_compiling_markdown(
        compiler, input_path, output_path, input_str, expected_output
    ):
        output = markdown_compile(compiler, input_path, output_path, input_str)
>       assert output.strip() == expected_output.strip()
E       assert '<table class...</tr></table>' == '<table class...</tr></table>'
E         Skipping 73 identical leading characters in diff, use -v to show
E         - div"><pre>1</pre></div></td><td class="code"><pre class="code literal-block"><span></span><code><span class="kn">from</span> <span class="nn">this</span>
E         + div"><pre><span class="normal">1</span></pre></div></td><td class="code"><pre class="code literal-block"><span></span><code><span class="kn">from</span> <span class="nn">this</span>
E         ?           +++++++++++++++++++++ +++++++
E           </code></pre>
E           </td></tr></table>
```

matthew16550:
Made #3514 for why the build is failiing.

Kwpolska:
> A new command_starting signal including command name as a param

What commands, other than `auto` (and perhaps `build`) could make use of that?

(Although `build` could end up allowing people to do fairly ugly things; it does make sense to use this in `auto` to launch PlantUML, or perhaps some JS/CSS live compiler.)

nikola/plugins/command/auto/__init__.py
```diff
@@ -135,6 +136,8 @@ class CommandAuto(Command):
     ]
 
     def _execute(self, options, args):
+        blinker.signal('auto_command_starting').send(self.site)
```
Kwpolska, line 139: The docstring should be the first line in a function. I’ll fix it.

Kwpolska (changes requested):

Kwpolska (approved):

Kwpolska:
I merged the `auto_command_starting` version, if there are any other commands which could use a signal (I don’t really think we need more), you can open further PRs. A signal for any command seems unnecessary to me (it also becomes harder to catch).

matthew16550:
| What commands, other than auto (and perhaps build) could make use of that?

Hadn't thought of any other use cases, just thinking out loud.

Thanks for finishing this one.

PR #3516, opened 2021-02-18
Add math CSS class to typogrify ignores

### LINKED ISSUE #3512: Consecutive capitals in Math are labeled as span caps
### Environment

**Python Version:** 3.9.1

**Nikola Version:** 8.1.3

**Operating System:** macOS

### Description:

In Nikola 8.1.2, the `caps` typogrify filter was re-enabled. Unfortunately, consecutive capital letters in math mode (when enabled) also trigger the filter, resulting in broken equations. I think the simplest fix is to add the `.math` CSS class to the ignored tags here: https://github.com/getnikola/nikola/blob/c49f959a5b3ad7ee2c37491c68668bd441e0c918/nikola/filters.py#L272

Aside from that, in trying to debug this issue, I could not determine how to pass options to the `typogrify_custom` filter. `typogrify_custom` is the only filter wrapped by `apply_to_text_file` that has more than one argument, and it is not wrapped by the `_ConfigurableFilter` class to read an option from `conf.py`. In the end, I had to create a `functools.partial` in my `conf.py` that passed the list of `typogrify_filters` to apply as well as the modified `ignore_tags`.

If that's the intended interface, it would be nice to support a default list of applied filters, in case the user only wants to edit the `ignore_tags` (and vice-versa).

Please let me know if I should create a pull request with my suggested fix for the `.math` issue!

References:

https://github.com/Cantera/cantera-website/issues/128
https://github.com/getnikola/nikola/pull/3447/ (and linked issues)

bryanwweber:
### Pull Request Checklist

- [X] I tested my changes.

### Description

The styling of div elements with the math class is handled by the math
formatter (MathJax or katex), so typogrify should not interefere.

Fixes #3512.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,6 +7,12 @@ Features
 * New ``auto_command_starting`` signal when ``nikola auto`` is
   starting
 
+Bugfixes
+--------
+
+* Typogrify ignores ``div`` elements with ``.math`` CSS class.
+  (Issue #3512)
+
 New in v8.1.3
 =============
 
--- a/nikola/filters.py
+++ b/nikola/filters.py
@@ -269,7 +269,7 @@ def minify_lines(data):
 def _run_typogrify(data, typogrify_filters, ignore_tags=None):
     """Run typogrify with ignore support."""
     if ignore_tags is None:
-        ignore_tags = ["title"]
+        ignore_tags = ["title", ".math"]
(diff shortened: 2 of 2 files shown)


Kwpolska (approved):

Kwpolska:
Thanks for fixing this!

PR #3517, opened 2021-02-19
Add default list of filters for typogrify_custom

bryanwweber:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Per discussion on #3512, this implements a default list of filters for `typogrify_custom`. This allows specifying only the ignore_tags option. It also adds documentation for the `typogrify_custom` filter using `functools.partial`.

### DIFF
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -2301,23 +2301,25 @@ filters.typogrify_sans_widont
    Same as typogrify without the widont filter
 
 filters.typogrify_custom
-    Run typogrify with a custom set of filters or ignored tags. Takes ``typogrify_filters``
-    (a list of callables) or ``ignore_tags`` (a list of HTML tags to ignore). These must be
-    specified as keyword arguments to ``functools.partial``. The following code should be
+    Run typogrify with a custom set of filters or ignored HTML elements. Takes one or
+    both of the arguments ``typogrify_filters`` or ``ignore_tags``. ``typogrify_filters``
+    must be a list of typogrify filter callables to run. ``ignore_tags`` must be a list
+    of strings specifying HTML tags, CSS classes (prefixed with ``.``), tag ``id`` names
+    (prefixed with ``#``), or a tag and a class or id. The following code should be
     placed in ``conf.py``.
 
     .. code-block:: python
 
       from nikola.filters import typogrify_custom
-      from functools import partial as ft_partial
+      import functools
       # This filter will ignore HTML elements with the CSS class "typo-ignore"
       FILTERS = {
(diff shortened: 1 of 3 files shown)


docs/manual.rst
```diff
@@ -2301,7 +2301,31 @@ filters.typogrify_sans_widont
    Same as typogrify without the widont filter
 
 filters.typogrify_custom
-    Run typogrify with a custom set or filters. Takes ``typogrify_filters`` (a list of callables) and ``ignore_tags`` (defaults to None).
+    Run typogrify with a custom set of filters or ignored tags. Takes ``typogrify_filters``
+    (a list of callables) or ``ignore_tags`` (a list of HTML tags to ignore). These must be
```
Kwpolska, line 2305: > (a list of HTML tags to ignore)

Not quite, typogrify takes elements, classes (with .), IDs (with #), and constructions like element#id. This should be described more precisely.

docs/manual.rst
```diff
@@ -2301,7 +2301,31 @@ filters.typogrify_sans_widont
...
+    specified as keyword arguments to ``functools.partial``. The following code should be
+    placed in ``conf.py``.
+
+    .. code-block:: python
+
+      from nikola.filters import typogrify_custom
+      from functools import partial as ft_partial
```
Kwpolska, line 2312: Do we really need an `as` import here? Much easier to just `import functools`.

nikola/filters.py
```diff
@@ -268,8 +268,11 @@ def minify_lines(data):
 
 def _run_typogrify(data, typogrify_filters, ignore_tags=None):
     """Run typogrify with ignore support."""
+    default_ignore_tags = ["title", ".math"]
     if ignore_tags is None:
-        ignore_tags = ["title", ".math"]
+        ignore_tags = default_ignore_tags
+    else:
+        ignore_tags.extend(default_ignore_tags)
```
Kwpolska, line 275: I’d prefer not to mutate this list and instead use:

    ignore_tags = ignore_tags + default_ignore_tags

Kwpolska (changes requested):

docs/manual.rst
bryanwweber: I did this to reduce the line length of the example. Happy to change it back.

Kwpolska: Clarity is much more important than line length. I personally recommend avoiding `as` except for a few established aliases (`import numpy as np; import pandas as pd`) and avoiding naming conflicts (as we do with `typogrify`in `filters.py`).

Kwpolska (approved):

Kwpolska:
Thanks for fixing this!

Issue #3532, opened 2021-03-23
Drop legacy Python support


JeppeKlitgaard:
Currently the documentation isn't entirely consistent as to which Python versions are supported, and there seems to be some legacy Python 2 code lying around in the codebase.

What is the the main developers' feelings about legacy support? Python 3.6 is due to stop getting security fixes at the end of 2021, so perhaps bumping the minimum supported Python version to 3.7 already would make sense? This would allow for more/nicer type hinting in the code, which I can see has already been started in the `log.py` code.

Kwpolska:
Our current support policy is 3.5 or newer. Any Python 2 leftovers shouldn’t be major things that make the code worse.

Any changes to the support policy have to take the users into account. Our users have different relationships with Python — ranging from people for whom it’s just some dependency of Nikola, to contributors to [PyPy](https://github.com/pypy/pypy.org). And our users’ approach to Python versioning ranges from “whatever Ubuntu has in their repos” to “whatever was on CPython’s `master` branch at midnight”. We aim to support as many users and use-cases as possible, which means we need a broader range of versions. This means following Ubuntu LTS and Debian stable (and possibly looking at RHEL). Ubuntu 18.04 (old LTS) offers 3.6, 20.04 offers 3.8, Debian buster (stable) offers 3.7, RHEL seems to be at 3.6. There’s also PyPy, which is  This means that we won’t set the version to less than 3.6 for the time being, and we might stay with it even when CPython support for that version ends. We might get rid of 3.5 in the near future (when it becomes difficult to support, or if we get a PR that adds 3.6+ type annotations everywhere). But if there is no reason to upgrade to 3.6, I’d prefer to stay with 3.5.

PR #3534, opened 2021-03-23
Bump Python version to 3.7

JeppeKlitgaard:
### Pull Request Checklist

- [X] I tested my changes.

### Description

I _think_ this covers most of a bump to Python 3.7.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,7 @@ It has many features, but here are some of the nicer ones:
 * Syntax highlighting for almost any programming language or markup
 * Multilingual sites, `translated to 50 languages.`__
 * Doesn't reinvent wheels, leverages existing tools.
-* Python 3.4+ compatible.
+* Python 3.7+ compatible.
 
 .. _Nikola Handbook: https://getnikola.com/handbook.html#why-static
 __ https://users.getnikola.com/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: [3.5, 3.6, 3.7, 3.8, 3.9]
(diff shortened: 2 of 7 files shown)


Kwpolska (changes requested):
3.7 is unacceptable (see discussion in #3532). 3.6 might be possible if there are other improvements that it would bring.

README.rst
```diff
@@ -37,7 +37,7 @@ It has many features, but here are some of the nicer ones:
 * Syntax highlighting for almost any programming language or markup
 * Multilingual sites, `translated to 50 languages.`__
 * Doesn't reinvent wheels, leverages existing tools.
-* Python 3.4+ compatible.
```
Kwpolska, line 40: Fixed (to say 3.5) in commit https://github.com/getnikola/nikola/commit/8cc31e0230a7209cc0bca558a9d085b2e179fe8c.

JeppeKlitgaard:
Thanks for having a look at this @Kwpolska.

The reasoning behind 3.7 specifically was to get access to `__future__.annotations`, since I was hoping to expand on the type hinting in `Nikola`. Not knowing fully what the project's approach to legacy Python versions was, I thought it might be acceptable for the next major release, Nikola 9, to only support 3.7+, leaving 3.6 and 3.5 users 'stuck' on Nikola 8.

Kwpolska:
We can change our minimum version version to 3.6 in a 8.x.0 release (3.5 was actually dropped in 8.0.4, a patch release), Nikola 9 isn’t really planned for now. Type annotations can be done without `from __future__ import annotations`, the required wrapping in strings isn’t that bad. We’d happily take a PR with 3.6+ type annotations and changing the minimum Python version at the same time.

JeppeKlitgaard:
Abandoned for now

Issue #3542, opened 2021-04-14
Cactus comments


pieterdavid:
From https://cactus.chat :
> Cactus Comments is a federated comment system for the web, based on the Matrix protocol.

It is a very new, but rather interesting comments system: each comments section is a [Matrix](https://matrix.org) room, which allows posting with an account on any Matrix homeserver, or as a guest user. Both the homeserver (several implementations exist, see [here](https://matrix.org/docs/projects/try-matrix-now)) and the Cactus Comments application service (which manages the creation and moderation of rooms) can also be [self-hosted](https://cactus.chat/docs/self-hosting/introduction/).

I have the integration working locally, so I am also submitting a PR.

Kwpolska:
Hello, we are not accepting new comment systems for inclusion in Nikola, especially when the comment systems are brand new projects. Those systems end up being a maintenance burden for us when things change in their implementation, probably won’t have many users, and can randomly disappear.

carlbordum:
For anyone who stumbles upon this in the future, @pieterdavid made a very nice project that can help you with this. You can check it out here: [docs](https://cactus.chat/docs/integrations/nikola/), [project](https://github.com/pieterdavid/nikola-cactus-comments), [MR](https://gitlab.com/cactus-comments/cactus.chat/-/merge_requests/4).

Kwpolska:
(Also, if you have ideas about an alternative approach to allow comment systems to be installed from `plugins.getnikola.com` without needing to change core, we’ll happily discuss and take a PR for that — see #3544)

PR #3543, opened 2021-04-14
Add Cactus Comments as a comments system

pieterdavid:
Fixes https://github.com/getnikola/nikola/issues/3542

### Pull Request Checklist

- [x] I tested my changes: tests pass, and I tried the new system with the bootstrap4 and bootstrap4-ninja themes (it's also working on [my personal blog](https://pieterdavid.github.io/posts/welcome/#comment-section))

### Description

I tried to follow the same approach as the other comments systems as much as possible, but including the script and stylesheet in `comment_link_script` didn't work, so I added a new `comment_extra_head` macro that should be included in the `extra_head` block for templates to support this.
The `COMMENT_SYSTEM_ID` should the be site name passed to the Cactus Comments service; since the latter could be self-hosted I put that in `GLOBAL_CONTEXT["cactus_config"]`. They could also be put together in `COMMENT_SYSTEM_ID`, but then the template needs to split that again.
For the page ID (which becomes part of the Matrix room name) I used `utils.slugify(title)` because I didn't see an easy way to get the slug from the template.
I'm not very familiar with web development, the template engines, or the Nikola codebase, so please let me know if there are better solutions.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -101,6 +101,7 @@
 * `Paul Ivanov <https://github.com/ivanov>`_
 * `Pelle Nilsson <https://github.com/pellenilsson>`_
 * `phora <https://github.com/phora>`_
+* `Pieter David <https://github.com/pieterdavid>`_
 * `Pierpaolo Da Fieno <https://github.com/numshub>`_
 * `pmav99 <https://github.com/pmav99>`_
 * `Puneeth Chaganti <https://github.com/punchagan>`_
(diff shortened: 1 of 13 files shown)


Kwpolska:
Hello, we are not accepting new comment systems for inclusion in Nikola, especially when the comment systems are brand new projects. Those systems end up being a maintenance burden for us when things change in their implementation, probably won’t have many users, and can randomly disappear.

Issue #3544, opened 2021-04-14
Comment systems as plugins


Kwpolska:
There are sometimes requests to add new comment systems into Nikola (eg. #3542 or #3454). Those systems don’t meet our quality and popularity requirements (since adding a comment system ends up being a maintenance burden to support that system). It might be useful to come up with a way for plugins to somehow register extra comment systems for us.

(The existing systems should stay as templates.)

PR #3546, opened 2021-04-24
Allow adding comment systems from plugins

pieterdavid:
Closes https://github.com/getnikola/nikola/issues/3544

I had a look at how https://github.com/getnikola/nikola/issues/3544 could be implemented. Since all existing (and probably most future) comment systems implement the same interface with `comments_helper_<comment_system>.tmpl`, the key change is to load the file corresponding to the comment_system configuration option, rather than comparing to a fixed list.
Then it is sufficient for templates to provide `comments_helper_<system>.tmpl` for both template engines (see e.g. https://github.com/getnikola/plugins/pull/381), and the existing systems work without changes.
There are at least two things that need to be improved in this PR (hence the WIP status):
- the dependency detection code in the mako plugin cannot handle expressions in the "file" attribute (while it is supported for rendering). I put an if-statement that ignores such dependencies (with a warning), which seems to work, but is almost certainly not the proper solution.
- I added the slugify filter for jinja, which is an unrelated change (but could be useful, I can make it a separate PR if that's preferred), because adding it to TEMPLATE_FILTERS from the plugin seemed not to work (maybe I missed something)

Any help or feedback on these two points, or on the approach in general, would be much appreciated.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -12,6 +12,7 @@ Features
 * The default ``ignore_tags`` are appended to the user-supplied
   ``ignore_tags`` added via ``typogrify_custom``.
 * Allow adding comment systems from a plugin (Issue #3544)
+* New ``CommentSystem`` plugin category (Issue #3544)
 
 Bugfixes
 --------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,7 @@ Features
   as the only option.
 * The default ``ignore_tags`` are appended to the user-supplied
   ``ignore_tags`` added via ``typogrify_custom``.
+* Allow adding comment systems from a plugin
 
 Bugfixes
 --------
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,7 +11,7 @@ Features
   as the only option.
 * The default ``ignore_tags`` are appended to the user-supplied
   ``ignore_tags`` added via ``typogrify_custom``.
-* Allow adding comment systems from a plugin
+* Allow adding comment systems from a plugin (Issue #3544)
 
 Bugfixes
 --------
--- a/nikola/data/themes/base-jinja/templates/comments_helper.tmpl
+++ b/nikola/data/themes/base-jinja/templates/comments_helper.tmpl
@@ -1,63 +1,15 @@
 {#  -*- coding: utf-8 -*- #}
 
-{% import 'comments_helper_disqus.tmpl' as disqus with context %}
-{% import 'comments_helper_intensedebate.tmpl' as intensedebate with context %}
-{% import 'comments_helper_muut.tmpl' as muut with context %}
-{% import 'comments_helper_facebook.tmpl' as facebook with context %}
-{% import 'comments_helper_isso.tmpl' as isso with context %}
-{% import 'comments_helper_commento.tmpl' as commento with context %}
-{% import 'comments_helper_utterances.tmpl' as utterances with context %}
+{% import 'comments_helper_%s.tmpl' % comment_system as comment_system_impl with context %}
 
 {% macro comment_form(url, title, identifier) %}
-    {% if comment_system == 'disqus' %}
-        {{ disqus.comment_form(url, title, identifier) }}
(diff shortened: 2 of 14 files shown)


nikola/data/themes/base-jinja/templates/comments_helper.tmpl
```diff
@@ -1,63 +1,15 @@
 {#  -*- coding: utf-8 -*- #}
 
-{% import 'comments_helper_disqus.tmpl' as disqus with context %}
-{% import 'comments_helper_intensedebate.tmpl' as intensedebate with context %}
-{% import 'comments_helper_muut.tmpl' as muut with context %}
-{% import 'comments_helper_facebook.tmpl' as facebook with context %}
-{% import 'comments_helper_isso.tmpl' as isso with context %}
-{% import 'comments_helper_commento.tmpl' as commento with context %}
-{% import 'comments_helper_utterances.tmpl' as utterances with context %}
+{% import 'comments_helper_%s.tmpl' % comment_system as comment_system_impl with context %}
```
Kwpolska, line 3: This code is different in jinja and mako, which means that an attempt to update themes with `scripts/jinjify.py` will break it. Could you add an override to that script?

nikola/plugins/template/mako.py
```diff
@@ -61,7 +61,11 @@ def get_string_deps(self, text, filename=None):
         for n in lex.template.nodes:
             keyword = getattr(n, 'keyword', None)
             if keyword in ["inherit", "namespace"] or isinstance(n, parsetree.IncludeTag):
-                deps.append(n.attributes['file'])
+                if n.attributes["file"] == n.parsed_attributes["file"].strip("'"):
+                    deps.append(n.attributes['file'])
+                else:
+                    LOGGER.warning("Ignoring file=%s for template string dependencies in %s",
```
Kwpolska, line 67: Perhaps we could add a special-case for the comment system case to avoid dependency glitches and warnings?

CHANGES.txt
```diff
@@ -11,6 +11,7 @@ Features
   as the only option.
 * The default ``ignore_tags`` are appended to the user-supplied
   ``ignore_tags`` added via ``typogrify_custom``.
+* Allow adding comment systems from a plugin
```
Kwpolska, line 14: Please add a `(Issue #3544)` tag here.

Kwpolska (changes requested):
Looks pretty good! (Some minor changes are needed to get it fully working though.)

Kwpolska:
Regarding your questions:

1. We could special-case the comment system handlers,  parsing the injected name could be parsed for a more general-case but that’s not mandatory.
2. That change’s okay, we’ll merge it in with this PR. I’m not sure why `TEMPLATE_FILTERS` doesn’t work.

nikola/plugins/template/mako.py
pieterdavid: Good idea, I put a test that it's the comments_helper include and removed the warning

nikola/data/themes/base-jinja/templates/comments_helper.tmpl
pieterdavid: Done (maybe not the most efficient or concise, but it seems to work - jinjify doesn't change base-jinja anymore now)

pieterdavid:
Thanks of the review! I think addressed the comments.
Parsing the imported template name with an expression included would need a way to get the context variables while resolving the dependencies - that should be possible, but requires quite some changes to TemplatePlugin, so if it's acceptable like this I think that's easier.

nikola/data/themes/base-jinja/templates/comments_helper.tmpl
Kwpolska: A `dumb_replacement` would be cleaner than all those regexps IMO — I’ll move the change there.

Kwpolska (approved):

Kwpolska:
@pieterdavid Thanks for tackling this! I added a more complete dependency handling mechanism for both Mako and Jinja, and they can correctly detect changes in comment helpers.

PR #3562, opened 2021-07-13
Use --delete-after in suggested rsync commands

spookylukey:
Use of `--delete` will cause files to be deleted at the beginning of the rsync by default which causes files to disappear temporarily from the site.

This is bad in itself (you get 404s temporarily), and if anything goes wrong (network breaks during deploy) the 404s become more permanent. We can avoid this using `--delete-after`. I have discovered all this the hard way!

### Pull Request Checklist

- [X] I tested my changes.


Kwpolska (approved):

Kwpolska:
Thanks for contributinng this!

PR #3566, opened 2021-08-09
Fix #3555: Let plugin name be correctly accessed.

### LINKED ISSUE #3555: Plugin name not being correctly accessed
The name specified for a plugin seems not to be correctly accessed - it's not being read from the `plugin_info`'s `plugin_object`, as it ought to be. Instead of

https://github.com/getnikola/nikola/blob/650b954294ab01627ba9b45d81835293a5e02bb8/nikola/nikola.py#L1152

it should be something like

```python
          name = getattr(plugin_info.plugin_object, 'name', plugin_info.name)
          self._commands[name] = plugin_info.plugin_object
```

Similar logic needs to be applied for the activation, at

https://github.com/getnikola/nikola/blob/650b954294ab01627ba9b45d81835293a5e02bb8/nikola/nikola.py#L1321

which could be something like

```python
          name = getattr(plugin_info.plugin_object, 'name', plugin_info.name)
          self.plugin_manager.activatePluginByName(name)
```

vsajip:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Get the plugin name from the correct source.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -1149,7 +1149,8 @@ class Nikola(object):
         command_plugins = self._activate_plugins_of_category("Command")
         for plugin_info in command_plugins:
             plugin_info.plugin_object.short_help = plugin_info.description
-            self._commands[plugin_info.name] = plugin_info.plugin_object
+            name = getattr(plugin_info.plugin_object, 'name', plugin_info.name)
+            self._commands[name] = plugin_info.plugin_object
 
         self._activate_plugins_of_category("Task")
         self._activate_plugins_of_category("LateTask")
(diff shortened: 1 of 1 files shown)


Kwpolska:
Is this change backwards-compatible? Are any plugins (in core or in plugins.getnikola.com) broken by this change?

vsajip:
I ran the tests locally and they gave the same results with and without this change (some xfails) but I didn't specifically check with all plugins in plugins.getnikola.com. I came across this because without this change the existing code doesn't actually pick up the name from where it should, and so IMO behaves contrary to expectations.

PR #3567, opened 2021-08-09
Add/send a metadata signal.

vsajip:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Send a "metadata" signal to allow interested parties to make additions/changes to the metadata. One example use of this would be to set the "updated" time of a post to the actual modification time of the source file for the post.

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -1233,6 +1233,12 @@ def get_meta(post, lang):
         # Direct access because setter is complicated
         post._is_two_file = is_two_file
 
+    # There seems no way to get other metadata in a flexible way, so we add a signal
+    # here to hook into
+    signal('metadata').send({
+        'path': post.source_path,
+        'meta': meta
+    });
     return meta, used_extractor
 
 


Kwpolska:
I don’t like the approach of mutating the meta dictionary in a signal. Perhaps a `MetadataExtractor` plugin would suit your needs better?

In any case, this should be documented in at least `extending.rst`

vsajip:
> I don’t like the approach of mutating the meta dictionary in a signal. Perhaps a MetadataExtractor plugin would suit your needs better?

I looked at that approach first. For my specific use case (setting the "updated" metadata for any post to the modification time of the source file) I couldn't see a way of doing it using `MetadataExtractor`, because `extract_filename()` is only called with filenames for paths which match `FILE_METADATA_REGEXP`. Since you can only have one such regexp, using it to match all posts would mean it couldn't be used with e.g. an actual regexp.
https://github.com/getnikola/nikola/blob/650b954294ab01627ba9b45d81835293a5e02bb8/nikola/post.py#L1196-L1201

vsajip:
> In any case, this should be documented in at least extending.rst

If this functionality is going to be merged, I will happily update the documentation to cover it.

Issue #3630, opened 2022-06-21
Nikola (markdown plugin) does not work with Python 3.11 due to problem with `GIST_RST_RE` and markdown regex compilation


AdamWill:
### Environment

**Python Version:** 3.11

**Nikola Version:** 8.2.2 (and git master, this code has not changed)

**Operating System:** Fedora Rawhide

### Description:

Nikola will not work with Python 3.11. Here's a simplification of something `nikola/plugins/compile/markdown/mdx_gist.py` does:

    #!/bin/python

    import markdown
    import re

    GIST_RST_RE = r'(?m)^\.\.\s*gist::\s*(?P<gist_id>[^\]\s]+)(?:\s*(?P<filename>.+?))?\s*$'
    re.compile(r"^(.*?)%s(.*)$" % GIST_RST_RE)

The original code does `GistPattern(GIST_RST_RE)`, and `GistPattern` inherits from `markdown.inlinepatterns.Pattern`, and that class's `__init__()` does `re.compile(r"^(.*?)%s(.*)$" % pattern)` where `pattern` is the arg passed in.

If you run that test script on Python 3.10, you get a warning:

    DeprecationWarning: Flags not at the start of the expression '^(.*?)(?m)^\\.\\.\\s*gi' (truncated) but at position 6

On Python 3.11, this warning has been turned into a fatal error:

    E                               re.error: global flags not at the start of the expression at position 6
    /usr/lib64/python3.11/re/_parser.py:841: error

The problematic bit of `GIST_RST_RE` is the `(?m)` at the start, which is trying to set multiline mode. If you drop that the error goes away, but of course I'm not sure if that is the appropriate fix or would cause other problems. I can't offhand see an easy way to fix this while preserving the multiline mode setting though.

AdamWill:
OK, yeah, I think just inheriting from `InlineProcessor` instead should work. Testing that now.

PR #3631, opened 2022-06-21
mdx_gist: inherit from InlineProcessor, work on Python 3.11 (#3630)

AdamWill:
The gist-rst pattern crashes Python 3.11, because its pattern
tries to set the multiline flag - it starts with `(?m)` - but
Markdown's `Pattern` class `__init__`, which is called by our
`GistPattern.__init__`, 'wraps' the pattern with additional
capture groups before compiling it. This causes the multiline
flag not to be at the start of the 'wrapped' pattern. From
Python 3.6 to Python 3.10 flags not being at the start of the
pattern was deprecated and triggered a warning when trying to
compile it; in Python 3.11 it's an error.

Markdown 3.0.0 added a preferred `InlineProcessor` class:
https://github.com/Python-Markdown/markdown/pull/629
which does not do this wrapping. It requires the subclass to
implement `handleMatch`, which we already do. Our `handleMatch`
uses named capture groups, so the change in the number of
capture groups in the compiled expression shouldn't matter. So
simply switching to inheriting from `InlineProcessor` instead
of `Pattern` should solve the problem.

We already required Markdown 3.0.0, so this does not mean we
require a newer Markdown than before.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### Pull Request Checklist

- [-] I updated AUTHORS.txt and CHANGES.txt (if the change is non-trivial) and documentation (if applicable).
- [x] I tested my changes.

### Description

Change seems fairly trivial and should not be visible to users, so I didn't bother with AUTHORS and CHANGES. I've 'tested' this insofar as nikola now at least passes its own test suite with Python 3.11 (it did not before); the test suite does not seem to cover this file, so I don't think we'll know if it's broken, but I don't think it should be.

### DIFF
--- a/nikola/plugins/compile/markdown/mdx_gist.py
+++ b/nikola/plugins/compile/markdown/mdx_gist.py
@@ -140,10 +140,8 @@ class GistPattern(InlineProcessor):
         return resp.text
 
     def handleMatch(self, m, _):
-        """
-        Handle pattern match. The third arg is "data", wider
-        context around the match; we don't need it.
-        """
+        """Handle pattern match."""
+        # The third arg is "data", wider context around the match; we don't need it.
         gist_id = m.group('gist_id')
         gist_file = m.group('filename')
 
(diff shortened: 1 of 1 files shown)


Kwpolska (approved):

Kwpolska:
Thanks for contributing this fix!

PR #3637, opened 2022-07-22
docs: fix simple typo, begining -> beginning

timgates42:
There is a small typo in nikola/plugins/command/import_wordpress.py.

Should read `beginning` rather than `begining`.

Semi-automated pull request generated by [advertisement removed].

### DIFF
--- a/nikola/plugins/command/import_wordpress.py
+++ b/nikola/plugins/command/import_wordpress.py
@@ -1189,7 +1189,7 @@ def separate_qtranslate_tagged_langs(text):
             if not c.strip():
                 continue
         elif c[2:qt_end_with_lang_len].startswith(qt_end):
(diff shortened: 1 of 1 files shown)


Kwpolska:
Advertising isn't welcome here. I also don't think semi-auttomated pull requests with trivial typo fixes are a good idea, leave it to the humans.

timgates42:
Yeah this repository seems to be doing very well with only one error found, a lot of other repositories don’t do nearly as well. I respect your position but over all these fixes are welcomed, if I did generally get negative feedback then I would stop so I’m sorry this PR upset you - I can add getnikola repositories to a deny list if you would like to prevent any future fixes?

Kwpolska:
I’d prefer you didn’t spam GitHub with minor typo fixes in unimportant comments.

PR #3673, opened 2023-03-26
Fix some deprecation warnings generated during the test suite

tartley:
### Pull Request Checklist

- [x] I tested my changes.

### Description

I added a trivial item to .gitignore. The 'tags' file is produced by utilities like 'ctags' (incl. the venerable exhuberant-ctags, and the modern universal-ctags), and is used by programs like Vim for 'go to definition' functionality.

I fixed two deprecation warnings that were generated when I ran the test suite, by making the substitutions suggested by the deprecation warnings themselves. Tests still pass, and the relevant functionality (RST to HTML and antialiasing in thumbnails) still seems to work in a freshly-built site.

The test suite does still generate two deprecation warnings, but they are from code in a dependency of ours (yapsy), and we are using the latest version of that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 tmp/
 cache/
 output/
-tags
+/tags
 
 # GitHub token
 .pypt/gh-token
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 tmp/
 cache/
 output/
+tags
 
 # GitHub token
 .pypt/gh-token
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@
 tmp/
 cache/
 output/
+tags
```
Kwpolska, line 8: Is this a file or a directory? Is it placed in the repository root (then it should be `/tags`) or in some other static location?

tartley: Ah, of course, silly me. Thank you for helping me get it right.

Kwpolska (approved):

Issue #3671, opened 2023-03-23
WebP images in a gallery cause an unhandled exception during build


tartley:
## Environment

**Python Version:**: Python3.10.6  (and 3.11.0rc1 behaves the same)
**Nikola Version:** v8.2.3 (and latest git clone of commit 483ffd8f Wed Mar 1 22:52:02 2023 behaves the same)
**Operating System:** Linux Pop\_OS 22.04 LTS (derived from Ubuntu 22.04)

## Description:

Hi, good people! I am happy to look into contributing a PR to fix this myself, but wanted to discuss it first before wading in. Thanks!

### Symptom

Adding a WebP image (with '.webp' extension) to my gallery causes `nikola build` to fail partway through, with a traceback (below). This happens even on a fresh site, created using `nikola init -dq site`.

After the failure, running `nikola serve` displays the gallery correctly, including all the WebP images, but the gallery RSS feed is empty. Because this is so close to working, I'm hopeful that a small fix might be possible.

I notice that galleries with an empty file named "image.webp" provoke this error, while gallery files named "image", with no extension, or "image.nonsense", are correctly ignored. So I infer that webp images *are* expected to work in galleries. (and the section below, "*Any other image types we should handle?*", seems to corroborate.)

The expected behavior is:

* WebP images ought to "just work", both at site generation time, and at gallery viewing time, along with their gallery RSS feed.
* For any gallery image file with an unrecognized mime type, options for how we should behave are enumerated below in section "*Be robust to files with unrecognized mime type*".

### The Traceback

```
$ nikola build
Scanning posts........done!
.  render_galleries:output/galleries/demo/sp.webp
.  render_galleries:output/galleries/demo/index.html
.  render_galleries:output/galleries/demo/rss.xml
TaskError - taskid:render_galleries:output/galleries/demo/rss.xml
PythonAction Error
Traceback (most recent call last):
  File "/home/jhartley/.virtualenvs/nikola/lib/python3.10/site-packages/doit/action.py", line 461, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/home/jhartley/Dropbox/jon/data/code/3rdparty/nikola/nikola/nikola/plugins/task/galleries.py", line 773, in gallery_rss
    data = rss_obj.to_xml(encoding='utf-8')
  File "/home/jhartley/.virtualenvs/nikola/lib/python3.10/site-packages/PyRSS2Gen.py", line 39, in to_xml
    self.write_xml(f, encoding)
[...]

Issue #3672, opened 2023-03-24
Testing supported image formats in a gallery


tartley:
## Environment

**Python Version:** 3.10.6
**Nikola Version:** git latest (commit 483ffd8f Wed Mar 1 22:52:02 2023)
**Operating System:** Linux Pop!_OS 22.04 LTS (Derived from Ubuntu 22.04)
**Browser:** Firefox 110.0.

## Description:

I tested all the supported image formats in galleries. I plan to submit some fixes for the problems I found, but I know nothing, so let's discuss here & correct my misunderstandings first.

I infer the supported image file name extensions are defined at [nikola/image_processing.py:46](https://github.com/getnikola/nikola/blob/master/nikola/image_processing.py#L46), which equates to:

    bmp, gif, jpeg, jpg, png, svg, svgz, tiff, webp

With each of these image types in a gallery, I tested:

* Run `nikola build`
* Preview the image in a gallery
* Click on the preview to view the full size image in the default viewer, BaguetteBox.

Also, I record some diagnostic information:

* Does bagetteBox claim to suport this image type?
* Does Python's `mimetype.guess_type()` know the type for this image filename extension?

Extension | Build<br/>site | Image<br/>preview | View<br/>image | baguetteBox<br/>support | `.guess_type(x)`
----------|:-----:|:-------:|:--------:|:-----:|:--------------
.bmp      |  y    |  y      | KINDA[4] | NO,BUT[5] | ('image/bmp', None)
.gif      |  y    |  y      |  y       |  y        | ('image/gif', None)
.jpeg     |  y    |  y      |  y       |  y        | ('image/jpeg', None)
.jpg      |  y    |  y      |  y       |  y        | ('image/jpeg', None)
.png      |  y    |  y      |  y       |  y        | ('image/png', None)
.svg      |  y    |  y      | KINDA[4] | NO,BUT[5] | ('image/svg+xml', None)
.svgz     |  y    | NO[2]   | KINDA[4] | NO,BUT[5] | ('image/svg+xml', 'gzip')
.tiff     |  y    | NO[3]   | NO[3]    | NO,BUT[5] | ('image/tiff', None)
.webp     | NO[1] |  y      |  y       |  y        | (None, None) [1]

* NO[1] Builds with webp gallery images fail, as described in [issue #3671](https://github.com/getnikola/nikola/issues/3671).
[...]

PR #3674, opened 2023-03-26
Fix exception from gallery RSS with WebP images

tartley:
### Pull Request Checklist

- [x] I tested my changes.

### Description

WebP images in a gallery cause `nikola build` to fail with an unhandled AttributeError exception, as described in #3671. This happens because the MIME type of WebP images is not known, and the resulting None value causes the RSS generation code to choke.

This PR fixes that, by augmenting Python's list of known mimetypes with a hardcoded WebP one, before we ask for the mimetype of each image in a gallery.

I don't think the content of the generated gallery RSS feed is tested yet. In the issue comments, I talk about adding a new test for that. But instead, I noticed that converting one of the images in the samplesite demo gallery into a WebP causes the site build to fail during the test suite, due to the above problem, causing many test failures. Adding this fix to the galleries.py file fixes all those tests again.

Also, running a manual build on a site containing webp galleries now works.

This addresses the above issue's checklist item "*Inject a hardcoded 'image/web' mimetype at runtime, before .guess_type() gets called.*", but does not complete the issue.


tartley:
It belatedly crosses my mind that perhaps, if the samplesite demo gallery is used throughout the automated tests, then perhaps its gallery ought to contain not just a placeholder WebP, but one example of every supported image type (jpg, webp, png, etc). This would help the automated tests to surface any problems in a highly-visible way (e.g. including tiff images here is desirable, so we can assert they are correctly handled)

However, the samplesite is not just test data. It is also prominently user-visible in many places. This is better suited to a demo that shows things working (e.g. tiff images here are a bad idea, many users would not be able to see them, depending on which browser they are using.)

I realize I'm a newcomer to this project (although have been using it happily for years), so I feel foolish making disruptive suggestions, but:

I would consider making the samplesite a separate thing from the test data. One can be optimized to look good, the other to be fragile. Initially, I would simply copy the samplesite into nikola/tests/data or somesuch, then modify the test data gallery to contain one of each image type. Presumably this would have exposed the WebP issue a long time ago?

I'm happy to submit a PR. Thoughts welcome! :-)

nikola/plugins/task/galleries.py
```diff
@@ -725,6 +725,9 @@ def gallery_rss(self, img_list, dest_img_list, img_titles, lang, permalink, outp
         def make_url(url):
             return urljoin(self.site.config['BASE_URL'], url.lstrip('/'))
 
+        # WebP is not yet an official MIME type but we need to recognize them
+        mimetypes.add_type('image/webp', '.webp')
```
Kwpolska, line 729: I’m not sure if this is the best placce for this, considering the enclosure feature may also trigger an exception. Perhaps a better place would be somewhere in `nikola.py`, in `Nikola.__init__` so that the mimetype is always defined, even if we are not rendering a gallery RSS file.

nikola/data/samplesite/galleries/demo/tesla_conducts_lg.webp
Kwpolska, line 1: Tangent, but wasn’t webp supposed to be more efficient than jpg? Or is this webp compressed in lossless mode?

Kwpolska:
> However, the samplesite is not just test data. It is also prominently user-visible in many places. These uses are better suited to a demo that shows things working (e.g. TIFF images here would be a bad idea, many users would not be able to see them, depending on which browser they are using.)

> I would consider making the samplesite a separate thing from the test data. One can be optimized to look good, the other to be fragile. Initially, I would simply copy the samplesite into nikola/tests/data or somesuch, then modify the test data gallery to contain one of each image type.

I think this would add more complexity and maintenance burden than it’s worth. We might end up making a change that works with the test site, but breaks the demo site. Converting a sample image to PNG/GIF, and perhaps adding some simple SVG would be beneficial for both the demo site and tests.

nikola/data/samplesite/galleries/demo/tesla_conducts_lg.webp
tartley: Once again, thank you for catching my silly mistake. It was late! :-)

Yes, you are right that webp gives a smaller file size for the same visual quality, but this file was bigger, which I hadn't spotted. It turns out, it wasn't that I'd saved it in lossless mode, it was just that the webp, since it was derived from a noisy original jpg, was attempting to recreate all the detail of the jpg artifacts (which for jpg, that details comes for free, as an unwanted side effect of the compression). So, webp will sometimes only be better in a  fair race, where both formats are created from the same uncompressed original.

Regardless, increasing the file size seemed annoying and distracting, so I resaved it at a lower webp quality setting. Whether this is better or worse looking than the original jpg seems debatable, but for these very low quality images, I don't think it matters.

I force pushed this change, so that your repo isn't forever sullied with the needlessly larger webp file.

tartley:
Be aware I force pushed this branch at this point, as mentioned in one of the above comments.

Kwpolska:
Force pushes are fine. Can you try placing the mimetype registration in a more global place, as suggested in the review comment?

nikola/plugins/task/galleries.py
tartley: Hi. I can do this. Am AFK now but will try later.

tartley: This is done.

tartley:
> Converting a sample image to PNG/GIF, and perhaps adding some simple SVG would be beneficial for both the demo site and tests.

I added this as a new checklist item to #3672

tartley:
As far as I know, this PR is ready, and completes checklist item "*Inject a hardcoded 'image/web' mimetype at runtime, before .guess_type() gets called*" on #3671. Thank you for your attention. Sorry it's been so wordy for such a small change.

tartley:
Let me know if the baseline test failure is something I can help with (the error says "maintainers only", so I'm assuming you want to be hands-on yourselves. Thanks!

Kwpolska (approved):

PR #3677, opened 2023-04-02
Make rebuild/reload act only when file is modified.

### LINKED ISSUE #3638: nikola auto loops infinitely
### Environment

**Python Version:** 3.7.13

**Nikola Version:** 8.2.3

**Operating System:** Debian

### Description:

When I run `nikola auto` for my site and load the local url in the browser, sometimes after making some changes to a blog post the site will reload over and over again.

In the terminal where `nikola auto` is running I can see it refreshing a lot of the posts/taxonomies and site xml files. I end up having to interrupt the process, restart it, and reload my browser.

Let me know if there's anything I can do to help debug this!

gipi:
Improve the event handler using directly the watchdog library's own handlers and trigger the actions only when a file is modified.

Should fix #3638

### Pull Request Checklist

- [x] I tested my changes.

### Description

The original implementation triggered rebuild and reload on each possible filesystem event that is time consuming and can cause infinite loops, it's probably better to limit the rebuilding on file modification.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,7 @@ Features
 Bug Fixes
 ---------
 
+* Fix ``auto`` command infinite loop (Issue #3677)
 * Fix API URL in CSS and JS minifiers (Issue #3658)
 * Fix ``:align: center`` for images in reST (Issue #3657)
 * ``GZIP_COMMAND`` parsing on ``win32`` platforms (Issue #3649)
--- a/nikola/plugins/command/auto/__init__.py
+++ b/nikola/plugins/command/auto/__init__.py
@@ -587,13 +587,11 @@ class NikolaEventHandler:
         self.function = function
         self.loop = loop
 
-    async def on_any_event(self, event):
-        """Handle all file events."""
-        await self.function(event)
-
     def dispatch(self, event):
         """Dispatch events to handler."""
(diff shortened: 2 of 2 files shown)


gipi:
I'll complete all the steps in the checklist once I'm sure the change is ok, maybe I could add some tests.

Kwpolska:
I was able to reproduce this issue, but only with the latest version of watchdog. The issue seems to be caused by file open and close events (which weren’t raised before, and in fact, there is no separate method for file open events).

The `on_modified` event is not enough for us, as we also want to trigger rebuilds when files are created or deleted. I think we should keep the `on_any_event` handler, but filter out the opened/closed events. I think a better change would be to keep overriding `dispatch`, but instead filter the event types:

```py
def dispatch(self, event):
    """Dispatch events to handler."""
    if event.event_type in {"opened", "closed"}:
        return
    self.loop.call_soon_threadsafe(asyncio.ensure_future, self.function(event))
````

Can you check if this fixes things for you as well? If yes, please change your pull request to the above.

gipi:
I changed the patch as indicated and it seems to fix my issue.

gipi:
Probably ``on_any_event()`` should be removed then.

Kwpolska (approved):

Issue #3700, opened 2023-07-14
nikola does not work with current git yapsy - multiple tests fail with `KeyError` in the plugin template injection code


AdamWill:
### Environment

**Python Version:** 3.12

**Nikola Version:** 8.2.4

**Operating System:** Fedora Rawhide

### Description:

We're landing Python 3.12 in Fedora Rawhide. As part of this, I had to backport https://github.com/tibonihoo/yapsy/pull/11 to make yapsy use importlib instead of imp. But that actually breaks nikola's template injection, because of an interesting change to what winds up in `sys.modules` with the different way of importing plugin modules. The errors look like this:

    self = <mdx_gist.GistExtension object at 0x7fded7e16420>

        def inject_templates(self):
            """Inject 'templates/<engine>' (if exists) very early in the theme chain."""
            try:
                # Sorry, found no other way to get this
    >           mod_path = sys.modules[self.__class__.__module__].__file__
    E           KeyError: 'mdx_gist'

    nikola/plugin_categories.py:79: KeyError

I have a fix for this ready to submit; I'm just filing an issue because the guidelines say there must be an issue for the PR.

PR #3701, opened 2023-07-14
Handle change to plugin loading in recent yapsy (#3700)

AdamWill:
https://github.com/tibonihoo/yapsy/pull/11 changes yapsy plugin loading to not use the deprecated imp module any more. However, as a side effect of that, it breaks this already-kinda-ugly hack, and we have to make it even uglier!

yapsy used to import the module like this:

imp.load_module(plugin_module_name,plugin_file...)

where `plugin_module_name` was the modified, "unique" name it creates early in `loadPlugins`. Interestingly, when you import a module like that, it gets added to `sys.modules` under *both* the modified name and its 'real' name, viz:

>>> import sys
>>> import imp
>>> imp.load_module("someothername", None, "/usr/lib/python3.12/site-packages/yapsy/__init__.py", ("py", "r", imp.PKG_DIRECTORY))
<module 'someothername' from '/usr/lib/python3.12/site-packages/yapsy/__init__.py'>
>>> sys.modules["someothername"]
<module 'someothername' from '/usr/lib/python3.12/site-packages/yapsy/__init__.py'>
>>> sys.modules["yapsy"]
<module 'yapsy' from '/usr/lib/python3.12/site-packages/yapsy/__init__.py'>

That's why this hack worked. However, now yapsy imports the module using importlib, then adds it to `sys.modules` itself, *only* under the modified "unique" name, not under its original name. So sys.modules["unmodifiedpluginname"] is now a KeyError.

I can't think of a less ugly fix than this, unfortunately. We *could* try sending a patch for yapsy to add it under both the modified and unmodified names, but that would be somewhat tricky in yapsy's design, and I also suspect yapsy would consider it to actually be unwanted behavior.

Maybe what we really need is to send a patch for yapsy to just provide an interface to find a plugin's filesystem path...

### Pull Request Checklist

- [X] I tested my changes.

### DIFF
--- a/nikola/plugin_categories.py
+++ b/nikola/plugin_categories.py
@@ -75,8 +75,31 @@ class BasePlugin(IPlugin):
     def inject_templates(self):
         """Inject 'templates/<engine>' (if exists) very early in the theme chain."""
         try:
+            mod_candidate = None
+            # since https://github.com/tibonihoo/yapsy/pull/11 ,
+            # yapsy only adds each imported plugin to sys.modules
+            # under its modified, "unique" name (see early in
+            # PluginManager.loadPlugins), so we recreate the
+            # modified name here to find it. we fudge the serial
+            # number here, assuming that if a plugin is loaded
+            # under the same name multiple times, the location
+            # will also be the same, so we can just use 0.
+            possible_names = (
+                self.__class__.__module__,
+                "yapsy_loaded_plugin_" + self.__class__.__module__ + "_0",
+                "yapsy_loaded_plugin_" + self.name + "_0",
+            )
+            for possible_name in possible_names:
+                mod_candidate = sys.modules.get(possible_name)
+                if mod_candidate:
+                    break
+            if not mod_candidate:
+                # well, we tried. we wind up here for the dummy
+                # plugins; honestly I'm not sure exactly why/how,
+                # but they don't have templates, so it's okay
+                return
             # Sorry, found no other way to get this
-            mod_path = sys.modules[self.__class__.__module__].__file__
+            mod_path = mod_candidate.__file__
             mod_dir = os.path.dirname(mod_path)
             tmpl_dir = os.path.join(
                 mod_dir, 'templates', self.site.template_system.name


AdamWill:
Revised to try two different possible 'unique' module names, since I found just doing one or the other wasn't enough in a real-world test with a real-world plugin...

AdamWill:
I'm not actually sure which of the three possible names is 'most preferred' and which is 'least preferred', we can tweak it if anyone has ideas.

nikola/plugin_categories.py
```diff
@@ -75,8 +75,30 @@ def set_site(self, site):
     def inject_templates(self):
         """Inject 'templates/<engine>' (if exists) very early in the theme chain."""
         try:
+            modname = ""
+            possmodnames = (
```
Kwpolska, line 79: Consider using longer, more descriptive names (`module_name`, `possible_names`).

nikola/plugin_categories.py
```diff
@@ -75,8 +75,30 @@ def set_site(self, site):
...
+                # modified name here to find it. we fudge the serial
+                # number here, assuming that if a plugin is loaded
+                # under the same name multiple times, the location
+                # will also be the same, so we can just use 0.
+                "yapsy_loaded_plugin_" + self.__class__.__module__ + "_0",
+                "yapsy_loaded_plugin_" + self.name + "_0",
+            )
+            for possmodname in possmodnames:
+                if possmodname in sys.modules:
+                    modname = possmodname
```
Kwpolska, line 94: Consider one less dictionary lookup:
```py
module_candidate = None
for possible_name in possible_names:
    module_candidate = sys.modules.get(possible_name)
    if module_candidate is not None:
        break
if module_candidate is None:
    return
```

Kwpolska (changes requested):

Kwpolska:
Adding paths to yapsy would be preferable, but if we can still get the module ourselves in some way, that’s good enough.

nikola/plugin_categories.py
AdamWill: But so much typing! 😆 sure.

nikola/plugin_categories.py
AdamWill: Yeah, I agree, I did the change to a loop quickly and it could be more elegant. I'll tweak it.

AdamWill:
I'll have a quick look at how viable it would be to add this to yapsy at some point.

AdamWill:
Revised. I used `mod_` not `module_` because there is already `mod_path` and `mod_dir`.

Kwpolska (approved):

Kwpolska:
Thanks for fixing this! I’ll merge this, and if there’s any improvement in yapsy, let us know (or send another PR 😄)

AdamWill:
that was quick! I think it *may* be possible to do this a 'better' way using yaspy's ability to set a callback for `loadPlugins`. I'm playing with that ATM. If it works out I'll send a follow-up PR.

PR #3702, opened 2023-07-15
Use callback to set plugin module path for template location

AdamWill:
This is a much cleaner (I think) approach to the problem of figuring out the plugin's template path (see #3701). We use yaspy's ability to provide a callback when loading the plugins. The callback is passed a `yaspy.PluginInfo` instance, which conveniently contains a reference to the plugin object itself, and its filesystem path. So we can just add a method on our plugin class to set the filesystem path as an attribute, and have the callback function call that method.

### Pull Request Checklist

- [X] I tested my changes.


AdamWill:
This worked first time (bar persnickety linters). That always makes me suspicious, but I *think* it's actually okay! I tested it in the 'real world' (on my actual blog, which uses the `static_comments` plugin, which is affected by the underlying problem here) and it works, and the test suite passes.

AdamWill:
bumped the yapsy requirement to 1.12.0, which seems to be the first version with `callback_after` in it.

nikola/plugin_categories.py
```diff
@@ -72,45 +71,23 @@ def set_site(self, site):
...
                 mod_dir, 'templates', self.site.template_system.name
             )
             if os.path.isdir(tmpl_dir):
                 # Inject tmpl_dir low in the theme chain
                 self.site.template_system.inject_directory(tmpl_dir)
         except AttributeError:
-            # In some cases, __builtin__ becomes the module of a plugin.
-            # We couldn’t reproduce that, and really find the reason for this,
-            # so let’s just ignore it and be done with it.
+            # Just in case something went wrong and we don't have
```
Kwpolska, line 89: 1. Does this ever happen with the base site + some plugins that need templates?
2. Please log a message (warning level?) to let users know if something explodes.

AdamWill: Good question!

It doesn't happen at least on a test run on my blog (if I change the `pass` to `raise`). But it seems to me like there's no harm in being safe; if this ever *does* happen, it seems better to try accepting it and moving on than just blow up. I think in the worst case, we'll fail later when a required template isn't available (this is what happens in my 'real world' test when I get something wrong).

Adding a log message is a good idea, though - it doesn't seem like something that should *usually* happen, at least, so logging at warning level seems justified. I'll do that.

AdamWill:
Resolved. Note the log message assumes the plugin will always have `self.name` - I figured this is safe as other code seems to do the same, but if not, we can make that safer.

Kwpolska (approved):

Kwpolska:
Thanks for the fix!

PR #3720, opened 2023-11-14
Fix chdir leaking on error out of cd-contextmanager of tests/helper.py.

aknrdureegaesr:
### Pull Request Checklist

- [X] I tested my changes.

### Description

While working on something else, I happened to spot a bug in `tests/helper.py`: The change of directory can (unintentionally) leak out of the `cd` context manager if the code inside the context throws an exception. In other words, on error, the change of directory is not undone.

I added a test for the test helper that demonstrates the bug if run with the previous test helper code. Not sure whether you want that.

This is a rather trivial change not visible to the end-user, so there is no need to add to CHANGELOG.

I plead guilty not having created a separate bug issue first. Do you need it in such a trivial case?

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -30,9 +30,11 @@ __all__ = ["cd", "FakeSite"]
 @contextmanager
 def cd(path):
     old_dir = os.getcwd()
-    os.chdir(path)
-    yield
-    os.chdir(old_dir)
+    try:
+        os.chdir(path)
+        yield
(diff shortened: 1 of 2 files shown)


Kwpolska (approved):

Kwpolska:
> I plead guilty not having created a separate bug issue first. Do you need it in such a trivial case?

Not really.

Thanks for fixing this!

PR #3722, opened 2024-01-01
Introduce a test for the development server. Fix #3718.

### LINKED ISSUE #3718: Add an automated test for plugin/commands/auto (the development server).

In an experiment, I sabotaged the development server and ran the tests. To my dismay, no test failed. This feature request requests coming up with a test for the development server.

In detail, here is what I did:

- I turned the crucial line `webapp.router.register_resource(resource)` in the file `plugins/command/auto/__init.py__` into a comment.
- I ran `pytest tests/` as recommended by the contributing guidelines.

Expectation: Some tests fail.

Actually seen: The test was happy.

I double-checked that removing that line in an actual Nikola deployment causes the development server to fail miserably; it does not serve the document root (404 HTTP Code).

This feature request is not suggesting to change any production code or functionality, but only add material to `tests`.

- [ ] Yes, I have already written code for it (link if available and feasible)
- [X] Yes, I don’t have code ready yet (that’s okay!)
- [ ] No (that’s okay too!)

I have done stuff like that before. My suggestion / plan is:

* Add some directories with a file each that the server can serve (statically to the git repo).
* At runtime, open a server port at 127.0.0.1, fetch the port's number, and close it immediately. This yields a port number that is presently unused.
* Configure the server to use that port and start it.
* Fire a few HTTP requests from the test code against that port and see the answers are as expected.

No.

This is somewhat related to #3715. Maybe I'll provide a pull request for that one, too (still have not promised). I have a profound dislike regarding providing a pull request without a test.

aknrdureegaesr:
### Pull Request Checklist

- [X] I tested my changes.

### Description

This adds a test for the development server functionality.

### DIFF
--- a/tests/integration/test_dev_server.py
+++ b/tests/integration/test_dev_server.py
@@ -1,0 +1,150 @@
+import asyncio
+import nikola.plugins.command.auto as auto
+from nikola.utils import get_logger
+import pytest
+import pathlib
+import socket
+from typing import Optional
+import requests
+
+from ..helper import FakeSite
+
+SERVER_ADDRESS = "localhost"
+TEST_MAX_DURATION = 10  # Watchdog: Give up the test if it did not succeed during this time span.
+
+# Folder that has the fixture file we expect the server to serve:
+OUTPUT_FOLDER = pathlib.Path(__file__).parent.parent / "data" / "dev_server_sample_output_folder"
+
+LOGGER = get_logger("test_dev_server")
+
+
+def find_unused_port() -> int:
+    """Ask the OS for a currently unused port number.
+
+    (More precisely, a port that can be used for a TCP server servicing SERVER_ADDRESS.)
+    We use a method here rather than a fixture to minimize side effects of failing tests.
+    """
+    s = socket.socket()
+    try:
+        ANY_PORT = 0
+        s.bind((SERVER_ADDRESS, ANY_PORT))
+        address, port = s.getsockname()
+        LOGGER.info("Trying to set up dev server on http://%s:%i/", address, port)
+        return port
+    finally:
+        s.close()
+
+
+def test_serves_root_dir(site, expected_text) -> None:
+    command_auto = auto.CommandAuto()
+    command_auto.set_site(site)
+    options = {
+        "browser": False,
+        "ipv6": False,
+        "address": SERVER_ADDRESS,
+        "port": find_unused_port(),
+        "db-file": "/dev/null",
+        "backend": "No backend",
+        "no-server": False
+    }
+
+    # We start an event loop, run the test in an executor,
+    # and wait for the event loop to terminate.
+    # These variables help to transport the test result to
+    # the main thread outside the event loop:
+    test_was_successful = False
+    test_problem_description = "Async test setup apparently broken"
+    test_inner_error: Optional[BaseException] = None
+
+    async def grab_loop_and_run_test() -> None:
+        loop = asyncio.get_running_loop()
+        watchdog_handle = loop.call_later(TEST_MAX_DURATION, lambda: loop.stop())
+        nonlocal test_problem_description
+        test_problem_description = f"Test did not complete within {TEST_MAX_DURATION} seconds."
+
+        def run_test() -> None:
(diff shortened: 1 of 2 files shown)


tests/integration/test_dev_server.py
```diff
@@ -0,0 +1,156 @@
...
+        nonlocal test_problem_description
+        test_problem_description = f"Test did not complete within {TEST_MAX_DURATION} seconds."
+
+        def run_test() -> None:
+            nonlocal test_was_successful, test_problem_description, test_inner_error
+            try:
+                server_base_uri = f"http://{options['address']}:{options['port']}/"
+
+                # First subtest: Does the dev server serve the fixture HTML file as expected?
+                with urllib.request.urlopen(server_base_uri) as res:
```
Kwpolska, line 72: You can (should?) use `requests` instead of `urllib.request` for cleaner and clearer code. (`requests` is a mandatory dependency of Nikola.)

tests/integration/test_dev_server.py
```diff
@@ -0,0 +1,156 @@
...
+import urllib.request
+
+from ..helper import FakeSite
+
+SERVER_ADDRESS = "localhost"
+TEST_MAX_DURATION = 10  # Watchdog: Give up the test if it did not succeed during this time span.
+SERVER_STARTUP_TIME = 0.1  # Max time, in seconds, we expect the server to need to come up.
+
+# Folder that has the fixture file we expect the server to serve:
+OUTPUT_FOLDER = pathlib.Path.cwd() / "tests" / "data" / "dev_server_sample_output_folder"
```
Kwpolska, line 17: Depending on the current working directory is flaky and may break. Instead, start with `__file__` (path to the current `.py` file) and find the folder based on that.

tests/integration/test_dev_server.py
```diff
@@ -0,0 +1,156 @@
...
+import pathlib
+import socket
+from typing import Optional
+import urllib.request
+
+from ..helper import FakeSite
+
+SERVER_ADDRESS = "localhost"
+TEST_MAX_DURATION = 10  # Watchdog: Give up the test if it did not succeed during this time span.
+SERVER_STARTUP_TIME = 0.1  # Max time, in seconds, we expect the server to need to come up.
```
Kwpolska, line 14: 1. This seems to be unused.
2. The value feels a bit too optimistic.

Kwpolska (changes requested):

tests/integration/test_dev_server.py
aknrdureegaesr: That's a leftover that needs to be deleted.

TMI unless you are interested: I tried to schedule the test proper within the loop, with this much delay. But then, the blocking IO of the HTTP client blocks the thread, so the loop to no longer makes progress, so the server doesn't service the incoming HTTP request, so this whole thing yams. Running that HTTP client via `loop.run_in_executor` fixes that problem.

(The server startup time here is much faster than what we are used to, as there is no build happening here. It is really only HTTP server setup.)

tests/integration/test_dev_server.py
aknrdureegaesr: That's good news! I should have checked whether you use it.

aknrdureegaesr:
I squashed the changes into my previous commit, so there is still only one commit to be merged. I understand that's what the project prefers.

tests/integration/test_dev_server.py
aknrdureegaesr: I'd actually speculate `pytest tests/` is stable and the place of the test source in the directory tree is less stable. But it's not worth investing much thought into. So as you please.

Kwpolska (approved):

tests/integration/test_dev_server.py
Kwpolska: When writing tests that measure time, it’s always good to have some margin to avoid flakiness. The tests may end up running on a slower machine that’s busy with multiple concurrent processes (as CI servers are). For a timeout like this, I’d use 1-2 seconds to be safe.

Kwpolska:
Thanks for implementing this!

PR #3724, opened 2024-01-01
Remove the baseline test new-year race condition. Fix #3723.

### LINKED ISSUE #3723: Remove new-year race condition from "baseline testing".

The baseline tests make innocent pull requests fail early each year. Kindly fix the race condition.

Tests.

- [✔] Yes, I have already written code for it. See #3724 .
- [ ] Yes, I don’t have code ready yet (that’s okay!)
- [ ] No (that’s okay too!)

No, it does not. This concerns only tests.

There is a race condition. The precise circumstances to trigger the race condition are believed to be:

* The pull request is entered early in the year.
* The repo maintainers have not yet adjusted the baseline test to the new year.

You can see that race condition at work in [one early-2024 test run](https://github.com/getnikola/nikola/actions/runs/7377194078/job/20070889821?pr=3722), wrongly blaming pull request #3722.

This is not super-critical, as the baseline test is not required of a pull request.

**Implementation idea**

Integrate a filter that automatically adjusts the year number at that precise place where the diff (wrongly) finds it.

**Prospect: More stabilization. (Not necessarily part of this feature request.)**

Nikola is a Python project. The test could be ported from shell to Python. [subprocess.run](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) is our friend.

It is notoriously difficult to get shell script error handling correct. The baseline test is no exception. E.g., when run from another working directory different from what was intended, the `scripts/getpyver.py` already fails, but the script continues undeterred as if nothing was wrong. The filtering could be done as part of a port of the script to Python.

aknrdureegaesr:
### Pull Request Checklist

- [✔] I’ve read the [guidelines for contributing](https://github.com/getnikola/nikola/blob/master/CONTRIBUTING.rst).
- [✔] I tested my changes.

### Description

Have the baseline test patch the current year to the copyright notes.

Once at it, I also improved error handling and traceability somewhat.

### DIFF
--- a/scripts/baseline.sh
+++ b/scripts/baseline.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e -o pipefail -x
+
 PYVER=$(scripts/getpyver.py short)
 if [[ "$1" == "check" ]]; then
     echo -e "\033[36m>> Downloading baseline for $PYVER...\033[0m"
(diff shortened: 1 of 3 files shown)


scripts/baseline.sh
```diff
@@ -19,8 +22,38 @@ rm "pages/creating-a-theme.rst" "pages/extending.rst" "pages/internals.rst" "pag
 LC_ALL='en_US.UTF-8' PYTHONHASHSEED=0 nikola build --invariant
 if [[ "$1" == "check" ]]; then
     echo -e "\033[36m>> Testing baseline...\033[0m"
-    diff -ubwr ../baseline output
-    if [[ $? == 0 ]]; then
+    python3 -c '
+# In-place edit of copyright notes to adjust the copyright year.
+import time
+YEAR = str(time.gmtime().tm_year)
+for edit_me in [
```
Kwpolska, line 29: We shouldn’t hardcode a list of files, it may change and cause the code to crash.

I would also prefer to use `sed` (with a slightly longer fragment to be sure we’re replacing the right thing) instead of doing it in Python via string manipulation.

aknrdureegaesr: The hard-coded list should or at least could be fixed. I agree to that.

Probably, the whole "patch" approach is not ideal.

What would you think about this idea: When the `--invariant` switch is given, the year should not be pulled from `time.gmtime().tm_year`, but instead some fixed number is used.

But then: Do you think this is worth it in the grand scheme of things?

It is sort of a reflex of mine to try to remove sources of change. My way of doing things: If Nikola version x.y.z once passed all the tests, I tend to fight hard so it will continue to pass all tests forever.

But that is only my preference. It does not seem to be this project's philosophy. For one, you do not pin dependencies, not even to major numbers.

So, what do you think? Is #3723 worth being fixed?

You seem to suggest, it's not.

Kwpolska: I think the solution with `--invariant` would be a good solution to #3723 and one I would accept. This could be done by exposing a function somewhere in Nikola (`utils.py`?) and importing it in the sample `conf.py` file.

> For one, you do not pin dependencies, not even to major numbers.

Pinning dependencies makes things difficult for downstream distributors, and there are a handful. We generally don’t support versions beyond the latest, and we don’t run tests for things other than the `master` branch.

aknrdureegaesr: That makes sense.

TMI, in case you are interested: Now retired, I lived through the age of big enterprise software some 20 years ago, with a new production release every three months only, on a schedule. That official picture was only half-true in reality: We had production bug fixes much more frequently. In that world, it was very important that all automated tests that you ran at official release time would still pass at bug fix time 10 weeks later. The bug-fix build should only change the handful lines of sources we consciously had changed. You bet we fixed dependencies with all the nails and glue we could get our hands on! - 'twas a different world from what Nikola is facing today.

Kwpolska (approved):

PR #3730, opened 2024-01-05
Deprecate yapsy compatibility function names.

aknrdureegaesr:
### Pull Request Checklist

- [X] I tested my changes.

### Description

Output a deprecation warning for people that use the old, Yapsy function names, so those can eventually be removed.

It is ok with me if my work is in the end squashed into one commit, together with the other work on the `new-plugin-mananger` branch.

### DIFF
--- a/nikola/plugin_manager.py
+++ b/nikola/plugin_manager.py
@@ -32,7 +32,8 @@ import importlib.util
 import sys
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, List, Optional, Type, TYPE_CHECKING, Set
+import traceback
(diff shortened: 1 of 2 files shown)


Kwpolska:
Sorry, but the proposed implementation is far too complicated. Only two plugins use this now, I’ll implement something to show a warning unconditionally instead of showing it only once and with `traceback` magic.

aknrdureegaesr:
Fine with me.

> Only two plugins use this now,

Epistemological remark: I don't think you can know all Nikola plugins 🤓.

Kwpolska:
> Epistemological remark: I don't think you can know all Nikola plugins 🤓.

Well, two plugins in the `getnikola/plugins` repository :)

PR #3733, opened 2024-01-10
Obvious (provisorical) fix to lxml dependency incompatibility problem.

aknrdureegaesr:
### Pull Request Checklist

- [X] I tested my changes.

### Description

For now, pin the lxml version to the last 4.x.y to maintain compatibility with existing code.

This is a band-aid. No root cause analysis of the incompatibility has been made yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,7 @@ natsort>=3.5.2
 requests>=2.2.0
 piexif>=1.0.3
 Babel>=2.6.0
+
+# remove when the lxml 5.x.y compatibility has been re-established
+# via https://github.com/getnikola/nikola/issues/3732
+lxml>=4.9.4,<5.0.0


requirements.txt
```diff
@@ -14,3 +14,7 @@ natsort>=3.5.2
 requests>=2.2.0
 piexif>=1.0.3
 Babel>=2.6.0
+
+# remove when the lxml 5.x.y compatibility has been re-established
+# via https://github.com/getnikola/nikola/issues/3732
+lxml>=4.9.4,<5.0.0
```
felixfontein, line 20: Any reason why versions from 3.3.5 to 4.9.3 are not OK?

aknrdureegaesr: I didn't test that, so I cannot answer that question.

Kwpolska: Version 5.0.1 worked fine: https://github.com/getnikola/nikola/actions/runs/7438297625/job/20236910255#step:5:33

Kwpolska:
Superseded by #3735 (which introduced a workaround for the lxml bug).

requirements.txt
felixfontein: Line 9 of the file you edited says `lxml>=3.3.5`.

PR #3735, opened 2024-01-10
Fix #3732 — work around lxml 5.1.0 bug

### LINKED ISSUE #3732: Nikola is not compatible with lxml 5.1.0
### Environment

**Python Version:**
3.9 (not relevant)

**Nikola Version:**
Git commit b96f05a2f8e6f5b60073ca1fb6b24f3e35094db8 (not very relevant)

**Operating System:**
Debian GNU Linux "Bullseye" (probably not relevant)

### Description:

When lxml increased their version from 4.9.4 to 5.1.0, our tests started to fail.

Kwpolska:


tests/integration/test_relative_links.py
```diff
@@ -25,9 +25,11 @@ def test_relative_links(build, output_dir):
     with io.open(test_path, "rb") as inf:
         data = inf.read()
 
+    html = lxml.html.fromstring(data)
+
     assert not any(
         url.startswith("..")
-        for _, _, url, _ in lxml.html.iterlinks(data)
+        for _, _, url, _ in html.iterlinks()
```
aknrdureegaesr, line 32: Personally, I think the following denser version is more readable:

```
assert not any(
    url.startswith("..")
    for _, _, url, _ in lxml.html.fromstring(data).iterlinks()
```

That's somewhat of a matter of taste. If you want to leave that extra variable, fine, but then I suggest to delete the empty line (current 29) as it's only needed in the assert.

Kwpolska: Changed.

PR #3747, opened 2024-01-20
CONTRIBUTING.rst: inform about squash merges

Kwpolska:
Squash merges make history cleaner. I can’t force squash merges, but I can at least disable merge commits, so squash merges are the most likely.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -42,9 +42,13 @@ Here are some guidelines about how you can contribute to Nikola:
   <http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`_.
   It’s not necessary, but if you are doing something big, we recommend
   describing it in the commit message.
+* Your Pull Request should have an informative title and description.
 * While working, rebase instead of merging (if possible). You can use rebase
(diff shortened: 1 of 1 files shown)


felixfontein:
@Kwpolska actually you can force squash merges, assuming you have access to the repo's settings, by disabling both "Allow merge commits" and "Allow rebase merging". Also you can add/modify a branch protection for the `main` branch and enable "Require linear history" there (make sure that "Do not allow bypassing the above settings" is also checked), which will disallow pushing merge commits to that branch.

Kwpolska:
Thanks, I only looked in the branch protection options, I didn’t know about the options on the main settings page. I disabled the other merge options and removed *probably* from the added line.

felixfontein:
Yeah, the settings are quite spread out, it always takes me some time to find them all, even knowing that they exist :) (Especially when changing between GitHub and GitLab; both have their own settings maze...)

Kwpolska:
(Or Azure DevOps, the original Microsoft GitHub I use at work, which considers merge types part of [branch policy](https://learn.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops&tabs=browser).)

PR #3760, opened 2024-02-20
Template improvements

aknrdureegaesr:
### Pull Request Checklist

- [x] I tested my changes.

### Description

New features:

- Trace template usage if env variable NIKOLA_TEMPLATES_TRACE is set.
- Enable raw template engine control via opitonal `conf.py` factory method.
- Clarify template documentation: A parent template is not strictly needed.

Additional code improvements
helpful for those developing templates who actually read the code:

- Added several typing hints and improved others.
- A few code comments were added or improved.

### Details

The commit message of the first commit can be used as the commit message of this entire PR.

### DIFF
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -200,9 +200,8 @@ the rest from a suitable parent theme.
 
 .. Tip::
 
-   If you set the environment variable ``NIKOLA_TEMPLATES_TRACE`` to any non-empty value
-   (``true`` is recommended), Nikola will log template usage, both on output and also
-   into a file ``templates_log.txt``.
+   If you set the environment variable ``NIKOLA_TEMPLATES_TRACE`` to ``true``,
+   Nikola will log template usage, both to output and also into a file ``templates_log.txt``.
 
 Apart from the `built-in templates`_ listed below, you can add other templates for specific
 pages, which the user can then use in his ``POSTS`` or ``PAGES`` option in
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -130,9 +130,17 @@ The following keys are currently supported:
 
     The parent is so you don’t have to create a full theme each time: just
     create an empty theme, set the parent, and add the bits you want modified.
-    It is strongly recommended you define a parent.  If you don't, many features
-    won’t work due to missing templates, messages, and assets until your home-grown
-    template is complete.
+
+    While it is possible to create a theme without a parent, it is
+    **strongly discouraged** and not officially supported, in the sense:
+    We won't help with issues that are caused by a theme being parentless,
+    and we won't guarantee that it will always work with new Nikola versions.
+    The `base` and `base-jinja` themes provide assets, messages, and generic templates
+    that Nikola expects to be able to use in all sites. That said, if you are making
+    something very custom, Nikola will not prevent the creation of a theme
+    without `base`, but you will need to manually determine which templates and
+    messages are required in your theme. (Initially setting the ``NIKOLA_TEMPLATES_TRACE``
+    environment variable might be of some help, see below.)
 
     The following settings are recommended:
 
@@ -483,6 +491,28 @@ at https://www.transifex.com/projects/p/nikola/
 
 If you want to create a theme that has new strings, and you want those strings to be translatable,
 then your theme will need a custom ``messages`` folder.
+
+Configuration of the raw template engine
+----------------------------------------
+
+For usage not covered by the above, you can define a method
+`TEMPLATE_ENGINE_FACTORY` in `conf.py` that constructs the raw
+underlying templating engine. That `raw_engine` that your method
+needs to return is either a `jinja2.Environment` or a
+`mako.loopkup.TemplateLookup` object. Your factory method is
+called with the same arguments as is the pertinent `__init__`.
+
+E.g., to configure `jinja2` to bark and error out on missing values,
+instead of silently continuing with empty content, you might do this:
+
+.. code:: python
+
+    # Somewhere in conf.py:
+    def TEMPLATE_ENGINE_FACTORY(**args) -> jinja2.Environment:
+        augmented_args = dict(args)
+        augmented_args['undefined'] = jinja2.StrictUndefined
+        return jinja2.Environment(**augmented_args)
+
 
 `LESS <http://lesscss.org/>`__ and `Sass <https://sass-lang.com/>`__
 --------------------------------------------------------------------
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -510,7 +510,7 @@ instead of silently continuing with empty content, you might do this:
     # Somewhere in conf.py:
     def TEMPLATE_ENGINE_FACTORY(**args) -> jinja2.Environment:
         augmented_args = dict(args)
-        augmented_args['undefined'] = jinja2.StrictUndefined
+        augmented_args['undefined'] = jinja2.DebugUndefined
         return jinja2.Environment(**augmented_args)
 
 
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -7,7 +7,7 @@
 .. author: The Nikola Team
 
 :Version: 8.3.1
-:Author: Roberto Alsina <ralsina@netmanagers.com.ar>
+:Author: Roberto Alsina <ralsina@netmanagers.com.ar> and others
 
 .. class:: alert alert-primary float-md-right
 
@@ -130,8 +130,9 @@ The following keys are currently supported:
 
     The parent is so you don’t have to create a full theme each time: just
     create an empty theme, set the parent, and add the bits you want modified.
-    You **must** define a parent, otherwise many features won’t work due to
-    missing templates, messages, and assets.
+    It is strongly recommended you define a parent.  If you don't, many features
+    won’t work due to missing templates, messages, and assets until your home-grown
+    template is complete.
 
     The following settings are recommended:
 
@@ -184,8 +185,16 @@ so ``post.tmpl`` only define the content, and the layout is inherited from ``bas
 
 Another concept is theme inheritance. You do not need to duplicate all the
 default templates in your theme — you can just override the ones you want
-changed, and the rest will come from the parent theme. (Every theme needs a
-parent.)
+changed, and the rest will come from the parent theme.  If your theme does not
+define a parent, it needs to be complete.  It is generally a lot harder to
+come up with a complete theme, compared to only changing a few files and using
+the rest from a suitable parent theme.
+
+.. Tip::
+
+   If you set the environment variable ``NIKOLA_TEMPLATES_TRACE`` to any non-empty value
+   (``true`` is recommended), Nikola will log template usage, both on output and also
+   into a file ``templates_log.txt``.
 
 Apart from the `built-in templates`_ listed below, you can add other templates for specific
 pages, which the user can then use in his ``POSTS`` or ``PAGES`` option in
@@ -194,11 +203,11 @@ page via the ``template`` metadata, and custom templates can be added in the
 ``templates/`` folder of your site.
 
 If you want to modify (override) a built-in template, use ``nikola theme -c
-<name>.tmpl``.  This command will copy the specified template file to the
-``templates/`` directory of your currently used theme.
+<name>.tmpl``.  This command will copy the specified template file from the
+parent theme to the ``templates/`` directory of your currently used theme.
 
 Keep in mind that your theme is *yours*, so you can require whatever data you
-want (eg. you may depend on specific custom ``GLOBAL_CONTEXT`` variables, or
+want (e.g., you may depend on specific custom ``GLOBAL_CONTEXT`` variables, or
 post meta attributes). You don’t need to keep the same theme structure as the
 default themes do (although many of those names are hardcoded). Inheriting from
 at least ``base`` (or ``base-jinja``) is heavily recommended, but not strictly
--- a/nikola/__init__.py
+++ b/nikola/__init__.py
@@ -31,10 +31,13 @@ import sys
 
 # The current Nikola version:
 __version__ = '8.3.1'
+
 # A flag whether logging should emit debug information:
 DEBUG = bool(os.getenv('NIKOLA_DEBUG'))
+
 # A flag whether special templates trace logging should be generated:
 TEMPLATES_TRACE = bool(os.getenv('NIKOLA_TEMPLATES_TRACE'))
+
 # When this flag is set, fewer exceptions are handled internally;
 # instead they are left unhandled for the run time system to deal with them,
 # which typically leads to the stack traces being exposed.
--- a/nikola/__init__.py
+++ b/nikola/__init__.py
@@ -38,9 +38,7 @@ DEBUG = bool(os.getenv('NIKOLA_DEBUG'))
 # A flag whether special templates trace logging should be generated:
 TEMPLATES_TRACE = bool(os.getenv('NIKOLA_TEMPLATES_TRACE'))
 
-# When this flag is set, fewer exceptions are handled internally;
-# instead they are left unhandled for the run time system to deal with them,
-# which typically leads to the stack traces being exposed.
+# A flag to show tracebacks of unhandled exceptions.
 # This is a less noisy alternative to the NIKOLA_DEBUG flag.
 SHOW_TRACEBACKS = bool(os.getenv('NIKOLA_SHOW_TRACEBACKS'))
 
--- a/nikola/__init__.py
+++ b/nikola/__init__.py
@@ -31,13 +31,14 @@ import sys
 
 # The current Nikola version:
 __version__ = '8.3.1'
-# A flag whether logging should emmit debug information:
+# A flag whether logging should emit debug information:
 DEBUG = bool(os.getenv('NIKOLA_DEBUG'))
 # A flag whether special templates trace logging should be generated:
 TEMPLATES_TRACE = bool(os.getenv('NIKOLA_TEMPLATES_TRACE'))
 # When this flag is set, fewer exceptions are handled internally;
 # instead they are left unhandled for the run time system to deal with them,
 # which typically leads to the stack traces being exposed.
+# This is a less noisy alternative to the NIKOLA_DEBUG flag.
 SHOW_TRACEBACKS = bool(os.getenv('NIKOLA_SHOW_TRACEBACKS'))
 
 if sys.version_info[0] == 2:
--- a/nikola/__init__.py
+++ b/nikola/__init__.py
@@ -29,8 +29,15 @@
 import os
 import sys
 
+# The current Nikola version:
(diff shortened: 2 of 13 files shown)


docs/theming.rst
```diff
@@ -130,8 +134,9 @@ The following keys are currently supported:
 
     The parent is so you don’t have to create a full theme each time: just
     create an empty theme, set the parent, and add the bits you want modified.
-    You **must** define a parent, otherwise many features won’t work due to
-    missing templates, messages, and assets.
+    It is strongly recommended you define a parent.  If you don't, many features
```
Kwpolska, line 137: Nope. Every theme must inherit from at least `base`/`base-jinja`, we explicitly refuse to support any other scenario (even if the code does not prevent this, we would prefer not to admit that it is a possibility, and we reserve the right to break it at any time.)

docs/theming.rst
```diff
@@ -163,6 +168,13 @@ The following keys are currently supported:
   * ``ignored_assets`` — comma-separated list of assets to ignore (relative to
     the ``assets/`` directory, eg. ``css/theme.css``)
 
+* ``jinja`` - This section is ignored unless your theme's engine is ``jinja``.
+
+  * ``extensions`` - comma-separated list of
```
Kwpolska, line 173: I’m OK with this, but perhaps it would be more useful to do this in `conf.py` instead?

docs/theming.rst
```diff
@@ -184,8 +196,16 @@ so ``post.tmpl`` only define the content, and the layout is inherited from ``bas
 
 Another concept is theme inheritance. You do not need to duplicate all the
 default templates in your theme — you can just override the ones you want
-changed, and the rest will come from the parent theme. (Every theme needs a
-parent.)
+changed, and the rest will come from the parent theme.  If your theme does not
```
Kwpolska, line 199: Again, unsupported and should not be documented.

nikola/__init__.py
```diff
@@ -29,8 +29,15 @@
 import os
 import sys
 
+# The current Nikola version:
 __version__ = '8.3.0'
+# A flag whether logging should emmit debug information:
```
Kwpolska, line 34:
```suggestion
# A flag whether logging should emit debug information:
```

nikola/__init__.py
```diff
@@ -29,8 +29,15 @@
...
 DEBUG = bool(os.getenv('NIKOLA_DEBUG'))
+# A flag whether special templates trace logging should be generated:
+TEMPLATES_TRACE = bool(os.getenv('NIKOLA_TEMPLATES_TRACE'))
+# When this flag is set, fewer exceptions are handled internally;
+# instead they are left unhandled for the run time system to deal with them,
+# which typically leads to the stack traces being exposed.
```
Kwpolska, line 40:
```suggestion

# A flag to show tracebacks of unhandled exceptions.
# An alternative to the DEBUG flag with less noise.
```

(`NIKOLA_DEBUG` used to be very spammy with yapsy, it’s still not great to have it always on; `export NIKOLA_SHOW_TRACEBACKS=1` is in my `.zshrc` though.)

nikola/log.py
```diff
@@ -137,6 +136,39 @@ def get_logger(name: str, handlers=None) -> logging.Logger:
...
+    fhandler.setFormatter(formatter)
+    fhandler.setLevel(logging.DEBUG)
+
+    TEMPLATES_LOGGER.handlers = [shandler, fhandler]
+    TEMPLATES_LOGGER.propagate = False
+
+    TEMPLATES_LOGGER.info("Template usage being traced to file %s", filename)
+
+
+if TEMPLATES_TRACE:
```
Kwpolska, line 170: Setting NIKOLA_DEBUG should give you *all* the debug info possible.

```suggestion
if DEBUG or TEMPLATES_TRACE:
```

nikola/plugin_categories.py
```diff
@@ -201,10 +200,14 @@ class TemplateSystem(BasePlugin):
 
     name = "dummy_templates"
 
-    def set_directories(self, directories: 'typing.List[str]', cache_folder: str):
+    def set_directories(self, directories: List[str], cache_folder: str) -> None:
         """Set the list of folders where templates are located and cache."""
         raise NotImplementedError()
 
+    def user_configuration(self, user_config: Mapping[str, str]) -> None:
```
Kwpolska, line 207: Consider adding a verb to the method name, perhaps `apply_user_configuration`? (`Apply` would also work nicely in the docstring.)

nikola/plugin_categories.py
```diff
@@ -412,19 +415,19 @@ def split_metadata_from_text(self, source_text: str) -> (str, str):
                 # Necessary?
                 return split_result[0], split_result[-1]
 
-    def extract_text(self, source_text: str) -> 'typing.Dict[str, str]':
+    def extract_text(self, source_text: str) -> Dict[str, str]:
         """Split file, return metadata and the content."""
+        # TODO: The name and interface of this method is a mess and needs to be cleaned up.
         split = self.split_metadata_from_text(source_text)
-        if not split:
+        if len(split[0]) == 0:
```
Kwpolska, line 422: `not` is more Pythonic and can catch things like `None` if it somehow ended up there.

```suggestion
        if not split[0]:
```

nikola/plugin_manager.py
```diff
@@ -35,13 +35,12 @@
...
 from pathlib import Path
-from typing import Dict, List, Optional, Type, TYPE_CHECKING, Set
+from typing import Dict, Iterable, List, Optional, Type, Set
 
 from .plugin_categories import BasePlugin, CATEGORIES
 from .utils import get_logger
 
-if TYPE_CHECKING:
-    import logging
+import logging
```
Kwpolska, line 43: Consider moving this to the top (line 32).

nikola/plugins/template/jinja.py
```diff
@@ -45,15 +46,35 @@ class JinjaTemplates(TemplateSystem):
...
 
     name = "jinja"
-    lookup = None
+    if jinja2 is None:
+        lookup = None
+    else:
+        lookup: Optional[jinja2.Environment] = None
     dependency_cache = {}
     per_file_cache = {}
+    _user_configured_jina_extensions: List[str] = []
```
Kwpolska, line 55: Typo: `jina` → `jinja`

nikola/plugins/template/jinja.py
```diff
@@ -45,15 +46,35 @@ class JinjaTemplates(TemplateSystem):
     """Support for Jinja2 templates."""
 
     name = "jinja"
-    lookup = None
+    if jinja2 is None:
+        lookup = None
+    else:
+        lookup: Optional[jinja2.Environment] = None
```
Kwpolska, line 52: Consider using a string annotation:

```suggestion
    lookup: 'Optional[jinja2.Environment]' = None
```

nikola/utils.py
```diff
@@ -1938,19 +1943,19 @@ def sort_posts(posts, *keys):
...
     'foo; bar'
     >>> smartjoin(' to ', ['count', 42])
     'count to 42'
+
+    The treatment of bytes (calling str(string_or_iterable)) is somewhat dubious. Is this needed?
     """
-    if isinstance(string_or_iterable, (str, bytes)):
-        return string_or_iterable
-    elif isinstance(string_or_iterable, Iterable):
+    if isinstance(string_or_iterable, Iterable):
```
Kwpolska, line 1958: This won’t work, please undo:

```pycon
>>> isinstance("a", typing.Iterable)
True
```

nikola/utils.py
```diff
@@ -1938,19 +1943,19 @@ def sort_posts(posts, *keys):
...
+    """Join string_or_iterable with join_char if it is iterable; otherwise convert it to string.
 
     >>> smartjoin('; ', 'foo, bar')
     'foo, bar'
     >>> smartjoin('; ', ['foo', 'bar'])
     'foo; bar'
     >>> smartjoin(' to ', ['count', 42])
     'count to 42'
+
+    The treatment of bytes (calling str(string_or_iterable)) is somewhat dubious. Is this needed?
```
Kwpolska, line 1956: Might be worth checking if `bytes` is ever passed here.

nikola/__init__.py
```diff
@@ -29,8 +29,15 @@
 import os
 import sys
 
+# The current Nikola version:
 __version__ = '8.3.0'
+# A flag whether logging should emmit debug information:
 DEBUG = bool(os.getenv('NIKOLA_DEBUG'))
+# A flag whether special templates trace logging should be generated:
```
Kwpolska, line 36:
```suggestion

# A flag whether special templates trace logging should be generated:
```

Kwpolska (changes requested):

nikola/log.py
```diff
@@ -137,6 +136,39 @@ def get_logger(name: str, handlers=None) -> logging.Logger:
...
+    init_template_trace_logging("templates_log.txt")
```
Kwpolska, line 171: Consider using the `.log` extension, to avoid confusion with things like reST posts (which can also end with `.txt`).

```suggestion
    init_template_trace_logging("templates_trace.log")
```

docs/theming.rst
aknrdureegaesr: Thinking about it a bit further: From my point of view, it would be useful if Nokia would simply officially state that

* it is possible to go without a parent template
* things might break on any Nikola version update, as new template files become required from the code.

As for support, that would mean for Nokia a new template requirement needs to be announced in the change file.

Would that be something you'd be willing to do if asked nicely?

Kwpolska: Here’s some example wording that might work:

> While it is possible to create a theme without a parent, it is **strongly discouraged** and not officially supported. The `base` and `base-jinja` themes provide assets, messages, and generic templates that Nikola expects to be able to use in all sites. That said, if you are making something very custom, Nikola will not prevent the creation of a theme without `base`, but you will need to manually determine which templates and messages are required in your theme.

It’s unlikely we’ll ever introduce a new template, but I guess it might be mentioned in the changelog if that happens.

aknrdureegaesr: As far as the changelog is concerned: Anything in the templates can be user-changed. So I think Nikola has to mention any change in the template arena in the changelog anyway. I see that as independent of the question "with base or not".

aknrdureegaesr: Some stance as the one you suggest would answer my concerns.

For some details: I would prefer to warn people of consequences, yes, but otherwise treat them as adults who are able to make grown-up decisions. So I would think it nicer if we left out the "strongly discouraged" part. Maybe warn people that growing their own theme is harder than they might think at first sight.

As for the "not officially supported": Before people file a bug, we should require them to add the pertinent parent. If that makes the bug go away, it must not be filed.

Anything else the "not officially supported" entails?

Kwpolska: “Strongly discouraged” and “not officially supported” are meant to suggest “don’t do this unless you really know what you’re doing”. For user-facing software, phrases like this are a good idea IMO to clearly show the limitations.

“Not officially supported” means basically what you said, i.e. we won’t help with issues caused by a parentless theme, and we don’t guarantee it will always work.

docs/theming.rst
aknrdureegaesr: I first thought the opportune place for template configuration is the theme configuration file. It is there I tell Nikola which engine to use in the first place.

On the other hand, `conf.py` would allow real python code, as opposed to mere strings. It also has the global context and other such goodies, they may interact with the template. Since presenting this PR, I've also encountered the desire to have my templates [bark at undefined varialbes](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.StrictUndefined) as opposed to just printing them as an empty string.

So I'll agree doing something like this in `conf.py` is the better idea, and doing it in a way that allows any configuration change (via Python), as opposed to just allowing extension activation and nothing else.

docs/theming.rst
aknrdureegaesr: Via the polyfill discussion, a solid reason surfaced why Nikola might rethink its policy of not officially supporting templates without parent.

Where GDPR rules, one needs a certain written agreement with any company before sending web traffic there way. This is a consequence of [GDPR art 28](https://gdpr-info.eu/art-28-gdpr/), in particular section 2.. (I've personally been one of many developers who ported all references to Google fonts to locally hosted, in the web presence of a big Germany company (some 300.000+ employed).)

I believe small, private web sites are inside the target group of Nikola. People responsible for such web sites might not have the time and legal power to establish all agreements needed. So they may choose to use no external resources, but self-host everything their web site needs.

Nikola makes not promises to that effect presently. Indeed, build-in Nikola templates freely use external resources. Any Nikola version update might conceivably introduce or change external references via the HTML produced. Wanting to play it safe is a very solid reason to use a template without parent.

felixfontein: Generally there's the option `use_cdn`, but as #3763 shows this isn't always respected, like for `polyfill.io`.

I'm personally also using a template without a parent for my sites. (More precisely I have my own base template, and some templates deriving from my base.)

nikola/__init__.py
aknrdureegaesr: Done. Thanks.

nikola/__init__.py
aknrdureegaesr: The wording "of unhandled exceptions" suggests whether the exceptions is handled or not is independent of this setting. As far as I remember, this is not the case.

nikola/plugins/template/jinja.py
aknrdureegaesr: Could be done and would technically work. But hides the concern and leaves the reader puzzling why the string has been used, as opposed to the plain `jinja2.Environment`. By putting it into the `else` part, the concern is clear and explicit.

nikola/utils.py
aknrdureegaesr: Good point, thanks!

But what we had previously also doesn't work, as `bytes` isn't a `str`.

How about calling `str( )` for both `str` (where it is a no-op) and `bytes` (where you can be lucky, so it might work)?

nikola/utils.py
aknrdureegaesr: Yes!

nikola/__init__.py
aknrdureegaesr: Done.

docs/theming.rst
aknrdureegaesr: What do you think about the compromise "we document that this is unsupported", which is what I'm trying to do here?

nikola/log.py
aknrdureegaesr: Done

nikola/plugin_categories.py
aknrdureegaesr: Done.

nikola/plugin_manager.py
aknrdureegaesr: Done.

(I'm used to `isort` doing this kind of thing for me.)

docs/theming.rst
Kwpolska: I’m still not a fan of documenting it at all, but I guess it can stay as-is.

docs/theming.rst
```diff
@@ -184,8 +193,16 @@ so ``post.tmpl`` only define the content, and the layout is inherited from ``bas
...
+define a parent, it needs to be complete.  It is generally a lot harder to
+come up with a complete theme, compared to only changing a few files and using
+the rest from a suitable parent theme.
+
+.. Tip::
+
+   If you set the environment variable ``NIKOLA_TEMPLATES_TRACE`` to any non-empty value
+   (``true`` is recommended), Nikola will log template usage, both on output and also
+   into a file ``templates_log.txt``.
```
Kwpolska, line 205: Unnecessary detail.

```suggestion
   If you set the environment variable ``NIKOLA_TEMPLATES_TRACE`` to ``true``,
   Nikola will log template usage, both on output and also into a file ``templates_log.txt``.
```

nikola/__init__.py
Kwpolska: Technically, all exceptions are handled. This setting affects exceptions that bubble up to the global handler and therefore crash the process.

nikola/plugin_categories.py
```diff
@@ -61,7 +60,7 @@ class BasePlugin:
     """Base plugin class."""
 
     logger = None
-    site: 'nikola.nikola.Nikola'
+    site: Optional['nikola.nikola.Nikola'] = None # NOQA  # Fixme: Circular import
```
Kwpolska, line 63: Not really something we can ever fix.

nikola/plugin_categories.py
```diff
@@ -202,10 +201,14 @@ class TemplateSystem(BasePlugin):
 
     name = "dummy_templates"
 
-    def set_directories(self, directories: 'typing.List[str]', cache_folder: str):
+    def set_directories(self, directories: List[str], cache_folder: str) -> None:
         """Set the list of folders where templates are located and cache."""
         raise NotImplementedError()
 
+    def user_engine_factory(self, factory: Callable[..., Any]) -> None:
```
Kwpolska, line 208: Consider a different name that includes a verb. Perhaps `set_user_engine_factory`?

nikola/plugin_categories.py
aknrdureegaesr: Actually, yes we can. And I have a feeling we should.

We could have an abstract base clase for `site` that has all the members that are actually used by whoever uses `site`, and `Nikola` could implement `site`. Then this member here could be of type `Optional[site]` and - puff - the need to import `Nikola` has vanished.

I have a feeling this would actually be an improvement, not only because of this import problem. The `site` is something we use in a lot of places, and separating contract (new site type) and implementation (Nikola type) might be a good idea.

Would you like to have a new issue towards that end? I'm volunteering to write it.

Kwpolska: I am against introducing unnecessary abstractions.

Kwpolska: I am against introducing unnecessary abstractions, and I consider defining a special `Site` type unnecessary. Nikola is a project where users can, and have, used internals extensively. An abstraction would add more maintenance cost and zero benefit.

docs/theming.rst
aknrdureegaesr: I take it that means we can resolve this conversation. Feel free to re-open if I got that wrong.

nikola/__init__.py
aknrdureegaesr: Technically, all exceptions are handled. I tend to use a somewhat lax terminology: If an exception bubbles up to the global handler (that causes it to crash the process), it is essentially *not* handled.

docs/theming.rst
aknrdureegaesr: I wanted to avoid tempting people to set it to ``false``, which will lead to unexpected results.

aknrdureegaesr: But as the result will not be catastrophic, we can use the shorter version alright. Or else change the code to specifically check for `"true"`.

Kwpolska (approved):

PR #3763, opened 2024-03-01
Remove unnecessary `polyfill.io`

SukkaW:
### Pull Request Checklist

- [x] I tested my changes.

### Description

I haven't updated the `AUTHORS.txt` as I though small changes like this should not qualify.

The PR replaces all `polyfill.io` with `cdnjs.cloudflare.com/polyfill` in the docs and the template.

`polyfill.io` was acquired by **a China-based CDN company** "Funnull", see [the announcement from the `polyfill.io` domain owner's Twitter](https://x.com/JakeDChampion/status/1761315227008643367) and https://github.com/polyfillpolyfill/polyfill-service/issues/2834. Despite Funnull's claims of operating in the United States, the predominance of Simplified Chinese on its website suggests otherwise, and it turns out that **"Funnull" is notorious for providing service for the betting and pornography industries**.

[The original creator of the `polyfill.io` has voiced his concern on Twitter](https://twitter.com/triblondon/status/1761852117579427975). And since the acquisition, numerous issues have emerged (https://github.com/polyfillpolyfill/polyfill-service/issues/2835, https://github.com/polyfillpolyfill/polyfill-service/issues/2838, https://github.com/alist-org/alist/issues/6100), rendering the `polyfill.io` service **extremely unstable**. Since then, Fastly ([Announcement](https://community.fastly.com/t/new-options-for-polyfill-io-users/2540)) and Cloudflare ([Announcement](https://blog.cloudflare.com/polyfill-io-now-available-on-cdnjs-reduce-your-supply-chain-risk)) has hosted their own instances of `polyfill.io` service.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -22,6 +22,7 @@ Other
   promising support for versions supported by the Python core team,
   Ubuntu LTS, or Debian stable, and taking into consideration
   Debian oldstable and PyPy.
+* Remove polyfill from `polyfill.io`.
 
 New in v8.3.0
 =============
--- a/docs/manual.rst
+++ b/docs/manual.rst
@@ -1690,10 +1690,6 @@ For Mako:
 .. code:: html
 
     % if date_fanciness != 0:
-    %if date_fanciness == 2:
-        <!-- Polyfill for relative dates in Safari -- best handled with a CDN -->
-        <script src="https://polyfill.io/v3/polyfill.js?features=Intl.RelativeTimeFormat.%7Elocale.${luxon_locales[lang]}"></script>
-    %endif
     <!-- required scripts -- best handled with bundles -->
     <script src="/assets/js/luxon.min.js"></script>
     <script src="/assets/js/fancydates.js"></script>
@@ -1712,10 +1708,6 @@ For Jinja2:
 .. code:: html
 
     {% if date_fanciness != 0 %}
-    {% if date_fanciness == 2 %}
-        <!-- Polyfill for relative dates in Safari -- best handled with a CDN -->
-        <script src="https://polyfill.io/v3/polyfill.js?features=Intl.RelativeTimeFormat.%7Elocale.{{ luxon_locales[lang] }}"></script>
-    {% endif %}
     <!-- required scripts -- best handled with bundles -->
     <script src="/assets/js/luxon.min.js"></script>
     <script src="/assets/js/fancydates.js"></script>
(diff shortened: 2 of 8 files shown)


CHANGES.txt
```diff
@@ -22,6 +22,7 @@ Other
   promising support for versions supported by the Python core team,
   Ubuntu LTS, or Debian stable, and taking into consideration
   Debian oldstable and PyPy.
+* Replace `polyfill.io` with `cdnjs.cloudflare.com/polyfill`.
```
felixfontein, line 25: Please note that this file is using reStructuredText, so you need to use double backticks:
```suggestion
* Replace ``polyfill.io`` with ``cdnjs.cloudflare.com/polyfill``.
```

felixfontein (commented):
Thanks for your contribution!

SukkaW:
> I’m wondering if we actually need the polyfills. They were required for Safari, but Safari seems to have added this feature in 2020. Do you have access to a Mac to verify?

I only have a Apple-Silicon powered Mac thus can't install the Safari 14. But many repositories I've opened a PR to are suggesting removing the polyfill entirely, and we could do the same.

On the other hand, removing the polyfill could be a breaking change though.

Kwpolska:
Could you still test in whichever newer Safari version you have? If [the feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) actually works there in different languages, we can remove the polyfill altogether (we’ll just assume the information about 2020/Safari 14 is correct), Apple users will probably be on a recent version anyway.

SukkaW:
> Could you still test in whichever newer Safari version you have? If [the feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) actually works there in different languages, we can remove the polyfill altogether (we’ll just assume the information about 2020/Safari 14 is correct), Apple users will probably be on a recent version anyway.

So I opened the MDN page on my Safari (Safari 17.4) and ran the example, it does work though.

image

Kwpolska:
Great, in that case, please remove the polyfill altogether :)

aknrdureegaesr:
Side remark: For web sites that try to comply with EU GDPR, one needs a signed written agreement of a certain kind with any CDN or similar entity, before sending traffic their way. I vividly remember a concerted effort across many dev teams to remove Google fonts from all web presences of a big German business that was my employer at the time. From a legal point of view, it is much safer to self-hosting all resources needed.

SukkaW:
@felixfontein @Kwpolska I've rebased the PR and all `polyfill.io` usage are removed.

Kwpolska (approved):

Kwpolska:
Thanks for bringing this to our attention and fixing it!

PR #3765, opened 2024-03-02
Remove more leftovers from annotateit.org

### LINKED ISSUE #3764: `annotation_helper.tmpl` includes jquery via HTTP, includes no longer existing assets.annotateit.org
`annotation_helper.tmpl` references two resources via HTTP on `assets.annotateit.org`, which no longer exists, and includes jquery via HTTP as well (which should be changed to HTTPS).

On the other hand, it seems that `annotation_helper.tmpl` is no longer really used; it's still included in some theme, but not used at all. 8dc03e129b171e3fee78e379763e4557339280bf removed `notes.css()`, and before that, 4dff6741c8d4663f2c626623a9a7a0ea2761ef17  removed most of annotation support.

So I guess this is just a leftover.

felixfontein:
Fixes #3764.


Kwpolska (approved):

felixfontein:
Thanks for reviewing and merging!

Issue #3766, opened 2024-03-14
nikola auto throws ValueError exception on rebuild


wohali:
### Environment

**Python Version:** 3.11.2

**Nikola Version:** `master` [0a533d6](https://github.com/getnikola/nikola/commit/0a533d682565f80b4e9190c20554e584ded636d1)

**Operating System:** WSL 2.0

### Description:

nikola auto throws an exception when rebuilding, though the rebuild seems to be successful:

```
$ nikola auto
[2024-03-14 17:54:06] INFO: auto: REBUILDING SITE
[2024-03-14 17:54:07] INFO: auto: Rebuild successful
Scanning posts........done!

[2024-03-14 17:54:07] INFO: auto: Serving on http://127.0.0.1:8000/ ...
[2024-03-14 17:54:15] INFO: auto: REBUILDING SITE
[2024-03-14 17:54:16] ERROR: asyncio: Task exception was never retrieved
future: <Task finished name='Task-25' coro=<CommandAuto.reload_page() done, defined at /usr/local/lib/python3.11/dist-packages/nikola/plugins/command/auto/__init__.py:426> exception=ValueError('no path specified')>
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/nikola/plugins/command/auto/__init__.py", line 434, in reload_page
    p = os.path.relpath(event_path, os.path.abspath(self.site.config['OUTPUT_FOLDER'])).replace(os.sep, '/')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 490, in relpath
ValueError: no path specified
[2024-03-14 17:54:16] ERROR: asyncio: Task exception was never retrieved
future: <Task finished name='Task-26' coro=<CommandAuto.reload_page() done, defined at /usr/local/lib/python3.11/dist-packages/nikola/plugins/command/auto/__init__.py:426> exception=ValueError('no path specified')>
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/nikola/plugins/command/auto/__init__.py", line 434, in reload_page
    p = os.path.relpath(event_path, os.path.abspath(self.site.config['OUTPUT_FOLDER'])).replace(os.sep, '/')
[...]

PR #3775, opened 2024-05-22
properly support watchdog 4+ where events always have dest_path

evgeni:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Since watchdog 4 all events have the `dest_path` property,
but that property can be *empty*, thus breaking `nikola auto`:

    ERROR: asyncio: Task exception was never retrieved
    future: <Task finished name='Task-57' coro=<CommandAuto.reload_page() done, defined at nikola/plugins/command/auto/__init__.py:427> exception=ValueError('no path specified')>
      File "nikola/plugins/command/auto/__init__.py", line 440, in reload_page
        p = os.path.relpath(event_path, os.path.abspath(self.site.config['OUTPUT_FOLDER'])).replace(os.sep, '/')
      File "/usr/lib64/python3.8/posixpath.py", line 453, in relpath
        raise ValueError("no path specified")
    ValueError: no path specified

### DIFF
--- a/nikola/plugins/command/auto/__init__.py
+++ b/nikola/plugins/command/auto/__init__.py
@@ -429,7 +429,7 @@ class CommandAuto(Command):
         # Move events have a dest_path, some editors like gedit use a
         # move on larger save operations for write protection
         if event:
-            event_path = event.dest_path if hasattr(event, 'dest_path') else event.src_path
+            event_path = event.dest_path if (hasattr(event, 'dest_path') and event.dest_path) else event.src_path
         else:
             event_path = self.site.config['OUTPUT_FOLDER']
         p = os.path.relpath(event_path, os.path.abspath(self.site.config['OUTPUT_FOLDER'])).replace(os.sep, '/')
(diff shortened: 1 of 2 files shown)


evgeni:
See https://github.com/gorakhargosh/watchdog/commit/41fca1eb601139922817a17b6335a8c2d2c8db51 for the change in watchdog

evgeni:
Fixes https://github.com/getnikola/nikola/issues/3766, I guess

nikola/plugins/command/auto/__init__.py
```diff
@@ -429,7 +429,7 @@ async def reload_page(self, event) -> None:
         # Move events have a dest_path, some editors like gedit use a
         # move on larger save operations for write protection
         if event:
-            event_path = event.dest_path if hasattr(event, 'dest_path') else event.src_path
+            event_path = event.dest_path if (hasattr(event, 'dest_path') and event.dest_path) else event.src_path
```
evgeni, line 432:
```suggestion
            event_path = getattr(event, 'dest_path', None) or event.src_path
```
This was suggested in #3766 and also works, not sure which is more Pythonic tho.

Kwpolska: I think the suggested change is less Pythonic.

Kwpolska (approved):

Kwpolska:
Thanks for contributing!

PR #3780, opened 2024-06-19
Derra branch Mine functions I improved the coverage for

DarianTheGoat1:
### Pull Request Checklist

- [ ] I tested my changes.


Kwpolska (changes requested):
This is not how code coverage works. To increase the code coverage, you should write tests that exercise the edge cases, but also **verify the output makes sense**.

You should not add things like `branch_coverage`, `report_coverage`, `reset_coverage` to the code. You should change as little as possible in the production code (only change the tests, maybe adjust the production code to make it more testable).

Issue #3783, opened 2024-07-21
[possible regression] inline code breaks build


Sponge5:
### Environment

**Python Version:**
Python 3.12.4
**Nikola Version:**
Nikola v8.3.1
**Operating System:**
Archlinux
**Docutils:**
Docutils 0.21.2

### Description:

Hi, this is possibly a docutils issue but I'm first opening it here. After updating my environment and updating (reinstalling) all packages in it, `nikola build` is failing on `.rst` files with inline code. Running `find posts/ -type f | xargs sed -i 's/:code:/:subscript:/g'` fixes the problem. Maybe I'm missing a package or something, but if not, here is the stacktrace:

```
$ nikola build
...
PythonAction Error
Traceback (most recent call last):
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/doit/action.py", line 461, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/nikola/post.py", line 729, in compile
    self.compile_html(
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/nikola/plugins/compile/rest/__init__.py", line 151, in compile
    output, error_level, deps, shortcode_deps = self.compile_string(data, source, is_two_file, post, lang)
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/nikola/plugins/compile/rest/__init__.py", line 134, in compile_string
    output, error_level, deps, _ = rst2html(
                                   ^^^^^^^^^
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/nikola/plugins/compile/rest/__init__.py", line 382, in rst2html
    pub.publish(enable_exit_status=enable_exit_status)
  File "/home/adam/Other/PythonEnv/adamprocio.cz/lib/python3.12/site-packages/docutils/core.py", line 237, in publish
    output = self.writer.write(self.document, self.destination)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3784, opened 2024-08-11
fix: [rest] inline code: html5tagname is missing in node

Sponge5:
Fixes: #3783


matclab:
Seems to work well on my side. Thank you !

Kwpolska (approved):

PR #3820, opened 2025-02-26
Switch to pyproject.toml file

arunpersaud:
### Pull Request Checklist

- [x] I tested my changes.

### Description

I ran pytest (which in master and here didn't finish clean, but showed the same errors in both branches) and also

    uv venv -p 3.8 --seed myenv
    source myenv/bin/activate 
    pip install -e .[extras,tests]
    myenv/bin/nikola init --demo mydemo

I tried this for 3.8 and 3.13.

One change is that the new setup doesn't use -r for requirements files anymore, but the .[extras,tests] syntax during pip isntall. I believe one could just take those sections out again and put them back into the requirements files, but this way all is in one place. Let me know what you prefer. I also added some sections that were not in the setup.py files like issue page, documentation page, etc.

I also added 3.13.


Kwpolska:
It seems that:

1. `nikola/winutils.py` needs to be adjusted to work with `pathlib`
2. The flake8 configuration in `pyproject.toml` does not work

You can ignore the baseline testing failure, we will need to reconfigure the building setup to make it work on 3.13.

arunpersaud:
I think I fixed the issues above, please have a look.

I don't understand what this code does though in winutils.py (line 94):

        with io.open(os.path.join(topdir, dst), 'r', encoding='utf-8-sig') as f:
                    text = f.read()
        dst_dir = os.path.dirname(dst)
        try:
              src = os.path.normpath(os.path.join(dst_dir, text))

in the first line we join `topdir` with `dst`, but according to

        dst = os.path.join(topdir, name)  # line 86

`dst` already is based on `topdir`. Then we load the context using `f.read()` and join the context of a file with `dst_dir` which I also don't understand, so not sure if the new code is correct. It would be great, if you could have a closer look at this section. Thanks.

nikola/winutils.py
```diff
@@ -66,12 +68,12 @@ def fix_all_git_symlinked(topdir):
     """
     # Determine whether or not symlinks need fixing (they don’t if installing
     # from a .tar.gz file)
-    with io.open(topdir + r'\nikola\data\symlink-test-link.txt', 'r', encoding='utf-8-sig') as f:
-        text = f.read()
-        if text.startswith("NIKOLA_SYMLINKS=OK"):
-            return -1
-    with io.open(topdir + r'\nikola\data\symlinked.txt', 'r', encoding='utf-8-sig') as f:
-        text = f.read()
+    text = (topdir / '/nikola/data/symlink-test-link.txt').read_text(
```
Kwpolska, line 71:
```suggestion
    text = (topdir / 'nikola/data/symlink-test-link.txt').read_text(
```

nikola/winutils.py
```diff
@@ -66,12 +68,12 @@ def fix_all_git_symlinked(topdir):
...
+        encoding='utf-8-sig'
+    )
+    if text.startswith('NIKOLA_SYMLINKS=OK'):
+        return -1
+    text = (topdir / '/nikola/data/symlinked.txt').read_text(encoding='utf-8-sig')
```
Kwpolska, line 76:
```suggestion
    text = (topdir / 'nikola/data/symlinked.txt').read_text(encoding='utf-8-sig')
```

CONTRIBUTING.rst
```diff
@@ -68,7 +68,7 @@ Here are some guidelines about how you can contribute to Nikola:
   often easiest, and a good way to tell whether the feature you are modifying is
   easily testable.
 * **Test your code.** If you can, run the test suite with ``pytest tests/``
-  (you will need to install pytest and some other requirements). Alternatively,
+  (you can install all the dependencies via 'pip install .[extras,tests]'). Alternatively,
```
Kwpolska, line 71:
```suggestion
  (you can install all the dependencies via ``pip install '.[extras,tests]'``). Alternatively,
```

Kwpolska:
`winutils` appears to be working correctly after the two suggested fixes. What it tries to do is fixing symlinks on Windows. In environments where symlinks are broken, they will be replaced by a plaintext file that contains the symlinked path.

The `topdir / dst` on line 97 seems to be nonsensical though.

arunpersaud:
> The `topdir / dst` on line 97 seems to be nonsensical though.

Tried to fix this. Also renamed `is_file_into_dir` to `is_file_in_dir`.

pyproject.toml
```diff
@@ -0,0 +1,121 @@
+[build-system]
+requires = ["setuptools>=60.0.5", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "Nikola"
+version = "8.3.1"
+description = "A modular, fast, simple, static website and blog generator"
+readme = "README.rst"
+license = {text = "MIT License"}
```
Kwpolska, line 10:
```suggestion
license = {file = "LICENSE.txt"}
```

Kwpolska (approved):

Kwpolska:
Thanks for contributing this!

PR #3825, opened 2025-03-02
 nikola serve: Better error handling for already used port. Fixes #3824 

### LINKED ISSUE #3824: "nikola serve": Better error message when the port is already in use.

A better error message when the port is already in use, for example, when you start a second `nikola serve` instance.

command line interface

- [x] Yes, I have already written code for it (link if available and feasible)
- [ ] Yes, I don’t have code ready yet (that’s okay!)
- [ ] No (that’s okay too!)

No

Tell the user what the problem is and how to solve it (-p option). Alternatively nikola could also just try to find an open port (e.g. try 8001, 8002, etc)

arunpersaud:
### Pull Request Checklist

- [x] I tested my changes.

### Description
 Adds a quick test for the exception raised and prints out better error message

### DIFF
--- a/nikola/plugins/command/serve.py
+++ b/nikola/plugins/command/serve.py
@@ -26,6 +26,7 @@
 
 """Start test server."""
 import atexit
+import errno
 import os
 import sys
 import re
@@ -144,7 +145,16 @@ class CommandServe(Command):
                 handler_factory = OurHTTPRequestHandler
             else:
                 handler_factory = _create_RequestHandler_removing_basepath(base_path)
-            self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+
+            try:
+                self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+            except OSError as e:
+                if e.errno == errno.EADDRINUSE:
+                    print(f"Port address {options['port']} already in use, please use "
+                          f"the `-p <port>` option to select a different one.")
+                    sys.exit(1)
+                else:
+                    raise
 
             sa = self.httpd.socket.getsockname()
             if ipv6:
--- a/nikola/plugins/command/serve.py
+++ b/nikola/plugins/command/serve.py
@@ -151,8 +151,8 @@ class CommandServe(Command):
             except OSError as e:
                 if e.errno == errno.EADDRINUSE:
                     self.logger.error(f"Port address {options['port']} already in use, please use "
-                          "the `-p <port>` option to select a different one.")
-                    sys.exit(1)
+                                      "the `-p <port>` option to select a different one.")
+                    return 3
                 else:
                     raise
 
--- a/nikola/plugins/command/serve.py
+++ b/nikola/plugins/command/serve.py
@@ -150,8 +150,8 @@ class CommandServe(Command):
                 self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
             except OSError as e:
                 if e.errno == errno.EADDRINUSE:
-                    print(f"Port address {options['port']} already in use, please use "
-                          f"the `-p <port>` option to select a different one.")
+                    self.logger.error(f"Port address {options['port']} already in use, please use "
+                          "the `-p <port>` option to select a different one.")
                     sys.exit(1)
                 else:
                     raise
--- a/tests/integration/test_dev_server_serve.py
+++ b/tests/integration/test_dev_server_serve.py
@@ -1,6 +1,7 @@
 import logging
 import re
 import socket
+import sys
 from io import StringIO
 from time import sleep
 import requests
@@ -39,11 +40,14 @@ def test_server_on_used_port(site_and_base_path: Tuple[MyFakeSite, str]):
                 command_serve.shutdown()
                 result = future_to_run_web_server.result()
                 assert 3 == result
-                assert re.match(
-                    r"Port address \d+ already in use, "
-                    r"please use the `\-p \<port\>` option to select a different one\.",
-                    catch_log.getvalue()
-                )
+
+
+                if not sys.platform == 'win32':
+                    assert re.match(
+                        r"Port address \d+ already in use, "
+                        r"please use the `\-p \<port\>` option to select a different one\.",
+                        catch_log.getvalue()
+                    )
                 assert "OSError" not in catch_log.getvalue()
         finally:
             s.close()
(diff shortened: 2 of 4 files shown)


nikola/plugins/command/serve.py
```diff
@@ -144,7 +145,16 @@ def _execute(self, options, args):
...
             else:
                 handler_factory = _create_RequestHandler_removing_basepath(base_path)
-            self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+
+            try:
+                self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+            except OSError as e:
+                if e.errno == errno.EADDRINUSE:
+                    print(f"Port address {options['port']} already in use, please use "
+                          f"the `-p <port>` option to select a different one.")
```
aknrdureegaesr, line 154: Needs no `f"` here, just `"` is sufficient.

nikola/plugins/command/serve.py
```diff
@@ -144,7 +145,16 @@ def _execute(self, options, args):
                 handler_factory = OurHTTPRequestHandler
             else:
                 handler_factory = _create_RequestHandler_removing_basepath(base_path)
-            self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+
+            try:
+                self.httpd = OurHTTP((options['address'], options['port']), handler_factory)
+            except OSError as e:
+                if e.errno == errno.EADDRINUSE:
+                    print(f"Port address {options['port']} already in use, please use "
```
aknrdureegaesr, line 153: We should not use `print` for this. We use logging!

nikola/plugins/command/serve.py
```diff
@@ -144,7 +145,16 @@ def _execute(self, options, args):
...
+                          f"the `-p <port>` option to select a different one.")
+                    sys.exit(1)
```
aknrdureegaesr, line 155: I would like to see the `test_dev_server_server` augmented to verify this works as intended.

But doing `exit` makes this improvement much harder to test than it would otherwise be.

arunpersaud:
I fixed using the logger and the f-string issue, but I'm not sure how to test this in pytest... e.g. how do I run 2 servers on the same port and how to I run them on different ports and what do I assert? I added 2 tests, but they are currently not working. Any help would be appreciated.

aknrdureegaesr:
I wouldn't run two servers. I would simply open a port and then run the server _on the same port_.

For how to open a port, check file `dev_server_test_helper.py`, in particular the method `find_unused_port()`. It opens a port (or a socket, to be nit-picky about this), leaving the choice of port to the operating system. This will cause the OS to pick a port that's free to be opened. Then my code closes the port right away and returns the port number. As that port was free to be opened before, and the code closed the socket right away, it will be free to open after the code run. (Also, as far as I know, the OS will not hand that same port number to any other application asking for a free port number.I'm not sure that behavior is officially documented.)

You could open a port, but leave it open, and, while it is still open, ask the dev server to service that port. This should not work. This should trigger the error condition you are trying to test.

Finally, after the test is through (either showing the error text was what we expect it to be, or else the test fails), close the port again.

nikola/plugins/command/serve.py
aknrdureegaesr: After looking into this a bit more (and opening https://github.com/getnikola/nikola/issues/3828 for another issue we have with this), I found all that's needed here is a humble `return 3` instead of the `exit`.

aknrdureegaesr:
Does that idea of mine make sense to you, or do you want more detailed help on that? It is not my intention to demand big tests of you for a simple change, and then leave you "out alone in the rain", so to speak.

aknrdureegaesr:
I've been coming up with a test for your code. That test now sits in a pull request https://github.com/arunpersaud/nikola/pull/1/files which targets your branch, the branch that's again below this pull request. So if you think that's a good test, you can accept that pull request, and the test will show up as part of this pull request.

At the end of the day, we hope this will be merged here. If so, it will be all one commit, which will appear to be by you, somewhat eclipsing the fact that part of the work was be me. I don't much care, that's fine with me.

arunpersaud:
@aknrdureegaesr Thanks again for the help with the test!

aknrdureegaesr:
Sorry, my fault: We are doing pioneer work, the first test that actually checks on our exception handling.

As often, Windows behaves a bit differently, so we get a different exception than the one we are testing for in this PR. Thus the changed error handling from my PR https://github.com/getnikola/nikola/pull/3829 now gets used. But that change needs an attribute `site.show_tracebacks`. In production, that is not a problem. But the `FakeSite` we use in the tests does not provide that yet.

So this should also be set in `helper.py` in the `FakeSite` constructor, preferably one line below `self.debug = True` we need `self.show_tracebacks = True`.

When that has been done, the error for Windows will fully emerge. Other than Linux, as far as I understand, Windows differentiates between a socket opened by the same program and a socket opened by another program. We test the "socket already open" error condition here via opening a socket by the same (test) program. So this test does not carry over to Windows at all.

There is no Windows machine in this household. So I don't volunteer developing a test for Windows. If we have no such volunteer, I suggest to simply not do the `assert re.match` thingy if running under Windows.

We then do not know whether this PR works under Windows as well. It might, but I'm not sure. I'm confident it won't hurt.

arunpersaud:
I also don't have a windows computer here, so I cannot test this. I added the two items you mentioned. Let me know if this is what you had in mind.

tests/integration/test_dev_server_serve.py
```diff
@@ -9,11 +14,55 @@
...
+                    "detach": False,
+                    "ipv6": False,
+                }
+                future_to_run_web_server = executor.submit(lambda: command_serve.execute(options=options))
+                command_serve.shutdown()
+                result = future_to_run_web_server.result()
+                assert 3 == result
+
+
+                if not sys.platform == 'win32':
```
aknrdureegaesr, line 46: Imagine someone like you stumbles over this in a year or two. This line raises an eyebrow and leaves the puzzling question to be answered: "Why?"

Can you please add a comment here that'll help that someone?

(We need our Nikola community members and want to make life nice for them!)

aknrdureegaesr: We now see the original Windows exception. We would like to see the `result` being 3. I think we'll get that by setting `show_tracebacks` to `False` in `MyFakeSite`.

It should be `True` in general, which is what `FakeSite` now has, but we need `False` for this particular test, which `MyFakeSite` being the convenient place to get that.

arunpersaud: ok, so I would set site.show_tracebacks=False inside the test (say right at the top of the test). Is this only needed for windows or for all tests?

I can add some comments later today and can also test this on linux and mac.

arunpersaud:
Just checking in on the status of this.

PR #3831, opened 2025-03-07
Add 'slug_source' path handler

laodzu:
This allows the creation of links to the source for a blog post.

### Pull Request Checklist

- [x] I tested my changes.

### DIFF
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -379,6 +379,7 @@ class Nikola(object):
         # Register our own path handlers
         self.path_handlers = {
             'slug': self.slug_path,
+            'slug_source': self.slug_source,
             'post_path': self.post_path,
             'root': self.root_path,
             'filename': self.filename_path,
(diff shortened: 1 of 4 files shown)


nikola/nikola.py
```diff
@@ -1970,6 +1971,21 @@ def slug_path(self, name, lang):
...
+
+        link://slug_source/yellow-camaro => /posts/cars/awful/yellow-camaro.rst
+        """
+        results = [p for p in self.timeline if p.meta('slug') == name]
+        if not results:
+            utils.LOGGER.warning("Cannot resolve path request for slug: {0}".format(name))
+        else:
+            if len(results) > 1:
+                utils.LOGGER.warning('Ambiguous path request for slug: {0}'.format(name))
+            return [_f for _f in results[0].source_path.split('/')]
```
Kwpolska, line 1987: Let’s also support translations: (please test this before accepting the suggestion)

```suggestion
            return [_f for _f in results[0].translated_source_path(lang).split('/')]
```

laodzu: That is of course an excellent idea!  As my blog is not multi-lingual, I needed some time to even test this.  But my testing was not successful and by inserting logs, I think there is a more fundamental problem.  I instrumented 'slug_path' and 'slug_source' and the 'lang' parameter passed to them is always 'en' for me, even when the 'de' translation is rendered. So of course your suggestion currently yields the same result and to make it work, one would need to find the cause for the incorrect lang parameter.  Not being too familiar with the codebase, this will take me some more time, but maybe you have an obvious explanation, or maybe I am doing something very stupid?
Just as an additional data point - for my (artificial) 'de' translation I at least see the 'source' link at the top of the post to point to the German translation instead of the English version, so I think the fundamental multi-lang machinery is working.

laodzu: Ah, operator error, sorry.  I also need to enable the other language in conf.py :facepalm:
Once I did this, thinks work like you expected them to. Very cool! So I can commit this change.

laodzu: Ok, I amended the commit message to include the fact that this works also for translations and force pushed the branch. While doing this, I started questioning the naming of 'slug_source'. Maybe 'slug_src' is more concise? I don't know, but before we commit to it, I wanted to raise my own doubts and ask for input on this. Thanks in advance!

Kwpolska: Nah, the longer name is better.

laodzu: Ok, thanks for confirming. Then I'll just wait and hope that it can be merged at some time.

Kwpolska (approved):

Kwpolska:
Thanks for contributing!

laodzu:
Thanks for the immediate feedback and the valuable improvement! Being able to contribute in such a way is a pure joy :heart:

PR #3834, opened 2025-03-13
Run CI using free-threading Python 3.13t

Kwpolska:


Kwpolska:
The `setup-python` action does not support `3.13t` yet [(it will in the next release)](https://github.com/actions/setup-python/commit/9e62be81b28222addecf85e47571213eb7680449).

PR #3846, opened 2025-05-03
Close ``<a>`` tag instead of opening another one in ``tag.tmpl``

aknrdureegaesr:
### Pull Request Checklist

- [X] I tested my changes.

### Description

There is a HTML glitch in the tags template: Some `<a>` is not closed with a corresponding `</a>`, but instead, another `<a>`  is opened.

This trivial pull request fixes this.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,8 @@ Bugfixes
 
 * Fix jinja2 plugin failing to load when jinja2 is not installed
   (Discussion #3848)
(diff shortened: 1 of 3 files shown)


felixfontein:
Thanks for your contribution! Please note that the jinja templates are created automatically from the mako templates, so you also need to update nikola/data/themes/base/templates/tag.tmpl. Otherwise it will be changed back when the jinja templates are regenerated the next time :) (Also the mako template is broken in the same way, so it will also be great to fix that ;) )

Kwpolska (approved):

PR #3850, opened 2025-05-24
Remove setuptools from runtime dependencies specification

lahwaacz:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Minor updates for dependencies specification.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,6 @@ dependencies = [
     "lxml>=4.5.2",
     "PyRSS2Gen>=1.1",
     "blinker>=1.3",
-    "setuptools>=60.0.5",
     "natsort>=5.1.0",
     "requests>=2.31.0",
     "piexif>=1.0.3",
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,7 @@ Bugfixes
 * Fix jinja2 plugin failing to load when jinja2 is not installed
   (Discussion #3848)
 * Close ``<a>`` tag instead of opening another one in ``tag.tmpl``
+* Setuptools is not required anymore at runtime (PR #3850)
 
 New in v8.3.3
 =============


pyproject.toml
```diff
@@ -54,11 +54,10 @@ dependencies = [
     "lxml>=4.5.2",
     "PyRSS2Gen>=1.1",
     "blinker>=1.3",
-    "setuptools>=60.0.5",
     "natsort>=5.1.0",
     "requests>=2.31.0",
     "piexif>=1.0.3",
-    "Babel>=2.12.0",
+    "Babel>=2.17.0",
```
felixfontein, line 60: Do versions < 2.17 not work? Or why do you bump the requirement?

pyproject.toml
```diff
@@ -54,11 +54,10 @@ dependencies = [
     "lxml>=4.5.2",
     "PyRSS2Gen>=1.1",
     "blinker>=1.3",
-    "setuptools>=60.0.5",
```
felixfontein, line 57: Can you mention this in `CHANGES.txt` in a new top-level section `New in master`?

felixfontein (commented):
Thanks for your contribution!

pyproject.toml
lahwaacz: Sure, done.

pyproject.toml
felixfontein: So generally Babel works fine, it's just the unit test that fails? I don't think we want to bump the requirement in that case. This only makes life hard for our users.

lahwaacz: Well there was some behavioral change and since there is a test for it, it seemed important enough :shrug:

Kwpolska: The minimum versions in `pyproject.toml` do not guarantee all tests pass. Our tests assume the latest versions of all dependencies. If we changed the minimum version here, then Arch Linux would be unable to package Nikola until they upgraded Babel. But Nikola is working fine with the older version, just might have slightly different output for some date formats/locales. Please undo this change.

Kwpolska (changes requested):
Please undo the change to Babel.

Kwpolska (approved):

Issue #3865, opened 2025-10-15
Option to strip extensions from permalinks


jwise:
**Requested Feature:** (short description)

It would be nice to have an option to remove filename extensions from canonical permalinks.  That is to say, rather than a canonical name of either `https://mysite/foo/index.html`, or `https://mysite/foo.html`, or `https://mysite/foo/`, the name is `https://mysite/foo`.

**Related Area:** (eg. tasks, compilers, configuration, templates…)

timeline, posts

**Do you want to contribute this yourself as a pull request?** (don’t worry about it if you don’t want to/can’t — someone else can take care of it)

- [x] Yes, I have already written code for it (link if available and feasible)
- [ ] Yes, I don’t have code ready yet (that’s okay!)
- [ ] No (that’s okay too!)

Here is what I will put together as as a PR soon: https://github.com/jwise/nikola/tree/jwise/strip_extensions

**Does this feature affect backwards compatibility? If yes, in what way?**

No.

**Rationale and full description:** (why should it be added to Nikola?)

As Sir Tim Berners-Lee says, "[cool URIs don't change](https://www.w3.org/Provider/Style/URI)`!  In this, he gives an example:

> File name extension. This is a very common one. "cgi", even ".html" is something which will change. You may not be using HTML for that page in 20 years time, but you might want today's links to it to still be valid. The canonical way of making links to the W3C site doesn't use the extension.

My site has been doing this for quite some time -- previously, I had mod_rewrite rules set up to re-aim paths inside of Dokuwiki, and before that, I had PHP scripts but did not ever write the `.php` on the end, exactly for this reason.  So I want to maintain backwards compatibility with URIs that I had in the past -- and forward compatibility with URIs that I may have in the future.

As per the help for the config item that I suggest:

```
+# If a link ends in .html, drop the .html part only.  (Works with Apache
+# when configured with `Options MultiViews`, or nginx when configured with
+# `try_files`.)
+#
+# http://mysite/foo/bar.html => http://mysite/foo/bar
+STRIP_EXTENSIONS = ${STRIP_EXTENSIONS}
```

PR #3866, opened 2025-10-17
Add STRIP_EXTENSIONS config item

jwise:
### Pull Request Checklist

- [X] I tested my changes.

### Description

This implements #3865 -- as seen in production on https://joshuawise.com/ .

### DIFF
--- a/nikola/post.py
+++ b/nikola/post.py
@@ -172,6 +172,7 @@ class Post(object):
             self.current_time = current_time(self.config['__tzinfo__'])
         self.base_url = self.config['BASE_URL']
         self.strip_indexes = self.config['STRIP_INDEXES']
+        self.strip_extensions = self.config['STRIP_EXTENSIONS']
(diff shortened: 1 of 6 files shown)


nikola/post.py
```diff
@@ -1069,6 +1070,8 @@ def permalink(self, lang=None, absolute=False, extension='.html', query=None):
         index_len = len(self.index_file)
         if self.strip_indexes and link[-(1 + index_len):] == '/' + self.index_file:
             link = link[:-index_len]
+        if self.strip_extensions and link[-len(extension):] == extension:
+            link = link[:-len(extension)]
```
felixfontein: While above you explicitly handle `.html`, here you handle general extensions. This doesn't fit together.

How about making `STRIP_EXTENSIONS` a list instead that provides a list of extensions that will be stripped? Then users can explicitly configure to strip `.html` but not `.php`, or both, or ...

PR #3893, opened 2026-04-26
Add type annotations to core classes and plugin infrastructure

arunpersaud:
### Pull Request Checklist

- [x] I tested my changes.

### Description

Adds type hints for site, logger, and various Nikola class attributes to improve type checking. Testing using pytest and 'ty'

### DIFF
--- a/nikola/hierarchy_utils.py
+++ b/nikola/hierarchy_utils.py
@@ -26,8 +26,6 @@
 
 """Hierarchy utility functions."""
 
-from __future__ import annotations
-
 import natsort
 
 __all__ = ('TreeNode', 'clone_treenode', 'flatten_tree_structure',
@@ -41,8 +39,6 @@ class TreeNode:
     indent_levels = None  # use for formatting comments as tree
     indent_change_before = 0  # use for formatting comments as tree
     indent_change_after = 0  # use for formatting comments as tree
-    classification_path: list[str]
-    classification_name: str
 
     # The indent levels and changes allow to render a tree structure
     # without keeping track of all that information during rendering.
--- a/nikola/hierarchy_utils.py
+++ b/nikola/hierarchy_utils.py
@@ -26,6 +26,8 @@
 
 """Hierarchy utility functions."""
 
+from __future__ import annotations
+
 import natsort
 
 __all__ = ('TreeNode', 'clone_treenode', 'flatten_tree_structure',
@@ -39,6 +41,8 @@ class TreeNode:
     indent_levels = None  # use for formatting comments as tree
     indent_change_before = 0  # use for formatting comments as tree
     indent_change_after = 0  # use for formatting comments as tree
+    classification_path: list[str]
+    classification_name: str
 
     # The indent levels and changes allow to render a tree structure
     # without keeping track of all that information during rendering.
--- a/nikola/nikola.py
+++ b/nikola/nikola.py
@@ -27,6 +27,7 @@
 """The main Nikola site object."""
 
 import datetime
+from datetime import timezone, tzinfo
 import io
 import json
 import functools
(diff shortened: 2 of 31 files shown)


nikola/plugin_categories.py
```diff
@@ -59,10 +64,10 @@
 class BasePlugin:
     """Base plugin class."""
 
-    logger = None
-    site: Optional['nikola.nikola.Nikola'] = None # NOQA  # Circular import not easy to fix.
+    logger: logging.Logger
+    site: Nikola
```
felixfontein: Both can be `None` (and are in `__init__`, they only get values in `set_site()`).

nikola/hierarchy_utils.py
```diff
@@ -39,6 +41,8 @@ class TreeNode:
     indent_levels = None  # use for formatting comments as tree
     indent_change_before = 0  # use for formatting comments as tree
     indent_change_after = 0  # use for formatting comments as tree
+    classification_path: list[str]
+    classification_name: str
```
felixfontein: Having the types here is not good, but not having them here is not good either.

There's currently no way to model this correctly in Python's type system: https://github.com/python/typing/discussions/1912

It might be better to comment these out, and use `type: ignore` when accessing them.

nikola/nikola.py
felixfontein: Since the type is above, it should be removed here:
```suggestion
        self._template_system = None
```

nikola/nikola.py
```diff
@@ -372,7 +373,15 @@ class Nikola:
     """
 
     plugin_manager: PluginManager
-    _template_system: TemplateSystem
+    _template_system: TemplateSystem | None
+    tzinfo: datetime.tzinfo | None
```
felixfontein, line 377: How can `tzinfo` be `None`?

nikola/nikola.py
```diff
@@ -372,7 +373,15 @@ class Nikola:
...
+    themes_dirs: list[str]
+    state: Persistor
+    posts_per_classification: dict[str, dict[str, list[Post]]]
+    hierarchy_per_classification: dict[str, dict[str, list]]
+    flat_hierarchy_per_classification: dict[str, dict[str, list]]
+    hierarchy_lookup_per_classification: dict[str, dict[str, dict]]
+    page_count_per_classification: dict[str, dict[str, dict]]
```
felixfontein: These are fields that don't necessarily exist, and thus these type hints are not correct (these hints indicate that they always exist).

felixfontein (commented):
Thanks for working on this!

I guess this shows that adding type hints to an existing non-trivial Python codebase is hard. (I've been struggling with that myself more than once, even if you don't have the trouble of attributes that aren't always there.)

Regarding the attributes that aren't always there: maybe we should change them to `attribute: type | None` and initialize them with `None`?

Also regarding attribute type hints: I (nowadays) think it's better to add their types in `__init__` than already in the class definition. The rationale is that when they are defined on the class level, the type checker won't notice if you use them too early in `__init__` (before actually setting them). (They might still notice, I don't know how sophisticated they are nowadays, but better safe than sorry.) Also it's harder to forget to actually set an attribute that was declared to exist. Especially for complex constructors like `Nikola.__init__` with its many hundreds of lines, this can make a difference.

nikola/nikola.py
arunpersaud: tzinfo is set by dateutil.tz.gettz, which returns None on a non-valid timezone request
```
>>> dateutil.tz.gettz('asdfsadf') is None
True
```

arunpersaud: This is using 

```
>>> dateutil.__version__
'2.9.0.post0'
```

Perhaps in addition to the try-except block NIkola should check for None and do the same as in the try except?

nikola/nikola.py
arunpersaud: ok, I will remove these

nikola/hierarchy_utils.py
arunpersaud: I'll just remove them for now and keep with the original code for now

arunpersaud:
As for moving things to __init__. Happy to look into this.  The main Nikola class now has

```
class Nikola:
    #[...]

    plugin_manager: PluginManager
    _template_system: TemplateSystem | None
    tzinfo: datetime.tzinfo | None
    themes_dirs: list[str]
    state: Persistor
```

And I think some, for example, `state` could just be deleted, since it is set to `self.state = Persistor(...)` and can be inferred by tools like `ty`? The other ones I can move where they are defined as `None`. 

I could add this here or perhaps as another PR in the future for more work on the typing. Let me know what you prefer

binggao1230:
The current lint failure is from one unused import line in `nikola/nikola.py`:

```text
nikola/nikola.py:30:1: F401 `datetime.timezone` imported but unused
nikola/nikola.py:30:1: F401 `datetime.tzinfo` imported but unused
```

The class annotation in this PR uses `datetime.tzinfo | None`, so the added direct import is not needed. Removing this line should fix the lint job:

```diff
 import datetime
-from datetime import timezone, tzinfo
 import io
```

The rest of the CI matrix is green.

PR #3894, opened 2026-04-26
Fix Python deprecation warnings (LOGGER.warn, datetime.utcnow, docutils   2.0)

arunpersaud:
### Pull Request Checklist

- [x] I tested my changes.

### Description
Updates deprecated APIs to their modern equivalents for better Python 3.13+ compatibility. These showed up in 'ty' and pytest.


nikola/plugins/compile/rest/listing.py
```diff
@@ -40,7 +40,7 @@
 from docutils import core
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
-from docutils.parsers.rst.roles import set_classes
+from docutils.parsers.rst.roles import normalize_options
```
felixfontein: Here I would use a conditional import:
```suggestion
try:
    from docutils.parsers.rst.roles import normalize_options
except ImportError:
    from docutils.parsers.rst.roles import set_classes as normalize_options
```
This way it also works with docutils < 0.22.

nikola/plugins/compile/rest/__init__.py
```diff
@@ -371,10 +371,12 @@ def rst2html(source, source_path=None, source_class=docutils.io.StringInput,
...
                               })
 
-    pub = docutils.core.Publisher(reader, parser, writer, settings=settings,
+    pub = docutils.core.Publisher(reader,
+                                  parser=parser or parser_name,
+                                  writer=writer or writer_name,
+                                  settings=settings,
                                   source_class=source_class,
                                   destination_class=docutils.io.StringOutput)
-    pub.set_components(None, parser_name, writer_name)
```
felixfontein, line 377: This one is harder. Maybe conditional code based on the docutils version will work here?

felixfontein (commented):
While most changes look good, I'm not sure requiring docutils >= 0.22 is a good idea. 0.22 got released less than a year ago, and is likely not available on some of the platforms Nikola should work on. Ubuntu 24.04 for example only 0.20.1, and even Ubuntu 25.10 only has 0.21.2. (Ubuntu 26.04 has 0.22.4.)

nikola/plugins/compile/rest/listing.py
arunpersaud: good point, reverted this partially and added the above

nikola/plugins/compile/rest/__init__.py
arunpersaud: I just reverted this for now

arunpersaud:
@felixfontein thanks for looking at this. Let me know if this looks better

PR #3895, opened 2026-04-26
Created a new theme (mako and jinja) based on bulma

arunpersaud:
### Pull Request Checklist

- [x] I tested my changes.

### Description

This adds a bulma theme for both template engines. I created a `nikola init -qd mytest`, changed the conf.py to the new theme (bulma or bulma-jinja), build the page and served it. I adjusted the css so that somewhat matches the old schemes. If there are other good tests that one could use, I'm happy to run those too.

I'm not a designer ;) so there probably could be more improvements to be made, but I think it's nice to have a more modern CSS framework available. Happy to modify as needed and perhaps further changes can come in future PRs?

Also, it seems that html5shiv got removed, but was still present in the npm_asset directory, so this PR also removes this.


Kwpolska:
New themes are not allowed in the main repo. You can contribute to the [themes repo](https://github.com/getnikola/nikola-themes).

PR #3907, opened 2026-09-20
Skip WordPress attachments whose URL escapes the files directory

### LINKED ISSUE #3904: import_wordpress writes attachment files outside output/files when a WXR uses ../ in attachment_url
Splitting this out from the closed thread as a plain bug, not framing it as a security issue.

In import_wordpress, import_attachment builds the download path straight from the WXR attachment_url with no containment:

    path = urlparse(url).path
    dst_path = os.path.join(*([self.output_folder, 'files'] + list(path.split('/'))))
    ...
    self.download_url_content_to_file(url, dst_path)

urlparse doesn't collapse "..", so a WXR whose <wp:attachment_url> is something like http://host/../../../../etc/whatever ends up written outside output_folder/files. In a quick local test the file landed on conf.py. The postmeta thumbnail handler a bit further down does the same join.

Simplest fix is to skip the write if the resolved path escapes the files dir, e.g. after building dst_path:

    base = os.path.realpath(os.path.join(self.output_folder, 'files'))
    if os.path.commonpath([base, os.path.realpath(dst_path)]) != base:
        LOGGER.warn("Skipping attachment outside files dir: %s", url)
        continue

Happy to send a PR for both spots if you want.

niukanen1:
In `import_wordpress`, `import_attachment` built the download path straight from the WXR `attachment_url` with no containment, so a URL containing `..` wrote files outside `output/files` — in a quick local test the file landed on `conf.py`. The postmeta thumbnail handler a bit further down joined paths the same way.

Both destination paths are now resolved against the real files directory and the write is skipped with a warning when the resolved path escapes it.

Fixes #3904


nikola/plugins/command/import_wordpress.py
```diff
@@ -538,7 +543,9 @@ def import_attachment(self, item, wordpress_namespace):
                             'foo')
         path = urlparse(url).path
         dst_path = os.path.join(*([self.output_folder, 'files'] + list(path.split('/'))))
-        if self.no_downloads:
+        if not self.path_in_files_dir(dst_path):
+            LOGGER.warning("Skipping attachment outside files dir: {0}".format(url))
```
felixfontein: I'd use f-strings here, and also print the bad path:
```suggestion
            LOGGER.warning(f"Skipping attachment outside files dir: {url} would ned up in {dst_path}")
```

nikola/plugins/command/import_wordpress.py
```diff
@@ -643,7 +650,9 @@ def add(our_key, wp_key, is_int=False, ignore_zero=False, is_float=False):
 
                         path = urlparse(url).path
                         dst_path = os.path.join(*([self.output_folder, 'files'] + list(path.split('/'))))
-                        if self.no_downloads:
+                        if not self.path_in_files_dir(dst_path):
+                            LOGGER.warning("Skipping attachment outside files dir: {0}".format(url))
```
felixfontein: Same here:
```suggestion
                            LOGGER.warning(f"Skipping attachment outside files dir: {url} would end up in {dst_path}")
```

felixfontein (commented):
Thanks for your contribution!

niukanen1:
Done in b79e85c — both warnings now use f-strings and include the destination path.

